PR #3, opened 2019-03-12
Rename all references to gitlab.com/regen-network/regen-ledger -> github.com/regen-network/regen-ledger

aaronc:


aaronc:
@cpolitano just making sure you've seen this, but going to go ahead and merge it

Issue #5, opened 2019-03-13
Spec and basic types for minimum viable oracle module


aaronc:
The goal of the oracle module is to implement consensus protocols via which we can verify a claim that the computation `F` with input `A` has result `B`.

* Should specify both the minimum viable on-chain and off-chain data structures for computation results
* Should specify the mechanics of the minimum viable oracle consensus protocol(s). We may specify one called "fiat" which simply says that an oracle with address X can verify any computations it likes
* It would be good to expand on a more complete oracle consensus protocol and reach an initial draft, but it is okay to mark it as WIP

clevinson:
@aaronc should this get closed ?

PR #29, opened 2019-04-01
#5 minimum viable oracle module

aaronc:


clevinson:

PR #30, opened 2019-04-01
#27 Create claim module

### LINKED ISSUE #27: Create claim module
This is to replace or at least complement the current `esp` module with something more generic, offloading the responsibility of whether a claim is valid to other logic.

Use Cases and Rationale:
The claims model is a generic way to represent who said what and why. The what is contained in the data graph - the claim itself, and who is managed by the claims module which allows many entities to assert the same claim. Many times claims are going to be signed by trusted third parties. So what creates the audit trail if that is not contained in the claim? This is where evidence fits in - oftentimes there might be off-chain data that is even just scanned paper documents which represents what the claim means or somebody has done thorough investigation and produced a report. These can all be linked to a claim via evidence which is the "why the claim is true" and the claim signature effectively who said what and why.

Claims themselves will now to be stored in the `data` module which will eventually abstract "graph descriptors" that include what the graph schema and subjects (i.e `GeoAddress`'s or `sdk.AccAddress`'s. The `claim` module just includes a very basic apparatus for signing claims.

A claim signature includes three basic pieces of data:
- the claim itself (a `DataAddress`)
- the signer(s) (`[]sdk.AccAddress`)
- any supporting evidence (`[]DataAddress`) - on or off chain graphs, and importantly off-chain raw data which could be photographs, PDFs, etc.

One important piece of behavior:
Given an existing claim that is already signed
When somebody tries to sign the same claim
Then new evidence and/or signatures will be recorded if any

aaronc:
Closes #27

### DIFF
--- a/cmd/xrncli/main.go
+++ b/cmd/xrncli/main.go
@@ -23,13 +23,13 @@ import (
 	bankcmd "github.com/cosmos/cosmos-sdk/x/bank/client/cli"
 	bank "github.com/cosmos/cosmos-sdk/x/bank/client/rest"
 	"github.com/regen-network/regen-ledger"
-	claimcli "github.com/regen-network/regen-ledger/x/claim/client/cli"
 	consortiumclient "github.com/regen-network/regen-ledger/x/consortium/client"
 	dataclient "github.com/regen-network/regen-ledger/x/data/client"
 	datarest "github.com/regen-network/regen-ledger/x/data/client/rest"
 	espclient "github.com/regen-network/regen-ledger/x/esp/client"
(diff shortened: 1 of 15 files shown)


aaronc:
@cpolitano so another thing that I want to make sure we do for each module is have at least basic documentation.

You can check the status of this by installing [golint](https://github.com/golang/lint) and running `golint x/claim/...`. Eventually I want to turn this on in CI, but first we'll have to add doc strings everywhere or CI will fail

aaronc:
@cpolitano wrapped up some docs on this and going to merge this

PR #51, opened 2019-04-24
Support basic off chain raw data tracking

### LINKED ISSUE #39: Support basic off-chain raw data tracking
The minimum viable functionality for this is to let users store an SHA-256 hash with an associated URL where the data can be retrieved from. SHA-256 was chosen because of its ubiquity and usage in IPFS which may be a common data store used.

We probably want a message like this:
```go
type MsgTrackRawData struct {
  Sha256Hash []byte
  Url string
  Signer sdk.AccAddress
}
```

AC:
- [ ] tx tags should include the `DataAddress` which includes a new case for Sha256 hashes and doesn't otherwise include anything about the URL
- [ ] it is not an error to submit multiple URL's for the same hash. We are essentially letting someone make a claim that the data exists at such and such URL. There is nothing wrong with submitting multiple URL's and oracles should deal with 404's or hash errors gracefully behind the scenes
- [ ] it is also not an error for `Url` to be `nil` because it should be possible to track data (i.e. timestamp) it without providing an URL
- [ ] there should be associated QueryStore method (not a `custom/` route) to get a list of URL's registered for the hash. This involves exposing a function for generating the store key for these URL's from the `DataAddress`
- [ ] there should be an `xrncli tx data track-raw` command that takes a hash and an optional URL
- [ ] there should be an `xrncli query data get <bech-address>` command and `/data/<bech-address>` REST route that work for any `DataAddress` and return the relevant info
- [ ] we should check that the length of the hash if valid for sha256 and reject it if not

aaronc:
Closes #39 support basic off chain raw data tracking.

Also removes existing store graph data functionality from the data module, as well as the claims module. These functionalities may be added again in the future, but they need more thorough design before being enabled.

- [ ] linked to a relevant issue or have written a clear description of the issue in the PR
- [ ] wrote or updated tests


clevinson:

PR #53, opened 2019-05-03
Update CONTRIBUTING.md with Maya Angelou

CShear:
I added in a poem, and more importantly, this is the first time I have ever made a pull request (besides the one I did on gitlab 20 minutes ago before clare informed me that we don't use that any more. 

Let me know if it looks like I did this right!


CShear:
Hey Aaron, I wanted to respect protocol and not commit this myself, so I submitted this pull request. Would you be willing to commit it for me? 
Thanks!

PR #54, opened 2019-05-21
Create stub for validator how to

swidnikk:
* Worked with pete to stub out 
[starts #166162869]


clevinson:

PR #66, opened 2019-07-25
Change PowerReduction from 1 million to 1 hundred for testnet

### LINKED ISSUE #65: Change PowerReduction for testnet
Currently it is set to `1000000` which causes `1000000tree` tokens to actually have a voting power of `1`. We want to change this temporarily for `regen-test-1001` and then set it to a higher value and simply issue more tokens for the next testnet and mainnet.

aaronc:
Close #65

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -11,6 +11,7 @@ build:
   script:
     - make install
   artifacts:
+    name: $CI_COMMIT_REF_SLUG
     paths:
       - "$GOPATH/bin/xrnd"
       - "$GOPATH/bin/xrncli"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,6 +10,10 @@ build:
   image: golang:1.11.5
   script:
     - make install
+  artifacts:
+    paths:
+      - "$GOPATH/bin/xrnd"
+      - "$GOPATH/bin/xrncli"
 
 test:
   image: golang:1.11.5
--- a/app/app.go
+++ b/app/app.go
@@ -20,6 +20,7 @@ import (
 	"github.com/tendermint/tendermint/libs/log"
 	"io"
 	"io/ioutil"
+	"math/big"
 	"os"
 	"path/filepath"
 	//"os"
@@ -68,6 +69,8 @@ func init() {
 		geo.AppModuleBasic{},
 		upgrade.AppModuleBasic{},
 	)
+
+	sdk.PowerReduction = sdk.NewIntFromBigInt(new(big.Int).Exp(big.NewInt(10), big.NewInt(2), nil))
 }
 
 // custom tx codec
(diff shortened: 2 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -10,6 +10,11 @@ build:
   image: golang:1.11.5
   script:
     - make install
+  artifacts:
```
ethanfrey: Maybe this can go in another PR?

Doesn't seem directly related to the PR title

app/app.go
```diff
@@ -68,6 +69,8 @@ func init() {
 		geo.AppModuleBasic{},
 		upgrade.AppModuleBasic{},
 	)
+
+	sdk.PowerReduction = sdk.NewIntFromBigInt(new(big.Int).Exp(big.NewInt(10), big.NewInt(2), nil))
```
ethanfrey: So... I dug quite a bit to the usage of this variable.

It seems that:

1. Setting it in the constructor is the best way to override AND
2. You do not actually need to re-calculate all the validators powers, as this is done every EndBlock automatically, so will be updated at the end of the first block after the upgrade.

```
// Apply and return accumulated updates to the bonded validator set. Also,
// * Updates the active valset as keyed by LastValidatorPowerKey.
// * Updates the total power as keyed by LastTotalPowerKey.
// * Updates validator status' according to updated powers.
// * Updates the fee pool bonded vs not-bonded tokens.
// * Updates relevant indices.
// It gets called once after genesis, another time maybe after genesis transactions,
// then once at every EndBlock.
//
// CONTRACT: Only validators with non-zero power or zero-power that were bonded
// at the previous block height or were removed from the validator set entirely
// are returned to Tendermint.
func (k Keeper) ApplyAndReturnValidatorSetUpdates(ctx sdk.Context) (updates []abci.ValidatorUpdate) {
```

app/app.go
```diff
@@ -198,6 +201,7 @@ func NewXrnApp(logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest bo
 	app.crisisKeeper = crisis.NewKeeper(crisisSubspace, invCheckPeriod, app.distrKeeper,
 		app.bankKeeper, app.feeCollectionKeeper)
 	app.upgradeKeeper = upgrade.NewKeeper(app.upgradeStoreKey, app.cdc)
+	app.upgradeKeeper.SetUpgradeHandler("regen-test-1000-upgrade-1", func(ctx sdk.Context, plan upgrade.Plan) { })
```
ethanfrey: I would add a comment here, saying this is a checkpoint for code upgrade - in particular PR #66 (powerreduction change)

app/app.go
ethanfrey: Adding a short comment before this line would not hurt however.

.gitlab-ci.yml
aaronc: Normally yes, but I want to have upgrade binaries automatically generated in CI. I tested it and it works...

aaronc: Okay if I keep it here for now @ethanfrey ?

app/app.go
aaronc: Okay, so this seems like a correct way to do this? It's in an `init` function.

.gitlab-ci.yml
ethanfrey: Sure. More a comment for style, and general PR flow.

But I see the urgency here and it's a super minor point. Let's get this out and merged.

app/app.go
ethanfrey: Anything before the first `EndBlock` call should work well.

`init` is good to ensure that the default sdk value is never read anywhere

Issue #108, opened 2020-08-17
Data Module


aaronc:
## Summary 

This proposes a module for tracking data related to ecological claims that lives both on and off-chain.

## Problem Definition

The data module aims to satisfy the following use cases:
* provide secure timestamps and signatures for data used to verify ecological state to allow for both programmatic verification and auditing
* provide signed data to on-chain programmatic smart contracts
* allow the blockchain to be used as an index for off-chain data
* store on-chain metadata for ecosystem service credit fractional NFTs (#78) using a generic data system  

Can you say any more about this @clevinson ?

### Anchoring Data

Anchoring data refers to storing a hash of a piece of data on-chain that allows for proof of existence of that data at some block height. This effectively creates a “secure timestamp” for the data which proves that the data was created no later than the block height where it was included.

We propose a simple message `MsgAnchorData` for anchoring data on chain which uses the [IPFS CID Specification](https://github.com/multiformats/cid) as the expected format for data hashes:

```protobuf
message MsgAnchorData {
  // sender is the address of the party submitting the transaction
  bytes sender = 1;
  // cid is a binary IPFS CID identifier without a multibase prefix
  bytes cid = 2; 
}
```

### Signing Data

Signing data refers to creating a claim that the signer of the piece of data attests to its veracity. What veracity means may be somewhat dependent on the context of the document being signed, but for simplistic purposes we draw an analogy to a legal document. If a party puts their signature on a legal document, its pretty clear from the contents of the legal document what the signature implies.

We propose a simple message `MsgSignData` for making on-chain signatures to data:

```protobuf
message MsgSignData {
  // signers is the addresses of the signers of the document
  repeated bytes signers = 1;
  // cid is the binary IPFS CID identifier of the document
  bytes cid = 2;
}
```

A few notes:

- `signers` is distinct from `sender` in `MsgAnchorData`. The `sender` in `MsgAnchorData` could be a third-party relayer like a “postman”. Just delivering a document does not mean that one has signed the document in the legal sense. So `MsgAnchorData` implies that the `sender` simply delivered the document to the registry for timestamping. `MsgSignData` implies that the `signers` _signed_ the document.
- submitting `MsgSignData` automatically “anchors” the document on-chain. A call to `MsgAnchorData` is not needed if this is the first time the data has appeared on-chain. Two different messages are proposed because the data may be anchored on-chain before it is signed and different signers may sign the same document at different points in time.
[...]

PR #124, opened 2020-11-03
Data Module Proof of Concept Init

aaronc:
ref #108 

This PR:
* adds .proto files for the `x/data` module based on the specification in #108 
* adds basic protobuf compilation support
* gitignores `.idea`

`Protobuf/breakage` is _expected_ to fail.

---

- [x] linked to a relevant issue or have written a clear description of the issue in the PR
- [ ] wrote or updated tests

Thanks!


proto/regen/data/v1alpha1/tx.proto
```diff
@@ -0,0 +1,92 @@
...
+    //
+    // StoreData implicitly calls AnchorData if the data was not already anchored.
+    rpc StoreData(MsgStoreDataRequest) returns (MsgStoreDataResponse);
+}
+
+// MsgAnchorDataRequest is the Msg/AnchorData request type.
+message MsgAnchorDataRequest {
+
+    // sender is the address of the sender of the transaction.
+    string sender = 1;
```
clevinson, line 46: Why must this be included in the msg content itself? Isn't it accessible from the transaction context?

proto/regen/data/v1alpha1/tx.proto
```diff
@@ -0,0 +1,92 @@
...
+    // cid is a Content Identifier for the data corresponding to the IPFS CID
+    // specification: https://github.com/multiformats/cid.
+    bytes cid = 2;
+}
+
+// MsgAnchorDataRequest is the Msg/AnchorData response type.
+message MsgAnchorDataResponse {
+
+    // timestamp is the timestamp of the block at which the data was anchored.
+    google.protobuf.Timestamp timestamp = 1;
```
clevinson, line 57: Does cosmos have a way to provide deterministic timestamps? At first glance this looks like there could be security issues if a validator has an incorrect / malicious clock time.

Also- is this timestamp stored in blockchain state somewhere? I don't see it in any other proto definitions. Do you still need to seperately add proto definitions for state encoding?

As for handling misbehavior- maybe that's something we can rectify later though in the form of evidence / slashing?

proto/regen/data/v1alpha1/tx.proto
```diff
@@ -0,0 +1,92 @@
...
+// MsgSignDataRequest is the Msg/SignData request type.
+message MsgSignDataRequest {
+    option (gogoproto.goproto_getters) = false;
+
+    // signers are the addresses of the accounts signing the data.
+    repeated string signers = 1;
+
+    // cid is a Content Identifier for the data corresponding to the IPFS CID
+    // specification: https://github.com/multiformats/cid.
+    bytes cid = 2;
```
clevinson, line 69: Should we specify if we expect this to be cidv1, vs v0? If we were using v0 then I may prefer string as opposed to bytes as the representation (after looking at [decoding a cid](https://github.com/multiformats/cid#decoding-algorithm) documentation).

proto/regen/data/v1alpha1/tx.proto
```diff
@@ -0,0 +1,92 @@
...
+    // timestamp is the timestamp of the block at which the data was anchored.
+    google.protobuf.Timestamp timestamp = 1;
+}
+
+// MsgSignDataRequest is the Msg/SignData request type.
+message MsgSignDataRequest {
+    option (gogoproto.goproto_getters) = false;
+
+    // signers are the addresses of the accounts signing the data.
+    repeated string signers = 1;
```
clevinson, line 65: Are hte signatures themselves intended to be accompanied elsewhere in the transaction? It's unclear to me how the signatures are verified if they aren't included in the msg itself.

clevinson (commented):
Looks good.

Maybe i'm missing something, but how are we planning to encode the actual state? Don't we need a set of proto definitions for the actual storing of state for this module as well?

aaronc:
What do you mean by actual state? These are just raw bytes records.

proto/regen/data/v1alpha1/tx.proto
aaronc: No

aaronc: MsgRequest.GetSigners

proto/regen/data/v1alpha1/tx.proto
aaronc: It's on the tendermint block header. I assume it's deterministic if it's passed in the context

proto/regen/data/v1alpha1/tx.proto
aaronc: We don't need to. Version is specified by the CID format. I'm a bit torn on whether to actually adopt CID but it seems like a reasonable starting point.

proto/regen/data/v1alpha1/tx.proto
aaronc: They're in the `TC `

proto/regen/data/v1alpha1/tx.proto
aaronc: https://docs.tendermint.com/master/spec/consensus/bft-time.html

proto/regen/data/v1alpha1/events.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto3";
+
+package regen.data.v1alpha1;
+
+option go_package = "github.com/regen-network/regen-ledger/x/data";
+
+// EventAnchorData is an event emitted when data is anchored on-chain.
+message EventAnchorData {
```
amaurymartiny, line 8: it seems to me that these 3 events would be emitted exactly when their corresponding `Msg{*}Response` are emitted, which brings the question: why have both?

proto/regen/data/v1alpha1/tx.proto
```diff
@@ -0,0 +1,92 @@
...
+    //
+    // On-chain signatures have the following benefits:
+    // - on-chain identities can be managed using different cryptographic keys
+    //   that change over time through key rotation practices
+    // - an on-chain identity may represent an organization and through delegation
+    //   individual members may sign on behalf of the group
+    // - the blockchain transaction envelope provides built-in replay protection
+    //   and timestamping
+    //
+    // SignData implicitly calls AnchorData if the data was not already anchored.
```
amaurymartiny, line 27: SignData has multiple `signers`, and AnchorData only one `sender`. Who will be the `sender` in this case? `signers[0]`? If yes, it would be good to clarify it here.

proto/regen/data/v1alpha1/tx.proto
```diff
@@ -0,0 +1,92 @@
...
+    bytes cid = 2;
+}
+
+// MsgSignDataResponse is the Msg/SignData response type.
+message MsgSignDataResponse { }
+
+// MsgStoreDataRequest is the Msg/StoreData request type.
+message MsgStoreDataRequest {
+    // sender is the address of the sender of the transaction.
+    string sender = 1;
```
amaurymartiny, line 78: If the data is already anchored, the sender here can be different than the MsgAnchor's sender, right? (assuming cid corresponds to content)

amaurymartiny (commented):
lgtm. The `bytes content = 3` in `SignData` seems a bit generic, it reminds me of the Cosmos ADR `MsgSignedMessage`. But maybe for a start it's alright.

aaronc:
> lgtm. The `bytes content = 3` in `SignData` seems a bit generic, it reminds me of the Cosmos ADR `MsgSignedMessage`. But maybe for a start it's alright.

You mean `StoreData` I assume. It is _very_ generic... probably too generic and I originally tried to specify things more. But that's a longer process, I think I got in over my head and right now I just want to have the bare minimum ready to test to see what formats we can get users to adopt... And also that buys us more time to specify the right thing. Wdyt?

proto/regen/data/v1alpha1/tx.proto
aaronc: Exactly

proto/regen/data/v1alpha1/tx.proto
aaronc: We don't need a `sender` here. `sender` is just needed for `MsgRequest.GetSigners` in the other cases, it's not used otherwise. i.e. it's just needed to pay fees. Whereas `signers` is used for both `MsgRequest.GetSigners` and stored. I can improve the docs

proto/regen/data/v1alpha1/query.proto
```diff
@@ -0,0 +1,35 @@
...
+// QueryDataRequest is the Query/Data request type.
+message QueryDataRequest {
+
+    // cid is a Content Identifier for the data corresponding to the IPFS CID
+    // specification: https://github.com/multiformats/cid.
+    bytes cid = 1;
+}
+
+// QueryDataRequest is the Query/Data response type.
+message QueryDataResponse {
```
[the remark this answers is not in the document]
aaronc: `sender` doesn't do anything useful in this model. The `sender` could just be some third-party service acting as a middleman that timestamps data. `sender` did not "sign" or attest to the truthiness of the data. Maybe I need to clarify that about `Sign`

proto/regen/data/v1alpha1/events.proto
aaronc: Well they're not quite the same. These are actually a combination of the `Request` and `Response` types first of all. Also, `EventAnchorData` is _also_ emitted when `Sign` and `Store` are called if the data was not already anchored.

aaronc:
Updated the docs to address comments. Are we okay to move forward with this for now? Just to be transparent about my timing, it would be good to get an ACK in the next hour so I'm able to proceed with the next pieces before my end of day.

clevinson (approved):

amaurymartiny (approved):

PR #119, opened 2020-10-31
Eco-Credit Module Proof of Concept

aaronc:
ref: #78, #117 

Implements the `x/ecocredit` module as defined in #125. Read that PR description first for context. Module format and tests are experimental and follow what was done in #126. 

Proto breaking changes from #125:
* `issuer_address` reverted to `issuer` for consistency (feel free to improve docs where account address isn't clear)
* `EventReceive` updated based on https://github.com/regen-network/regen-ledger/pull/125/files#r517647028

---

Feel free to submit a Draft Pull Request as soon as you start working on something. Before you mark your PR as Ready For Review, make sure you've checked off the following boxes or indicated N/A (not applicable):
- [ ] linked to a relevant issue or have written a clear description of the issue in the PR
- [ ] wrote or updated tests
- [ ] wrote or updated documentation
- [ ] added an appropriate entry in `CHANGELOG.md` following https://keepachangelog.com

Thanks!

Issue #117, opened 2020-10-28
Balance/Denom credit design


aaronc:
## Summary

This proposes an architecture for ecosystem service credits (#107) based on the existing Cosmos SDK `x/bank` and `x/ibc/applications/transfer` `Coin` model.

## Problem Definition

See #107 

## Proposal

*This is based off of work done in https://github.com/regen-network/reDAOmint/tree/reDAOmint/x/ecocredit and updated to actually use an updated `x/bank` under the hood.*

In [ADR 001](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-001-coin-source-tracing.md) a convention for modules defining a "namespace" of denom identifiers is outlined.

We define a denom identifier for a credit batch as: `fmt.Sprintf("ecs/%s/%s", classId, batchId)` where `classId` are `batchId` unique string identifiers for the credit class and batch.

The `ecs` (ecosystem-service credit) module, defines the following `Msg`s for creating credit classes and batches, and retiring credits:

```proto
service Msg {
  rpc CreateClass(MsgCreateClass) returns (MsgCreateClassResponse);
  rpc CreateBatch(MsgCreateBatch) returns (MsgCreateBatchResponse);
  rpc Retire(MsgRetire) returns (MsgRetireResponse);
}

message MsgCreateClass {
  string designer;
  repeated string issuers;
  bytes metadata;
}

message MsgCreateClassResponse {
  string class_id;
}

message MsgCreateBatch {
  string issuer;
  string class_id;
  repeated BatchIssuance issuance;
}

message MsgCreateBatchResponse {
  string batch_denom;
}

message BatchIssuance {
  string recipient;
  string tradeable_units;
  string retired_units;
}

message MsgRetire {
  string holder;
  repeated cosmos.base.Coin credits;
}

message MsgRetireResponse { } 
```

### Pros
* integrates smoothly with existing Cosmos paradigm (`sdk.Coin`, `x/bank` and IBC transfer)

### Cons
* requires upgrades to `x/bank` to be performant (https://github.com/cosmos/cosmos-sdk/issues/7092) and elegant (https://github.com/cosmos/cosmos-sdk/issues/7093 and https://github.com/cosmos/cosmos-sdk/issues/7113).

### Neutral
* the `ecs` module will need to track balances of retired/burned coins

robert-zaremba:
[...]

Issue #115, opened 2020-10-21
Review UTXO based Credit Notes RFC


clevinson:
Review Credit Notes RFC from @robert-zaremba 

https://docs.google.com/document/d/14___kg4DWpd5QMWNKdCqFbMzUMac4qQo3GaSwMjZyi8/edit

aaronc:
So my high-level thought on all of this is that at some point we're going to need to interface with IBC transfer which uses balances and denoms. I'm not sure I see the concrete advantages of this UTXO approach over the existing account/balance approach Cosmos uses and if we do something different we will need to write the interoperability layer (i.e. IBC transfer).

Also, how exactly would the UTXO model provide more privacy? It seems like everything is still traceable without something like ZKP or Monero. Couldn't a method like that provide privacy for an account/balance model too?

I will summarize the current design in a separate issue so that it's easy to find and up to date, and we can compare.

aaronc:
See #117.

robert-zaremba:
Re privacy:
All efficient solutions I know use kind of a note / utxo system to achieve a privacy (even the zcash ZKP). 
If we limit only for amount hiding, then there are much simpler and very efficientmechanisms, like things related to tapproot or pedersen commitments. 

Using x/bank would probably require TEE (what Enigma is doing) or MPC like Nucypeher (which will be very slow). I'm not aware about other efficient trustless methods without escaping accounting system like x/bank.

robert-zaremba:
The benefit of the Credit Note is: 
+ allows easily to trance every specific coin. I could answer that this note comes from Aaron, who had it from Alice and Bob. With accounting I can only say that I received this coin from Aaron, but then I can't identify which coin Aaron is sending me (I can only say from which vintage, without knowing if if is sending me a coin from Alice, or from Nicole).
+ aforementioned privacy

For `x/banking` advantages are:
+ most probably easier implementation 
+ ready integration with IBC. 

I don't know how extending x/banking impacts the IBC compatibility nor how difficult it is to add it to a new module.

glandua:
It seems like the key question to enable a clear assessment of the Credit Note as an option is an assessment of how much effort would be needed for IBC compatibility.  How long would that assessment take?  @robert-zaremba?

robert-zaremba:
Shall we summarize the review and close the task?
[...]

PR #107, opened 2020-05-22
docs: add ecosystem service credit RFC + RFC template

clevinson:
ref #78 

## Proposed Changes
  - Adds Ecosystem Service Credit RFC from google doc to .md file for archiving & visiblity
  - Adds RFC folder
  - Adds template for future RFCs

Feel free to submit a Draft Pull Request as soon as you start working on something. Before you mark your PR as Ready For Review, make sure you've checked off the following boxes or indicated N/A (not applicable):
- [ ] linked to a relevant issue or have written a clear description of the issue in the PR
- [ ] wrote or updated tests
- [ ] wrote or updated documentation
- [ ] added an appropriate entry in `CHANGELOG.md` following https://keepachangelog.com

Thanks!

PR #125, opened 2020-11-04
Eco-credit Module Proof of Concept Init

aaronc:
ref #78 

## Credit Module Proto Files

This includes protobuf files for the design outlined in #117, but it does so *without* leveraging the existing bank module. Instead the implementation (#119) does its own balance and supply tracking (IMHO more correctly than `x/bank`). As I've said before, we can definitely consider alternate designs (like #115 or whatever we find is needed by users). This is just the design I've had for over a year now (and originally PoC'ed in SF in Oct, 2019). It was pretty simple to implement and I'd just like to get it out there so we can iterate from there.

## Decimal Balance Experiment

This design also includes an implementation of https://github.com/cosmos/cosmos-sdk/issues/7113 done with https://github.com/cockroachdb/apd. This feels relevant especially concerning the problems we're having with `sdk.Dec` (https://github.com/cosmos/cosmos-sdk/issues/7773. In particular, I wanted to try out `apd`, to see if it may be a suitable replacement for `sdk.Dec`.

---

Feel free to submit a Draft Pull Request as soon as you start working on something. Before you mark your PR as Ready For Review, make sure you've checked off the following boxes or indicated N/A (not applicable):
- [ ] linked to a relevant issue or have written a clear description of the issue in the PR
- [ ] wrote or updated tests

Thanks!


proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -0,0 +1,162 @@
...
+service Msg {
+
+    // CreateClass creates a new credit class with an approved list of issuers and optional metadata.
+    rpc CreateClass(MsgCreateClassRequest) returns (MsgCreateClassResponse);
+
+    // CreateBatch creates a new batch of credits for an existing credit class. This will create a new batch denom
+    // with a fixed supply. Issued credits can be distributed to recipients in either tradable or retired form.
+    rpc CreateBatch(MsgCreateBatchRequest) returns (MsgCreateBatchResponse);
+
+    // Send sends tradeable credits from one account to another account. Sent credits can either be or retired on receipt.
```
blushi, line 17:
```suggestion
    // Send sends tradeable credits from one account to another account. Sent credits can either be tradable or retired on receipt.
```

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -0,0 +1,30 @@
...
+    string batch_denom = 2;
+    string issuer = 3;
+    string total_units = 4;
+}
+
+message EventReceive {
+    string sender = 1;
+    string recipient = 2;
+    string batch_denom = 3;
+    string units = 4;
```
blushi, line 23: will this represent the total units received, whether there are tradable or retired?

aaronc: I think total units right?

aaronc:
Other than those comments @blushi does this look okay to move forward with? You honestly have the most context from the registry side.

proto/regen/ecocredit/v1alpha1/events.proto
aaronc: I'll update the field name

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -0,0 +1,107 @@
...
+    // batch_denom is the unique ID of credit batch to query.
+    string batch_denom = 1;
+}
+
+// QueryPrecisionResponse is the Query/Precision response type.
+message QueryPrecisionResponse {
+
+    // max_decimal_places is the maximum number of decimal places that can be used to represent some quantity of credit units.
+    // It is an experimental feature to concretely explore an idea proposed in https://github.com/cosmos/cosmos-sdk/issues/7113.
+    uint32 max_decimal_places = 1;
```
robert-zaremba, line 106: Two things to consider: 
1. Using natural numbers 
2. Using fixed decimals

Ad 1. Natural numbers are very fast and they works. There are very good libraries for doing 128-bit arithmetic. 128 bit max value is `3.4e38`. And even if we will have 1e9 denomination (like in bitcoin), we could easily multiply numbers up to 1e19 (1e10 denominated in 1e9) without doing any conversion. Otherwise we can easily convert to `big.Int` for more complex operations.

Ad 2. Having fixed decimals works very well with natural numbers and tooling. All tools will know that they need to put a comma at a certain position, without querying an asset. We could still limit the tradeable amounts by defining `granularity` - this is what ERC777 is doing (which is also implemented by OpenZeppelin).

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// MsgCreateClassRequest is the Msg/CreateClass request type.
+message MsgCreateClassRequest {
+
+    // designer is the address of the account which designed the credit class. The designer has special permissions
+    // to change the list of issuers and perform other administrative operations.
+    string designer = 1;
+
+    // issuers are the addresses of the approved issuers.
```
robert-zaremba, line 45:
```suggestion
    // issuers are the account addresses of the approved issuers.
```

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -0,0 +1,167 @@
...
+    repeated string issuers = 2;
+
+    // metadata is any arbitrary metadata to attached to the credit class.
+    bytes metadata = 3;
```
robert-zaremba, line 49: Do you think that there will be an `url` string in all classes metedata? If some, maybe it's worth to lift it up?

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -0,0 +1,167 @@
...
+    string issuer = 1;
+
+    // class_id is the unique ID of the class.
+    string class_id = 2;
+
+    // issuance are the credits issued in the batch.
+    repeated BatchIssuance issuance = 3;
+
+    // metadata is any arbitrary metadata to attached to the credit batch.
+    bytes metadata = 4;
```
robert-zaremba, line 72: There is one very important field in all issuance requests: a certificate. I would prefer to have it in the top scope, rather hidden in metadata. Certificate is basically a string with url or a proof. Probably we should even formalize it - and have additional service which will be able to prove the certificates. 

At minimum, I would add a comment with TODO to consider it.

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -0,0 +1,167 @@
...
+        // recipient is the account of the recipient.
+        string recipient = 1;
+
+        // tradable_units are the units of credits in this issuance that can be traded by this recipient.
+        // Decimal values are acceptable.
+        string tradable_units = 2;
+
+        // retired_units are the units of credits in this issuance that are effectively retired by the issuer on receipt.
+        // Decimal values are acceptable.
+        string retired_units = 3;
```
robert-zaremba, line 86: Why issuing _retired_ credits? We can keep it for flexibility, but I'm curious if this makes sense.

robert-zaremba: Also, shouldn't we use `oneof` for tradeable and retired units / credits?

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -0,0 +1,167 @@
...
+        // batch_denom is the unique ID of the credit batch.
+        string batch_denom = 1;
+
+        // tradable_units are the units of credits in this issuance that can be traded by this recipient.
+        // Decimal values are acceptable within the precision returned by Query/Precision.
+        string tradeable_units = 2;
+
+        // retired_units are the units of credits in this issuance that are effectively retired by the issuer on receipt.
+        // Decimal values are acceptable within the precision returned by Query/Precision.
+        string retired_units = 3;
```
robert-zaremba, line 121: As above, retired units are not tradeable - so doesn't it make sense to send it over?

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+// MsgSendResponse is the Msg/Send response type.
+message MsgSendResponse { }
+
+// MsgRetireRequest is the Msg/Retire request type.
+message MsgRetireRequest {
+
+    // holder is the address of the holder of the credits.
```
robert-zaremba, line 131:
```suggestion
    // holder is the credits holder address.
```

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -0,0 +1,167 @@
...
+    string holder = 1;
+
+    // credits are the credits being retired.
+    repeated RetireUnits credits = 2;
+
+    // RetireUnits are the units of the batch being retired.
```
robert-zaremba, line 137: this description should use the same language as the comment in line 134. Also, related to the comment above about changing a vocabulary for units (in proto/regen/ecocredit/v1alpha1/events.proto)

proto/regen/ecocredit/v1alpha1/types.proto
```diff
@@ -0,0 +1,40 @@
...
+    string batch_denom = 2;
+
+    // issuer is the issuer of the credit batch.
+    string issuer = 3;
+
+    // total_units is the total number of units in the credit batch and is immutable.
+    string total_units = 4;
+
+    // metadata is any arbitrary metadata to attached to the credit batch.
+    bytes metadata = 5;
```
robert-zaremba, line 39: related to the comment above - let's consider adding a `certificate` field.

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -0,0 +1,75 @@
...
+    // the case that credits were retired upon issuance using Msg/CreateBatch or retired upon transfer
+    // using Msg/Send.
+    string retirer = 1;
+
+    // batch_denom is the unique ID of credit batch.
+    string batch_denom = 2;
+
+    // units is the decimal number of credits that have been retired.
+    string units = 3;
+}
```
robert-zaremba, line 75: We should also add a Claim Subject or Claim Reference - which is used to link the credits retire events with real world event.

robert-zaremba (changes requested):

robert-zaremba:
> we can definitely consider alternate designs (like #115 or whatever we find is needed by users).

This model is almost compatible with #115 design (with small changes to ` Msg/Send` and ` Msg/Retire`).

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -0,0 +1,75 @@
...
+
+    // designer is the designer of the credit class.
+    string designer = 2;
+}
+
+// EventCreateBatch is an event emitted when a credit batch is created.
+message EventCreateBatch {
+
+    // class_id is the unique ID of credit class.
+    string class_id = 1;
```
clevinson, line 21: What is the reason that these identifiers (class_id) are string and not integer?

I guess for `batch_denom` its useful to have as a string, as that's the same type as denoms in the broader cosmos sdk, but for `class_id` how do you see these ID's being generated? In my head, this would have been an integer identifier, indexed (as in a relational database) and auto-incrementing

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -0,0 +1,75 @@
...
+
+    // batch_denom is the unique ID of credit batch.
+    string batch_denom = 2;
+
+    // issuer is the issuer of the credit batch.
+    string issuer = 3;
```
clevinson, line 27:
```suggestion
    // issuer_address is the account address of the issuer of the credit batch.
    string issuer_address = 3;
```
Can we be more explicit that this represents an account address? Or do you think that is unecessary detail?

clevinson: If we don't make it explciit in the field, i think it'd be nice to atleast make it more explicit in the comment.

proto/regen/ecocredit/v1alpha1/tx.proto
clevinson, line 121: Seems strange to allow sending of pre-retired credits. IMO retiring should be a separate action done intentionally by the holder of credits.

We already cover the "retire on issuance" use case in the `BatchIssuance` message, so why is this needed here as well?

clevinson (approved):
Few comments on the retiring functionality, and whether users can transfer retired credits with this implementation, or what the purpose is of having `retired_units` as part of the `SendUnits` msg.

Happy for these to be resolved in a followup and use this as a starting point though.

proto/regen/ecocredit/v1alpha1/events.proto
clevinson: Actually... noticing how we use "issuer" "recipient" "holder" "sender" throughout. I don't mean to push for a change to all those field names, so maybe we can just resolve this within comments (adding an explicit mention to account addresses) and leave the actual field names unchanged.

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -0,0 +1,75 @@
...
+option go_package = "github.com/regen-network/regen-ledger/x/ecocredit";
+
+// EventCreateClass is an event emitted when a credit class is created.
+message EventCreateClass {
+
+    // class_id is the unique ID of credit class.
+    string class_id = 1;
+
+    // designer is the designer of the credit class.
+    string designer = 2;
```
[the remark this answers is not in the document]
aaronc: This is a market term agreed on with the registry term. Out of scope of this PR

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// EventCreateBatch is an event emitted when a credit batch is created.
+message EventCreateBatch {
+
+    // class_id is the unique ID of credit class.
+    string class_id = 1;
+
+    // batch_denom is the unique ID of credit batch.
+    string batch_denom = 2;
```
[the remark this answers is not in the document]
aaronc: Yep I can add some clarifying language on denominations

proto/regen/ecocredit/v1alpha1/events.proto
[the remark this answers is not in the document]
aaronc: yep

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -0,0 +1,75 @@
...
+    // sender (in the case of Msg/Send).
+    oneof from {
+        // issuer is the issuer of the credit batch in the case that this event is the result of a new batch being created.
+        // It will not be set if credits were transferred.
+        string issuer = 1;
+
+        // sender is the sender of the credits in the case that this event is the result of a transfer.
+        // It will not be set if credits were issued.
+        string sender = 2;
+    }
```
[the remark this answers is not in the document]
aaronc: I don't have to use oneof true, but they are mutually exclusive. I can just make `sender = ""` if that's preferred

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -0,0 +1,75 @@
...
+
+    // recipient is the recipient of the credits
+    string recipient = 3;
+
+    // batch_denom is the unique ID of credit batch.
+    string batch_denom = 4;
```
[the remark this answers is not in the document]
aaronc: yep

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -0,0 +1,75 @@
...
+    string class_id = 1;
+
+    // batch_denom is the unique ID of credit batch.
+    string batch_denom = 2;
+
+    // issuer is the issuer of the credit batch.
+    string issuer = 3;
+
+    // total_units is the total number of units in the credit batch.
+    string total_units = 4;
```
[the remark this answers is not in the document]
aaronc: for now I'm trying to stay consistent with the RFC that uses the term `units` and went through a review process. I'd rather this be addressed in the RFC PR #107

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -0,0 +1,107 @@
...
+    // retired_supply is the decimal number of retired units in the batch supply.
+    string retired_supply = 2;
+}
+
+
+// QueryPrecisionRequest is the Query/Precision request type.
+message QueryPrecisionRequest {
+
+    // batch_denom is the unique ID of credit batch to query.
+    string batch_denom = 1;
```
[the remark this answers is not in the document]
aaronc: not necessarily, they correspond to different projects which have wildly different sizes

proto/regen/ecocredit/v1alpha1/query.proto
aaronc: Under the hood this is basically `big.Int`. It simply allows configurable granularity so that doesn't need to be decided up front. It would be good to have these discussions in the context of https://github.com/cosmos/cosmos-sdk/issues/7113. This is simply intended to show a viable PoC pathway towards that

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: likely, but here I'm trying to postpone dealing with data formats and just deal with asset mechanics. I'm assuming people will put some json here so this basically is the decision to start schema-less and iterate

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: let's open a separate issue to discuss metadata. it's a big topic that I'd rather specify off-chain first and then put on-chain. we would need to discuss with Ron on the customer side

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: It's a fundamental part of the use case. It's not mutually exclusive either. You could issue both. In some cases it is desirable or even required to issue credits as retired (i.e. non-tradable) on creation

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: Not sure, but likely yes. A broker may specifically sell credits that can't be traded as part of the contract. Again a business discussion to have with Ron

proto/regen/ecocredit/v1alpha1/events.proto
aaronc: An integer converted to a string and so that it's obviously part of denom. We can reassess later

proto/regen/ecocredit/v1alpha1/events.proto
aaronc: we should update the doc string not the field name IMHO

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: Not necessarily. I believe a broker may specifically only sell credits that are retired on sale. We need to discuss with Ron It may or may not be a use case but better to be on the safe side

proto/regen/ecocredit/v1alpha1/types.proto
aaronc: again let's discuss metadata elsewhere. we don't have context on the engineering team to make these sorts of decisions

proto/regen/ecocredit/v1alpha1/events.proto
aaronc: ditto, this should be part of RFC process

proto/regen/ecocredit/v1alpha1/events.proto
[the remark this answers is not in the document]
robert-zaremba: OK. fair enough.

proto/regen/ecocredit/v1alpha1/events.proto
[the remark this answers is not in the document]
robert-zaremba: The whole thing is to eliminate the`oneof` and normalize the data in the RDBMS sense (so to not use the `issuer` field at all here). Adding a boolean flag is  only cosmetic if we want to be explicit.

robert-zaremba (approved):
Looks good to move forward. I will collect the questions / suggestion in an issue for review by a science and business team.

aaronc:
> Looks good to move forward. I will collect the questions / suggestion in an issue for review by a science and business team.

thanks @robert-zaremba

proto/regen/ecocredit/v1alpha1/events.proto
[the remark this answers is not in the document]
aaronc: well i'm trying to distinguish between issuer and sender as roles...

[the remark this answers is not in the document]
robert-zaremba: I know. My point is that `issuer` is known and it must be the same as the batch issuer. By adding it here we denormalize the data.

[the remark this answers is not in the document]
aaronc: Right, so I'm removing and just leaving `sender = ""`

proto/regen/ecocredit/v1alpha1/events.proto
blushi: sounds good.

proto/regen/ecocredit/v1alpha1/events.proto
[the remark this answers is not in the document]
blushi: @robert-zaremba see https://regen-registry.s3.amazonaws.com/GHG+and+Co-Benefits+in+Grazing+Systems+Credit+Class.pdf - "Definitions" section for reference (available from https://www.regen.network/resources)

proto/regen/ecocredit/v1alpha1/tx.proto
blushi: Indeed, there are cases where sold credits should be "auto-retired".

blushi: agreed cf. https://github.com/regen-network/regen-ledger/pull/125#discussion_r517845025

PR #153, opened 2020-11-05
Integrate ecocredits module

robert-zaremba:
Closes:  #134

- Created Server interface
- adding server to app module
- implementing Route


x/ecocredit/server/query_server.go
```diff
@@ -22,31 +22,22 @@ func (s serverImpl) ClassInfo(ctx context.Context, request *ecocredit.QueryClass
 func (s serverImpl) getClassInfo(ctx sdk.Context, classID string) (*ecocredit.ClassInfo, error) {
 	var classInfo ecocredit.ClassInfo
 	err := s.classInfoTable.GetOne(ctx, orm.RowID(classID), &classInfo)
-	if err != nil {
-		return nil, err
-	}
-
-	return &classInfo, nil
+	return &classInfo, err
```
aaronc, line 25: This feels like an unnecessary change. Result should be nil if there's an error

x/ecocredit/server/query_server.go
```diff
@@ -22,31 +22,22 @@ func (s serverImpl) ClassInfo(ctx context.Context, request *ecocredit.QueryClass
...
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
 	var batchInfo ecocredit.BatchInfo
 	err := s.batchInfoTable.GetOne(ctx, orm.RowID(request.BatchDenom), &batchInfo)
-	if err != nil {
-		return nil, err
-	}
-
-	return &ecocredit.QueryBatchInfoResponse{Info: &batchInfo}, nil
+	return &ecocredit.QueryBatchInfoResponse{Info: &batchInfo}, err
```
aaronc, line 33: Ditto

x/ecocredit/server/server.go
```diff
@@ -25,14 +25,31 @@ type serverImpl struct {
...
-
+	idSeq          orm.Sequence
 	classInfoTable orm.NaturalKeyTable
-
 	batchInfoTable orm.NaturalKeyTable
 }
 
-func newServer(storeKey sdk.StoreKey) serverImpl {
+// Server is the the ecocredits implementation of ADR 031 Msg Service
+type Server interface {
```
aaronc, line 34: Not sure this is desirable. I was trying to not expose this outside of RegisterServices. If we need an interface just inherit MsgServer and QueryServer rather than writing out each method.

aaronc (commented):
Mostly looks okay but left a few comments

x/ecocredit/server/query_server.go
robert-zaremba: Why? If there is an error, then the result by definition in Go SDK is not specified. 
I find it not useful to have the additional 3-4 lines of code just of an err check when the error can be returned directly. I found that google and other Go companies are using this patter across different libraries. Here example in Std lib: https://golang.org/src/strconv/atof.go  line 559.

x/ecocredit/server/server.go
robert-zaremba: It helps with testing. The pattern I'm following is to return interfaces wherever possible and following the excellent proverb: 

_“Accept interfaces, return structs”_

There are many benefits from it, especially if we design different components and will like to do unit-tests. It's also part of SOLID principles.

robert-zaremba: But you are right, we should reuse `MsgServer` and `QueryServer`

aaronc: Right I was just wanting to make the serverimpl private and force people to use clients

aaronc: The tests all use clients so they don't need this public

aaronc: Just for context I actually did exactly what you're doing first btw 😉 but then changed to make it private

aaronc:
Is it useful to do all this stylistic refactoring? I thought the goal was to wire things up and add cli commands not just reorganize stuff a bit and honestly I don't see how this reorganization improves things.

x/ecocredit/module/module.go
```diff
@@ -3,27 +3,33 @@ package module
...
+	"github.com/spf13/cobra"
+	abci "github.com/tendermint/tendermint/abci/types"
 
 	"github.com/regen-network/regen-ledger/x/ecocredit"
 	"github.com/regen-network/regen-ledger/x/ecocredit/server"
 )
 
 type AppModule struct {
 	Key sdk.StoreKey
+	Srv server.Server
```
aaronc, line 22: What value is there in adding this here? Where is it going to need to be publicly used?

x/ecocredit/module/module.go
```diff
@@ -59,16 +65,19 @@ func (a AppModule) ExportGenesis(sdk.Context, codec.JSONMarshaler) json.RawMessa
...
 
 func (a AppModule) LegacyQuerierHandler(*codec.LegacyAmino) sdk.Querier {
 	return nil
 }
 
-func (a AppModule) RegisterServices(configurator module.Configurator) {
-	server.RegisterServices(a.Key, configurator)
+func (a AppModule) RegisterServices(cfg module.Configurator) {
+	ecocredit.RegisterMsgServer(cfg.MsgServer(), a.Srv)
+	ecocredit.RegisterQueryServer(cfg.QueryServer(), a.Srv)
```
aaronc, line 80: Again why is this necessary?

x/ecocredit/server/server_test.go
```diff
@@ -3,20 +3,26 @@ package server
...
+
 func TestServer(t *testing.T) {
-	key := sdk.NewKVStoreKey("ecocredit")
+	key := sdk.NewKVStoreKey(ecocredit.ModuleName)
 	addrs := configurator.MakeTestAddresses(6)
-	configuratorFixture := configurator.NewFixture(t, []sdk.StoreKey{key}, addrs)
-	RegisterServices(key, configuratorFixture)
-	s := testsuite.NewIntegrationTestSuite(configuratorFixture)
+	cfg := configurator.NewFixture(t, []sdk.StoreKey{key}, addrs)
+	registerServers(key, cfg)
```
aaronc, line 24: Ditto why refactor this? What are you trying to accomplish?

x/ecocredit/module/module.go
```diff
@@ -59,16 +65,19 @@ func (a AppModule) ExportGenesis(sdk.Context, codec.JSONMarshaler) json.RawMessa
 
 func (a AppModule) RegisterInvariants(sdk.InvariantRegistry) {}
 
-func (a AppModule) Route() sdk.Route { return sdk.Route{} }
+func (a AppModule) Route() sdk.Route {
+	return sdk.Route{}
+}
 
-func (a AppModule) QuerierRoute() string { return "" }
+func (a AppModule) QuerierRoute() string { return a.Name() }
```
aaronc, line 72: Unneeded. This is legacy stuff

x/ecocredit/module/module.go
robert-zaremba: I wrote about it on [Discord]() and in the commit message: https://github.com/regen-network/regen-ledger/pull/153/commits/ff2a9036c68ee5fd9b96d0b2a9dcfe46c8e392f9

The whole idea is to allow modules to make it dependency: other modules can mock a server if needed.

x/ecocredit/module/module.go
robert-zaremba: Let's discuss in the comment above.

x/ecocredit/module/module.go
robert-zaremba: In fact the main use-case was to use them in handlers. Since we don't need to implement a Router, then it has less utility -  app module doesn't need to hold the server instance any more (at least I don't see other methods where it's needed, `RegisterGRPCGatewayRoutes` also doesn't use the server).

The only reason I to keep it here is to be able to customize the server within the `AppModule.RegisterServices`. Since the AppModule has much less utility now, then probably this is not needed. We can parametrize the package `ecocredits/RegisterServices` method (note, it's not in the server anymore, and it will take Server as an argument) and  external tests can use that method instead of AppModule. What do you think?

aaronc: Right, the app module doesn't need a server instance at all with this design and that's intentional. Maybe I need to document the reasons more but here's my thinking:
1. other modules shouldn't hold "keepers"/servers directly (according to the WIP ADR 033), instead they use clients which handle permissions correctly. So a direct server instance is undesirable because it breaks ocaps/encapsulation
2. an `AppModule` can eventually become sort of a "module descriptor" which instantiates module services but doesn't maintain any of its own state. Module descriptors would _only_ contain config parameters. In `RegisterServices`, services are constructed and registered but not owned by the app module - rather ownership is transfered to the various routers/service managers.

 My preference right now is to leave `RegisterServices` in `server` as it was. I originally had things the way you have them here, but I changed it because that felt like it had better encapsulation. Basically with `RegisterServices` in `server`, tests don't need to know how services are getting implemented or ever have a concrete server implementation - all of that is hidden so there is proper ocaps provided by the clients.

Does that make sense?

In general I'm leaning in the direction of AppModule being much leaner and tests would mainly need `RegisterInterfaces` and `RegisterServices` instead of all the other AppModule stuff currently there.

aaronc: I'm thinking that a better, less high coupled approach might be something like:
- what is `types` in the SDK and now the root in this approach just has types and maybe something like a `ModuleDescriptor` with just `RegisterInterfaces`
- `server` has a lightweight `ServerModule` which just has `RegisterServices` + `RegisterInterfaces`
- there are separate client modules wired up separately like `CLIClientModule`, `GRPCGatewayClientModule`, etc.

This has a dependency structure where `x/{module}/server` and `x/{module}/client` depend on `x/{module}` and app.go only imports `x/{module}/server.ServerModule`, and we don't have everything like clients and simulations getting imported into app.go.

x/ecocredit/module/module.go
robert-zaremba: moved to a deprecated section of the file.

robert-zaremba:
I was planning to merge do CLI in a separate PR. I added Query CLI implementation here, the Tx CLI is work in progress.

x/ecocredit/server/server.go
robert-zaremba: Thanks for the explanation. BTW: interface, not an implementation. But probably you are right - with the comments above, we won't need to expose anything about the server (neither the combined interface). I will look at it tomorrow again, meanwhile I added CLI code (query is done, Tx only one command).

x/ecocredit/module/module.go
robert-zaremba: This makes sense. I like that. Although I would keep the `AppModule` name - it describes it's goal very well. `x/{module}` shouldn't import `x/{module}/subpackage`.

x/ecocredit/module/module.go
```diff
@@ -36,43 +39,50 @@ func (a AppModule) DefaultGenesis(codec.JSONMarshaler) json.RawMessage {
...
+func (a AppModule) EndBlock(sdk.Context, abci.RequestEndBlock) []abci.ValidatorUpdate {
 	return nil
 }
 
-func (a AppModule) RegisterServices(configurator module.Configurator) {
-	server.RegisterServices(a.Key, configurator)
+/**** DEPRECATED ****/
+
+// Should we deprecate it? RegisterServices already registers everything for the module.
+func (a AppModule) RegisterInterfaces(codectypes.InterfaceRegistry) {}
```
aaronc, line 74: No it's definitely not deprecated. Rather it's actually rather new.

x/ecocredit/client/query.go
```diff
@@ -0,0 +1,126 @@
...
+			return print(ctx, res, err)
+		},
+	}
+}
+
+func queryBatchInfo() *cobra.Command {
+	return &cobra.Command{
+		Use:   "batch_info [batch_denom]",
+		Short: "Select the credit issuance batch info",
+		Long:  "Select the credit issuance batch info based on the bach_denom (ID)",
```
aaronc, line 56: ditto

x/ecocredit/client/query.go
```diff
@@ -0,0 +1,126 @@
...
+			return print(ctx, res, err)
+		},
+	}
+}
+
+func queryBalance() *cobra.Command {
+	return &cobra.Command{
+		Use:   "balance [batch_denom] [account]",
+		Short: "Select the tradable and retired balance of the credit batch",
+		Long:  "Select the credit tradable and retired balance of the credit batch denom (ID) and account address",
```
aaronc, line 75: ditto

x/ecocredit/client/query.go
```diff
@@ -0,0 +1,126 @@
...
+			return print(ctx, res, err)
+		},
+	}
+}
+
+func queryPrecision() *cobra.Command {
+	return &cobra.Command{
+		Use:   "precision [batch_denom]",
+		Short: "Select the maximum length of the fractional part of credits in the given batch",
+		Long:  "Select the maximum length of the fractional part of credits in the given batch. The precision tells what is the minimum unit of a credit.\nExample: a decimal number 12.345 has fractional part length equal 3. A precision=5 means that the minimum unit we can trade is 0.00001",
```
aaronc, line 113: ditto

x/ecocredit/client/query.go
```diff
@@ -0,0 +1,126 @@
...
+			return print(ctx, res, err)
+		},
+	}
+}
+
+func querySupply() *cobra.Command {
+	return &cobra.Command{
+		Use:   "supply [batch_denom]",
+		Short: "Select the tradable and retired supply of the credit batch",
+		Long:  "Select the tradable and retired supply of the credit batch batch denom (ID)",
```
aaronc, line 94: ditto

aaronc (commented):
The biggest issue is updating the usage strings of the cmd's. Those need just a bit of grammar editing.

I have different ideas about module organization, but I can maybe demonstrate that in a PoC. Otherwise, generally looks okay. Thanks @robert-zaremba .

x/ecocredit/module/module.go
aaronc: Tests are actually failing because the msg service descriptor needs to be wired up there

x/ecocredit/module/module.go
amaurymartiny: This makes sense to me too. Some comments:
- why not call it `ModuleServer` instead of `ServerModule`: it's literally a module's server.
- I'm not sure why we have `RegisterInterfaces` in both `ServerModule` and `ModuleDescriptor`. Only once should be enough? My preference goes to the root package.
- While `CLIClientModule` and `GRPCGatewayClientModule` makes things feel more organized, consistent and structured, it also seems overdone, as e.g. `GRPCGatewayClientModule` afaics will only ever have one line like:
```golang
ecocredits.RegisterQueryHandlerClient(context.Background(), mux, ecocredits.NewQueryClient(clientCtx))
````

Comment (robert-zaremba) [reply x/ecocredit/module/module.go:23]: The idea is not to expose the Server interfaces (I'm removing that interface), and the AppModule / ServerModule is not a server - it provides module information / descriptors.

File: x/ecocredit/module/module.go (hunk 8 above)
Comment (robert-zaremba) [reply x/ecocredit/module/module.go]: yes, thanks for discussion in Discord. I added the implementation for it and added AppModuleBasic.

File: x/ecocredit/module/module.go [hunk 13]
```diff
@@ -48,18 +44,24 @@ func (a AppModule) ExportGenesis(sdk.Context, codec.JSONMarshaler) json.RawMessa
...
+}
 
 func (a AppModule) RegisterGRPCGatewayRoutes(sdkclient.Context, *runtime.ServeMux) {}
 
 func (a AppModule) RegisterInvariants(sdk.InvariantRegistry) {}
 
 func (a AppModule) RegisterServices(cfg module.Configurator) {
-	ecocredit.RegisterMsgServer(cfg.MsgServer(), a.Srv)
-	ecocredit.RegisterQueryServer(cfg.QueryServer(), a.Srv)
+	key := sdk.NewKVStoreKey(ecocredit.ModuleName)
```
aaronc, line 69: The key cannot be created in `RegisterServices`. Then it won't get mounted with the multistore.

aaronc (approved):

PR #154, opened 2020-11-05
Add Group module

### LINKED ISSUE #136: Add groups module to regen ledger and wire up app.go
Add groups module to regen ledger. ORM dependency should be updated to depend on the `x/orm` package already in regen ledger.

The module should also be wired up in app.go.

### LINKED ISSUE #145: Finish the Group Module migration to Any
Ref: (original issue) https://github.com/regen-network/cosmos-modules/issues/69

### LINKED ISSUE #198: [x/group] Add query service
ref: https://github.com/regen-network/cosmos-modules/issues/42, https://github.com/regen-network/cosmos-modules/pull/63

blushi:
Closes #136
Closes #145
Closes #198
Closes https://github.com/regen-network/cosmos-modules/issues/69
Closes https://github.com/regen-network/cosmos-modules/issues/72

## Proposed Changes

- [x] Migrate group module from https://github.com/regen-network/cosmos-modules/pull/71.
- [x] Put proto definitions into separate files: `types.proto`, `tx.proto` and `genesis.proto` (no queries implemented for now).
- [x] Add events types (`events.proto`) for group create/update and group account create/update.
- [x] Migrate to use `Msg` service.
- [x] Keep using `sdk.Router` for now for the PoC. Further down the road we'll want to replace it with ADR 033 approach (https://github.com/regen-network/regen-ledger/pull/157).
- [x] Refactor tests to use `Msg` service.
- [x] Update ORM to use `Marshaler` to unpack `Any`s properly ([`GetCachedValue()` returned `nil` for `Any`s returned from the ORM](https://github.com/regen-network/regen-ledger/pull/154/checks?check_run_id=1363429686)).
- [x] Wire up group module in `app.go`.
- [x] #198 Implement Query service.

## Follow-ups

- #197 Add integration tests.
- #209 Add CLI support.
- Currently, we are using different endpoints for updating a group whether we wanna update its members, admin or comments. It would be better UX-wise to just have one update endpoint if possible.
- Implement group account update #224
- Add other event types #215
- Update `GenesisState` #214
- As described above, we're still using `sdk.Router` to route messages to their handlers. This means that we cannot use the group module right now to route messages from the `data` or `ecocredit` modules, unless we implement some kind of temporary handlers for them. Eventually, the group module should follow [ADR 033](https://github.com/cosmos/cosmos-sdk/pull/7459) approach (each group account would be a sub module account according to [ADR 028](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-028-public-key-addresses.md), then using the group’s `ModuleKey`, the proposal handler would execute messages using `ModuleKey.Invoke`).
- Replace `x/group/types/conditions.go` with ADR 028 #211

- Remaining [issues from cosmos-modules repo](https://github.com/regen-network/cosmos-modules/labels/Group%20module) should be sorted priority-wise and migrated to this repo @clevinson. 

Feel free to submit a Draft Pull Request as soon as you start working on something. Before you mark your PR as Ready For Review, make sure you've checked off the following boxes or indicated N/A (not applicable):
- [x] linked to a relevant issue or have written a clear description of the issue in the PR
- [x] wrote or updated tests
- [x] wrote or updated documentation

Thanks!


testutil/testdata/group/codec.go
```diff
@@ -0,0 +1,32 @@
+package group
+
+import (
+	"github.com/cosmos/cosmos-sdk/codec"
+	cryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
+)
+
+// RegisterLegacyAminoCodec registers all the necessary module concrete types and
+// interfaces with the provided codec reference.
+func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
```
aaronc, line 10: Can we just delete all the amino stuff?

x/group/conditions.go
```diff
@@ -0,0 +1,126 @@
+package group
```
aaronc, line 1: This needs to be replaced with ADR 028 eventually

aaronc (commented):
Looks like a lot of stuff in testdata, maybe unnecessarily?

app/app.go
```diff
@@ -84,11 +86,12 @@ import (
...
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
-
-	"github.com/CosmWasm/wasmd/x/wasm"
-	wasmclient "github.com/CosmWasm/wasmd/x/wasm/client"
+	regenmodule "github.com/regen-network/regen-ledger/types/module"
 	data "github.com/regen-network/regen-ledger/x/data/module"
 	ecocredit "github.com/regen-network/regen-ledger/x/ecocredit/module"
+	group "github.com/regen-network/regen-ledger/x/group/module"
+	groupserver "github.com/regen-network/regen-ledger/x/group/server"
+	grouptypes "github.com/regen-network/regen-ledger/x/group/types"
```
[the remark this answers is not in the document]
blushi: For `group` and `grouptypes` aliases, I'm just following the same format that is used for other modules in here. For `groupserver`, we could actually remove the alias.

go.mod
```diff
@@ -14,7 +14,9 @@ require (
 	github.com/ipfs/go-cid v0.0.7
 	github.com/lib/pq v1.8.0 // indirect
 	github.com/multiformats/go-multihash v0.0.14
+	github.com/pkg/errors v0.9.1
```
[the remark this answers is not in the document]
blushi: we can indeed remove it in favor of cosmos sdk errors, some leftovers from the previous implementation, thanks for catching that!

go.mod
```diff
@@ -14,7 +14,9 @@ require (
...
 	github.com/rakyll/statik v0.1.7
+	github.com/regen-network/cosmos-proto v0.3.0
```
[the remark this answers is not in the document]
blushi: we need it for annotations in proto files for `Any`s (eg `cosmos_proto.accepts_interface`)

orm/genesis.go
```diff
@@ -93,14 +94,14 @@ func ImportTableData(ctx HasKVStore, t TableExportable, src json.RawMessage, seq
 
 // forEachInTable iterates through all entries in the given table and calls the callback function.
 // Aborts on first error.
-func forEachInTable(ctx HasKVStore, table Table, f func(RowID, Persistent) error) error {
+func forEachInTable(ctx HasKVStore, table Table, f func(RowID, codec.ProtoMarshaler) error) error {
```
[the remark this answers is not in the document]
blushi: Before we were using `Marshal` and `Unmarshal` methods from a `Persistent` object but this is not sufficient to unmarshal `Any`s (see details in the PR description above), so now we're using [`Marshaler`](https://github.com/cosmos/cosmos-sdk/blob/320a852ee268bf806baa0a667098442a83da6361/codec/codec.go) instead which needs `codec.ProtoMarshaler` for binary encoding/decoding).

proto/regen/group/v1alpha1/types.proto
```diff
@@ -0,0 +1,262 @@
...
+    // CHOICE_VETO defines a voting choice with veto.
+    CHOICE_VETO = 4;
+}
+
+//
+// State
+//
+
+// GroupMetadata represents a group's metadata
+message GroupMetadata {
```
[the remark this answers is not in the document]
blushi: separately from what? This represents the data stored in an ORM table for a group.

proto/regen/group/v1alpha1/types.proto
```diff
@@ -0,0 +1,262 @@
...
+
+    // Result defines types of proposal results.
+    enum Result {
+        option (gogoproto.goproto_enum_prefix) = false;
+
+        // An empty value is invalid and not allowed
+        RESULT_UNSPECIFIED = 0 [(gogoproto.enumvalue_customname) = "ProposalResultInvalid"];
+        
+        // Until a final tally has happened the status is undefined
+        RESULT_UNDEFINED = 1 [(gogoproto.enumvalue_customname) = "ProposalResultUndefined"];
```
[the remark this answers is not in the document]
blushi: not really, `RESULT_UNSPECIFIED` 0 is just a result value that would make the proposal invalid while `RESULT_UNDEFINED` is valid and means that the proposal result is not defined because a final tally hasn't happened yet.

x/group/types/conditions.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/regen-network/regen-ledger/orm"
+)
+
+var (
+	// AddressLength is the length of all addresses
+	// You can modify it in init() before any addresses are calculated,
+	// but it must not change during the lifetime of the kvstore
+	AddressLength = 20
+
+	// it must have (?s) flags, otherwise it errors when last section contains 0x20 (newline)
```
robert-zaremba, line 23: regular expression is not a function it doesn't return an error. 
1. What does `perm` mean?
2. let's update a comment and explain what we want to find or match.

robert-zaremba: OK, I found later that `perm` is permission. Let's describe it in a comment.

x/group/types/conditions.go
```diff
@@ -0,0 +1,132 @@
...
+//   sprintf("%s/%s/%s", extension, type, data)
+type Condition []byte
+
+func NewCondition(ext, typ string, data []byte) Condition {
+	pre := fmt.Sprintf("%s/%s/", ext, typ)
+	return append([]byte(pre), data...)
+}
+
+// Parse will extract the sections from the Condition bytes
+// and verify it is properly formatted
```
robert-zaremba, line 40: Please update a comment: we extract a permission section.

x/group/types/conditions.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return nil
+}
+
+func (c Condition) MarshalJSON() ([]byte, error) {
+	var serialized string
+	if c != nil {
+		serialized = c.String()
+	}
+	return json.Marshal(serialized)
```
robert-zaremba, line 88:
```suggestion
if c == nil {
    return []byte(`""`)
}
return json.Marshal(c.String())
```

x/group/types/conditions.go
```diff
@@ -0,0 +1,132 @@
...
+	// it must have (?s) flags, otherwise it errors when last section contains 0x20 (newline)
+	perm = regexp.MustCompile(`(?s)^([a-zA-Z0-9_\-]{3,8})/([a-zA-Z0-9_\-]{3,8})/(.+)$`)
+)
+
+// Condition is a specially formatted array, containing
+// information on who can authorize an action.
+// It is of the format:
+//
+//   sprintf("%s/%s/%s", extension, type, data)
+type Condition []byte
```
robert-zaremba, line 32: Why don't we we represent a condition using a structure with 3 fields: extension, type, data? We can have 2 methods: `String()` and `MarshalJSON`, and one additional constructor `ParseCondition(str)`

x/group/types/proposal.go
```diff
@@ -0,0 +1,102 @@
+package types
+
+import (
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+func (p *Proposal) GetMsgs() []sdk.Msg {
```
robert-zaremba, line 9:
```suggestion
func (p *Proposal) CreateMesages() []sdk.Msg {
```

robert-zaremba: or: MkMessages, MakeMessages...

x/group/types/msgs_test.go
```diff
@@ -0,0 +1,350 @@
...
+	assert.Equal(t, []sdk.AccAddress{myAddr}, MsgCreateGroupRequest{Admin: myAddr}.GetSigners())
+}
+
+func TestMsgCreateGroupAccount(t *testing.T) {
+	_, _, myAddr := testdata.KeyTestPubAddr()
+
+	specs := map[string]struct {
+		admin     sdk.AccAddress
+		group     GroupID
+		threshold sdk.Dec
```
robert-zaremba, line 127: We don't want to use `sdk.Dec`. Instead, in Regen we want to use `apd.Decimal`. (see ecocredits module)
ref: https://github.com/cosmos/cosmos-sdk/issues/7773

robert-zaremba: Same for weight

x/group/types/types.go
```diff
@@ -0,0 +1,380 @@
...
+	Final bool
+}
+
+// DecisionPolicy is the persistent set of rules to determine the result of election on a proposal.
+type DecisionPolicy interface {
+	codec.ProtoMarshaler
+
+	orm.Validateable
+	GetTimeout() types.Duration
+	Allow(tally Tally, totalPower sdk.Dec, votingDuration time.Duration) (DecisionPolicyResult, error)
```
robert-zaremba, line 57: same here, I think we shouldn't use `sdk.Dec`

app/app.go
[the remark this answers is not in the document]
robert-zaremba: if you rename a directory and rename the package name to `grouptypes` you won't need to make any alias. same for `server`. Also, check`x/data` - in new modules we prefer to put types directly in the module root, so it would be `x/group`. Then you reference it as `group.MyStruct`

proto/regen/group/v1alpha1/types.proto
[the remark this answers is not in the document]
robert-zaremba: separately in the storage, why don't we store it in the group object?

proto/regen/group/v1alpha1/types.proto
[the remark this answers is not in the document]
robert-zaremba: Then let's rename `RESULT_UNDEFINED` to `RESULT_NOT_STARTED`.
Maybe we can also change the naming in all other enums: `UNSPECIFIED` -> `INVALID`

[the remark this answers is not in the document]
blushi: proto linter requires 0 enum values to end with `UNSPECIFIED` so we would have to use `RESULT_INVALID_UNSPECIFIED`, wdyt?
 Re: `RESULT_UNDEFINED`, I'm not sure if `RESULT_NOT_STARTED` is the right term, what about `RESULT_UNFINALIZED`?

proto/regen/group/v1alpha1/types.proto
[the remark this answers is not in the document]
blushi: there's actually no group object.

x/group/server/keeper.go
```diff
@@ -0,0 +1,542 @@
...
+	ProposalByGroupAccountIndexPrefix byte = 0x32
+	ProposalByProposerIndexPrefix     byte = 0x33
+
+	// Vote Table
+	VoteTablePrefix           byte = 0x40
+	VoteByProposalIndexPrefix byte = 0x41
+	VoteByVoterIndexPrefix    byte = 0x42
+)
+
+type Keeper struct {
```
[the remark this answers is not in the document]
blushi: I don't believe we can at this point, we need it in `RegenApp` in `app.go`.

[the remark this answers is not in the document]
blushi: I don't believe we can at this point, we need it in `x/group/module` `AppModule`.

x/group/types/proposal.go
blushi: This method is not creating or making messages, it's just getting the already stored messages in the form of `Any`s as `sdk.Msg`s. This is the format we have been using so far in the sdk (see for instance `MsgSubmitEvidence` `GetEvidence()` or `BaseAccount` `GetPubKey()`).

proto/regen/group/v1alpha1/types.proto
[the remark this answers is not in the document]
robert-zaremba: Ah, I didn't know about that linter rule. Then let's keep as is for unspecified.

`RESULT_UNFINALIZED` sounds better :+1:

x/group/server/keeper.go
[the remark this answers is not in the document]
robert-zaremba: OK, so probably later we will need to refactore it to a new design of MsgServices (see `x/data` for example - there is no Keeper).

x/group/server/keeper.go
```diff
@@ -0,0 +1,542 @@
...
+		}
+	}
+
+	if err := ensureMsgAuthZ(msgs, account.GroupAccount); err != nil {
+		return 0, err
+	}
+
+	blockTime, err := gogotypes.TimestampProto(ctx.BlockTime())
+	if err != nil {
+		return 0, sdkerrors.Wrap(err, "block time conversion")
```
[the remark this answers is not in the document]
blushi: this is actually a conversion https://godoc.org/github.com/gogo/protobuf/types#TimestampProto

x/group/server/keeper.go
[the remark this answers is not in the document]
blushi: yeah agreed and I would have followed the same design as `x/data` and `x/ecocredit` if I could but we need to provide a `paramstypes.Subspace` to set max comment length and the current way of registering services doesn't support that.

x/group/server/keeper.go
```diff
@@ -0,0 +1,542 @@
...
+			Comment:     comment,
+			SubmittedAt: *blockTime,
+		}
+		if err := proposal.VoteState.Add(newVote, voter.Weight); err != nil {
+			return sdkerrors.Wrap(err, "add new vote")
+		}
+
+		if err := k.voteTable.Create(ctx, &newVote); err != nil {
+			return sdkerrors.Wrap(err, "store vote")
+		}
```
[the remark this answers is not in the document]
aaronc: That's a good instinct @robert-zaremba but welcome to cosmos 😉. Currently the state machine can't access transaction history, but ideally there would be better integration with tendermint. I suggest we push for it upstream. Also the state store is for now doubling as a query engine. 

Now, the one legitimate reason for storing votes in state is if a voter can change their vote after having voted. I believe this current decision policy allows that. Some decision policies would, others might not.

x/group/server/keeper.go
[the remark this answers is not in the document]
robert-zaremba: If `paramstypes.Subspace` is available in `NewRegenApp`  then you can do it: 
* when creating a module, you can pass more parameters
* then you can store it in a module structure
* finally, the module object calls `server.RegisterServices` and it can pass that parameters to that function.

x/group/server/keeper.go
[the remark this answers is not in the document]
robert-zaremba: Ah, ok. I thought we have bytes there. :+1:

x/group/server/keeper.go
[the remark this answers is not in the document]
robert-zaremba: Would be good to do a simulation of data we are going to store and how this will impact the storage size.

app/app.go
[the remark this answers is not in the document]
blushi: Doing that was causing some import cycles back when I started working on the group module, but maybe I believe I could get around that now...

proto/regen/group/v1alpha1/types.proto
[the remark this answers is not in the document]
blushi: That's pretty much the same as `ClassInfo` or `BatchInfo` for the `ecocredit` module. Maybe we should just use the same terminology here, ie use `GroupInfo` for instance instead, wdyt?

x/group/server/keeper.go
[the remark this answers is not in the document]
blushi: good point, will try that approach, thanks!

x/group/types/conditions.go
blushi: TBH I haven't really touched this part of the code that was pre-existing in cosmos-modules. A group account address is directly derived from a `Condition` so I guess that's why it was originally designed as `[]byte`. I agree using a struct would be cleaner. That being said, that part will need to be revisited as part of ADR 028 and ADR 033 eventually...

proto/regen/group/v1alpha1/types.proto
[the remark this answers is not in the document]
robert-zaremba: Unifying naming is always good. :+1:

x/group/server/keeper.go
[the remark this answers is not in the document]
blushi: added this issue: https://github.com/regen-network/regen-ledger/issues/204

orm/iterator.go
```diff
@@ -92,6 +95,91 @@ func First(it Iterator, dest Persistent) (RowID, error) {
 	return binKey, nil
 }
 
+func Paginate(
+	it Iterator,
```
blushi, line 99: Here we provide an `Iterator` because it needs to be created based on the query needs, see following examples:
- If we wanna query all items from an `orm` table, we can use `table.PrefixScan(ctx, start, end)` with `start = pageReq.Key` if `pageReq.Key` is provided to get an `Iterator`
- If we wanna query certain items based on certain conditions (eg query groups by admin), we can use corresponding `MultiKeyIndex`, eg `groupByAdminIndex.Get(ctx, searchKey)` where `searchKey` corresponds to the admin account address bytes. In this case, we don't really use `pageReq.Key` if provided and the returned `*query.PageResponse`.`NextKey` will return the next `RowID`, which doesn't seem to be usable in this particular context. But maybe I'm missing something here, any thoughts @aaronc?

aaronc: Can you explain why `pageReq.Key` isn't used in the second case?

blushi: @aaronc nevermind I believe I figured it out, ie how to query a `MultiKeyIndex` (e.g. with the example of the `groupByAdminIndex`, querying groups by admin account address) only from a certain key, will upgrade the current implementation accordingly.

proto/regen/group/v1alpha1/types.proto
```diff
@@ -0,0 +1,266 @@
...
+import "google/protobuf/timestamp.proto";
+import "cosmos_proto/cosmos.proto";
+import "google/protobuf/any.proto";
+
+// Member represents a group member with an account address,
+// non-zero power and a comment.
+message Member {
+
+    // address is the member's account address.
+    bytes address = 1 [(gogoproto.casttype) = "github.com/cosmos/cosmos-sdk/types.AccAddress"];
```
clevinson, line 18: Should this be `string` so as to keep consistent with how we are encoding addresses as bech32 encoded strings elsehwere in the SDK ?

proto/regen/group/v1alpha1/tx.proto
```diff
@@ -0,0 +1,260 @@
...
+//
+// Proposals and Voting
+//
+
+// MsgCreateProposalRequest is the Msg/CreateProposal request type.
+message MsgCreateProposalRequest {
+    option (gogoproto.goproto_getters) = false;
+
+    // group_account is the group account address.
+    bytes group_account = 1 [(gogoproto.casttype) = "github.com/cosmos/cosmos-sdk/types.AccAddress"];
```
clevinson, line 208: Similarly, and across all other `types.AccAddress` fields, should this be `string` so as to keep consistent with how we are encoding addresses as bech32 encoded strings elsehwere in the SDK ?

aaronc:
@blushi is this ready to be merged?

blushi:
> @blushi is this ready to be merged?

@aaronc Just pushed my latest changes, so it should be ready to be merged (as long as it gets approved).

aaronc:
Great thanks @blushi 🙏 ! I'll take a look. Happy holidays!!

orm/auto_uint64.go
```diff
@@ -1,20 +1,21 @@
...
 import (
+	"github.com/cosmos/cosmos-sdk/codec"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 var _ Indexable = &AutoUInt64TableBuilder{}
 
 // NewAutoUInt64TableBuilder creates a builder to setup a AutoUInt64Table object.
-func NewAutoUInt64TableBuilder(prefixData byte, prefixSeq byte, storeKey sdk.StoreKey, model Persistent) *AutoUInt64TableBuilder {
+func NewAutoUInt64TableBuilder(prefixData byte, prefixSeq byte, storeKey sdk.StoreKey, model codec.ProtoMarshaler, cdc codec.Marshaler) *AutoUInt64TableBuilder {
```
aaronc, line 11: Can we replace `codec.ProtoMarshaler` with `proto.Message`? That would be more clear, maybe for a separate PR though.

aaronc: I guess that would require changing the `Marshaler` interface upstream. So not relevant for this PR, but we should do that upstream.

proto/regen/group/v1alpha1/events.proto
```diff
@@ -0,0 +1,45 @@
...
+
+// EventUpdateGroupAccount is an event emitted when a group account is updated.
+message EventUpdateGroupAccount {
+
+  // group_account is the address of the group account.
+  string group_account = 1;
+
+  // admin is the admin address of the group.
+  string admin = 2;
+}
```
aaronc, line 45: What is the purpose in only including `admin` for each of these events? If we wanted to be minimalist we would only include `group_id` and `group_account` I think. If we wanted more info, would we maybe reference each member??

Also, events are missing for proposals and voting. I'm fine if that happens in separate PRs, but we should have it.

proto/regen/group/v1alpha1/genesis.proto
```diff
@@ -0,0 +1,25 @@
...
+}
+
+// GenesisState defines the group module's genesis state.
+message GenesisState {
+    option (gogoproto.equal) = true;
+    option (gogoproto.goproto_stringer) = false;
+
+    // params are the genesis state Params.
+    Params params = 1 [(gogoproto.nullable)=false];
+}
```
aaronc, line 25: Don't we also need to include groups, group accounts, proposals, etc?

aaronc: I'm fine if that's a separate PR but let's just make sure we have an issue open.

blushi: indeed, let's make it part of a separate PR, I added a new issue: https://github.com/regen-network/regen-ledger/issues/214
it also seems like there was already some WIP on that: https://github.com/regen-network/cosmos-modules/pull/60/files

proto/regen/group/v1alpha1/events.proto
blushi: I guess it was to replicate the existing events https://github.com/regen-network/cosmos-modules/blob/master/incubator/group/group_handler.go#L105-L108 which included the signer but I agree we can just remove it for now to keep it minimalist.
If we wanted more info, we could just add the corresponding request data (eg comment, members...).

Also, as mentioned in the PR description above, right now we're are using different endpoints for updating a group whether we wanna update its members, admin or comments. It would be better UX-wise to just have one update endpoint if possible and that would corresponds better to the current events, wdyt?

I added a new issue to add more events: https://github.com/regen-network/regen-ledger/issues/215

orm/auto_uint64.go
blushi: How would we replace `codec.ProtoMarshaler` with `proto.Message` in the `Marshaler` methods since `codec.ProtoMarshaler`'s methods like `Marshal()` seems to be needed (eg in `ProtoCodec`)? We would need to add type assertions in the `BinaryMarshaler` methods like it's the case in the `JSONMarshaler` methods.

proto/regen/group/v1alpha1/events.proto
aaronc: Could you explain a little more what you mean by one update endpoint?

orm/auto_uint64.go
aaronc: We would just use the `proto.Marshal` method instead which is the official  approach.

blushi: If you look at https://github.com/regen-network/regen-ledger/blob/marie/136-x-group/proto/regen/group/v1alpha1/tx.proto#L18-L25, you can see that we have multiple methods for updating groups. I thought it would be better to have just one method `UpdateGroup` instead of `UpdateGroupMembers`, `UpdateGroupAdmin`, etc.

proto/regen/group/v1alpha1/events.proto
blushi: If you look at https://github.com/regen-network/regen-ledger/blob/marie/136-x-group/proto/regen/group/v1alpha1/tx.proto#L18-L25, you can see that we have multiple methods for updating groups. I thought it would be better to have just one method `UpdateGroup` instead of `UpdateGroupMembers`, `UpdateGroupAdmin`, etc.

aaronc:
Is this all ready to go again @blushi ?

Would be great to get a second set of eyes as this is a rather large PR... But I would love to get this merged ASAP.

blushi:
> Is this all ready to go again @blushi ?
> 
> Would be great to get a second set of eyes as this is a rather large PR... But I would love to get this merged ASAP.

Yes.
Ping @anilCSE @aleem1314 @atheeshp @amaurymartiny @clevinson @robert-zaremba :)

proto/regen/group/v1alpha1/query.proto
```diff
@@ -0,0 +1,246 @@
...
+// Query is the regen.group.v1alpha1 Query service.
+service Query {
+
+  // GroupInfo queries group info based on group id.
+  rpc GroupInfo(QueryGroupInfoRequest) returns (QueryGroupInfoResponse);
+
+  // GroupAccountInfo queries group account info based on group account address.
+  rpc GroupAccountInfo(QueryGroupAccountInfoRequest) returns (QueryGroupAccountInfoResponse);
+  
+  // GroupMembers queries members for agroup
```
anilCSE, line 20:
```suggestion
  // GroupMembers queries members for a group
```

proto/regen/group/v1alpha1/genesis.proto
anilCSE, line 25:
```suggestion
}

```

anilCSE, line 25:
```suggestion
}

```

proto/regen/group/v1alpha1/events.proto
```diff
@@ -0,0 +1,33 @@
...
+  // group_account is the address of the group account.
+  string group_account = 1;
+}
+
+// EventUpdateGroupAccount is an event emitted when a group account is updated.
+message EventUpdateGroupAccount {
+
+  // group_account is the address of the group account.
+  string group_account = 1;
+}
```
anilCSE, line 33: this file needs formating
```suggestion
}

```

proto/regen/group/v1alpha1/types.proto
```diff
@@ -0,0 +1,266 @@
...
+
+    // executor_result is the final result based on the votes and election rule. Initial value is NotRun.
+    ExecutorResult executor_result = 11;
+
+    // msgs is a list of Msgs that will be executed if the proposal passes.
+    repeated google.protobuf.Any msgs = 12;
+}
+
+// Tally represents the sum of weighted votes.
+message Tally {
```
anilCSE, line 233: Can't we use these definitions from `x/gov`?

proto/regen/group/v1alpha1/query.proto
```diff
@@ -0,0 +1,246 @@
...
+import "gogoproto/gogo.proto";
+import "cosmos/base/query/v1beta1/pagination.proto";
+
+option go_package = "github.com/regen-network/regen-ledger/x/group";
+
+// Query is the regen.group.v1alpha1 Query service.
+service Query {
+
+  // GroupInfo queries group info based on group id.
+  rpc GroupInfo(QueryGroupInfoRequest) returns (QueryGroupInfoResponse);
```
anilCSE, line 15: It can be a separate PR but grpc-gateway annotations are missing.

testutil/testdata/tx.proto
```diff
@@ -0,0 +1,11 @@
...
+
+package testdata;
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/regen-network/regen-ledger/testutil/testdata";
+
+message MsgAuthenticated {
+    repeated bytes signers = 1 [(gogoproto.casttype) = "github.com/cosmos/cosmos-sdk/types.AccAddress"];
+}
```
anilCSE, line 11:
```suggestion
}

```

x/group/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+A group account is an account associated with a group and a decision policy.
+Group accounts are abstracted from groups because a single group may have
+multiple decision policies for different types of actions. Managing group
+membership separately from decision policies results in the least overhead
+and keeps membership consistent across different policies. The pattern that
+is recommended is to have a single master group account for a given group,
+and then to create separate group accounts with different decision policies
+and delegate the desired permissions to from the master account to
+those "sub-accounts" using the `msg_authorization` module.
```
anilCSE, line 21:
```suggestion
those "sub-accounts" using the `x/authz` module.
```

anilCSE, line 21: How can we use the group account to `Authorize` users (other accounts) for some `messages`? Let's say I want to use the group account as the main account (granter) and grant some permissions to an account (grantee) to transfer funds. Who needs to sign this `authorization` transaction (using `x/authz`'s `grant` transaction)?

anilCSE (changes requested):
mostly lgtm, some queries

proto/regen/group/v1alpha1/query.proto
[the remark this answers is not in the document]
blushi: yes I guess we'll tackle that all modules in a separate PR, see https://github.com/regen-network/regen-ledger/issues/182

blushi: yes, cf. https://github.com/regen-network/regen-ledger/issues/182

orm/auto_uint64_test.go
```diff
@@ -13,34 +15,37 @@ import (
 )
 
 func TestAutoUInt64PrefixScan(t *testing.T) {
+	interfaceRegistry := types.NewInterfaceRegistry()
+	cdc := codec.NewProtoCodec(interfaceRegistry)
```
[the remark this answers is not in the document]
blushi: we could but it would add up some stuff that are not necessarily needed for this test so I'd rather keep it as is.

proto/regen/group/v1alpha1/types.proto
blushi: you mean `TallyResult`?
well the group module uses `apd.Decimal` internally to represent the weighted sum of votes while `x/gov` uses `sdk.Int`.

x/group/README.md
blushi: the group account could be seen as a multisig account

x/group/server/msg_server.go
```diff
@@ -0,0 +1,604 @@
...
+	// TODO: add event?
+
+	return &group.MsgCreateProposalResponse{ProposalId: group.ProposalID(id)}, nil
+}
+
+func (s serverImpl) Vote(ctx types.Context, req *group.MsgVoteRequest) (*group.MsgVoteResponse, error) {
+	id := req.ProposalId
+	voters := req.Voters
+	choice := req.Choice
+	comment := req.Comment
```
anilCSE, line 367: This is bit confusing. Why `voters` instead of `voter`?

x/group/server/msg_server.go
```diff
@@ -0,0 +1,604 @@
...
+	electorate, err := s.getGroupInfo(ctx, accountInfo.GroupId)
+	if err != nil {
+		return nil, err
+	}
+	if electorate.Version != proposal.GroupVersion {
+		return nil, sdkerrors.Wrap(group.ErrModified, "group was modified")
+	}
+
+	// count and store votes
+	for _, voterAddr := range voters {
```
anilCSE, line 416: This doesn't sound good, if we are considering `voters`, we might need to consider array of votes directly (so each voter can have different `choice`)

x/group/server/msg_server.go
```diff
@@ -0,0 +1,604 @@
...
+			Comment:     comment,
+			SubmittedAt: *blockTime,
+		}
+		if err := proposal.VoteState.Add(newVote, voter.Weight); err != nil {
+			return nil, sdkerrors.Wrap(err, "add new vote")
+		}
+
+		if err := s.voteTable.Create(ctx, &newVote); err != nil {
+			return nil, sdkerrors.Wrap(err, "store vote")
+		}
```
anilCSE, line 434: Should we also check if `voter` has `voted` already?

anilCSE (changes requested):
Mostly lgtm. Requested a few nits.

A couple of queries:
- I can't think of a usecase right away but I believe we might need a custom `VotingPeriodStart` time so that people can create/plan proposals before hand (definitely not scoped for this PR)
- Should we add minimum deposit to ensure the state is not spammed with fake `proposals`?

go.mod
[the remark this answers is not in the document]
aaronc: reminder that we should update these versions like we did in the SDK

go.mod
```diff
@@ -15,10 +15,14 @@ require (
 	github.com/lib/pq v1.8.0 // indirect
 	github.com/multiformats/go-multihash v0.0.14
 	github.com/rakyll/statik v0.1.7
+	github.com/regen-network/cosmos-proto v0.3.0
+	github.com/spf13/afero v1.3.4 // indirect
```
aaronc, line 19: what's this for? maybe we need to clean this file?

orm/uint64_index.go
```diff
@@ -36,8 +40,8 @@ func (i UInt64Index) Has(ctx HasKVStore, key uint64) bool {
 }
 
 // Get returns a result iterator for the searchKey. Parameters must not be nil.
-func (i UInt64Index) Get(ctx HasKVStore, searchKey uint64) (Iterator, error) {
-	return i.multiKeyIndex.Get(ctx, EncodeSequence(searchKey))
+func (i UInt64Index) Get(ctx HasKVStore, searchKey uint64, pageRequest *query.PageRequest) (Iterator, error) {
```
aaronc, line 43: Maybe this should be a separate `GetPaginated` method?

proto/regen/group/v1alpha1/types.proto
```diff
@@ -0,0 +1,266 @@
...
+message Member {
+
+    // address is the member's account address.
+    string address = 1;
+    
+    // power is the member's power that should be greater than 0.
+    string power = 2;
+    
+    // comment is the member's comment.
+    string comment = 3;
```
aaronc, line 24: Maybe we should rename all these comments to `metadata`? And have them be `bytes` link in `x/ecocredit`? Then they could maybe be linked to ID's in `x/data`? Wdyt?

proto/regen/group/v1alpha1/types.proto
```diff
@@ -0,0 +1,266 @@
...
+    // or any member is added or removed this version is incremented and will
+    // cause proposals based on older versions of this group to fail
+    uint64 version = 4;
+
+    // total_weight is the sum of the group members' powers.
+    string total_weight = 5;
+}
+
+// GroupMember represents the relationship between a group and a member.
+message GroupMember {
```
aaronc, line 85: I believe the type `GroupMember` is only used in query responses and should just be in that file, maybe called `GroupMemberPair` or something... That would address these TODOs right? And we can just embed `Member` inside here to reduce types.

scripts/protocgen.sh
```diff
@@ -43,11 +43,11 @@ Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
...
-Mgoogle/protobuf/duration.proto=github.com/gogo/protobuf/types,\
-Mgoogle/protobuf/struct.proto=github.com/gogo/protobuf/types,\
-Mgoogle/protobuf/timestamp.proto=github.com/gogo/protobuf/types,\
-Mgoogle/protobuf/wrappers.proto=github.com/gogo/protobuf/types,\
-Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
+google/protobuf/duration.proto=github.com/gogo/protobuf/types,\
+google/protobuf/struct.proto=github.com/gogo/protobuf/types,\
+google/protobuf/timestamp.proto=github.com/gogo/protobuf/types,\
+google/protobuf/wrappers.proto=github.com/gogo/protobuf/types,\
+google/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
```
aaronc, line 50: Are these `M`s unneeded?

testutil/server/configurator/configurator.go
```diff
@@ -0,0 +1,117 @@
+// Package configurator a Fixture defines an implementation of server.Fixture which uses an in-memory test data store
+// with no backing ABCI app. It implements the module.Configurator interface and is designed
+// to be used in server integration tests for each module independent of a larger app.
+//
+// Ex:
+//
+package configurator
```
aaronc, line 7: Can't we just use `types/module/server/testutil.go` instead of this whole package? This looks like code that already got removed

types/module/configurator.go
```diff
@@ -0,0 +1,43 @@
+package module
```
aaronc, line 1: Maybe this code is old code that can be removed and updated to what's in master?

types/module/server/manager.go
```diff
@@ -149,7 +149,7 @@ func (c *configurator) ModuleKey() RootModuleKey {
 	return c.key
 }
 
-func (c *configurator) BinaryMarshaler() codec.BinaryMarshaler {
+func (c *configurator) Marshaler() codec.Marshaler {
```
aaronc, line 152: Do we need JSON now?

x/group/README.md
```diff
@@ -0,0 +1,67 @@
...
+There are four choices to choose while voting - yes, no, abstain and veto. Not
+all decision policies will support them. Votes can contain an optional comment.
+During the voting window, accounts that have already voted may change their vote.
+In the current implementation, the voting window begins as soon as a proposal
+is submitted.
+
+## Executing Proposals
+
+Proposals will not be automatically executed by the chain in this current design,
+but rather a user must submit a `MsgExecRequest` transaction to attempt to execute the
```
aaronc, line 58: `Msg/Exec`
```suggestion
but rather a user must submit a `Msg/Exec` transaction to attempt to execute the
```

x/group/README.md
```diff
@@ -0,0 +1,67 @@
...
+proposal based on the current votes and decision policy. A future upgrade could
+automate this propose and have the group account (or a fee granter) pay.
+
+## Changing Group Membership
+
+In the current implementation, changing a group's membership (adding or removing members or changing their power)
+will cause all existing proposals for group accounts linked to this group
+to be invalidated. They will simply fail if someone calls `MsgExecRequest` and will
```
aaronc, line 66:
```suggestion
to be invalidated. They will simply fail if someone calls `Msg/Exec` and will
```

x/group/README.md
```diff
@@ -0,0 +1,67 @@
...
+### Threshold decision policy
+
+A threshold decision policy defines a threshold of yes votes (based on a tally
+of voter weights) that must be achieved in order for a proposal to pass. For
+this decision policy, abstain and veto are simply treated as no's.
+
+## Proposal
+
+Any member of a group can submit a proposal for a group account to decide upon.
+A proposal consists of a set of `sdk.Msg`s that will be executed if the proposal
```
aaronc, line 44:
```suggestion
A proposal consists of a set of message that will be executed if the proposal
```

x/group/conditions.go
aaronc: Is there an issue to track this @blushi ? I think we have a clearer sense of what we'll be doing now.

aaronc: Oh nevermind, there is...

x/group/module/module.go
```diff
@@ -0,0 +1,128 @@
...
+	return []abci.ValidatorUpdate{}
+}
+
+func (a AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONMarshaler) json.RawMessage {
+	var p group.Params
+	a.paramSpace.GetParamSet(ctx, &p)
+	return cdc.MustMarshalJSON(&group.GenesisState{
+		Params: p,
+	})
+}
```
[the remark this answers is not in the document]
aaronc: there's an issue to fully implement genesis

x/group/server/proposal_executor.go
```diff
@@ -0,0 +1,42 @@
...
+
+// DoExecuteMsgs routes the messages to the registered handlers. Messages are limited to those that require no authZ or
+// by the group account only. Otherwise this gives access to other peoples accounts as the sdk ant handler is bypassed
+func DoExecuteMsgs(ctx sdk.Context, router sdk.Router, groupAccount sdk.AccAddress, msgs []sdk.Msg) ([]sdk.Result, error) {
+	results := make([]sdk.Result, len(msgs))
+	if err := ensureMsgAuthZ(msgs, groupAccount); err != nil {
+		return nil, err
+	}
+	for i, msg := range msgs {
+		handler := router.Route(ctx, msg.Route())
```
aaronc, line 29: should we be using ADR 031 here?

aaronc (changes requested):
Okay generally looks good from what I've seen. Before approving I want to do one more high-level logic review of msg_server.go and other core logic. But I think we're pretty close.

One high level change request is switching to `types/module/server/testutil.go` and removing the `testutil/server` code which already was deleted.

x/group/module/module.go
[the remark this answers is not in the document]
blushi: https://github.com/regen-network/regen-ledger/issues/214

x/group/module/module.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (a AppModule) RegisterInvariants(sdk.InvariantRegistry) {
+	// todo: anything to check?
+	// todo: check that tally sums must never have less than block before ?
```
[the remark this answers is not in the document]
blushi: #223

x/group/server/msg_server.go
```diff
@@ -0,0 +1,604 @@
...
+	err = ctx.EventManager().EmitTypedEvent(&group.EventCreateGroupAccount{GroupAccount: accountAddr.String()})
+	if err != nil {
+		return nil, err
+	}
+
+	return &group.MsgCreateGroupAccountResponse{GroupAccount: accountAddr.String()}, nil
+}
+
+func (s serverImpl) UpdateGroupAccountAdmin(ctx types.Context, req *group.MsgUpdateGroupAccountAdminRequest) (*group.MsgUpdateGroupAccountAdminResponse, error) {
+	// TODO
```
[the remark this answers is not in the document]
blushi: #224 added next to TODO

scripts/protocgen.sh
blushi: This was causing some issues when using both `google.protobuf.Duration` and `google.protobuf.Timestamp` in the same package: `protoc-gen-go-cosmos2: Go package "github.com/gogo/protobuf/types" has inconsistent names duration (google/protobuf/duration.proto) and timestamp (google/protobuf/timestamp.proto)`, coming from https://github.com/protocolbuffers/protobuf-go/blob/master/compiler/protogen/protogen.go#L375

types/module/server/manager.go
blushi: yes for the ORM genesis import/export functions

x/group/typesupport.go
```diff
@@ -0,0 +1,45 @@
...
+func (a AccAddresses) ValidateBasic() error {
+	index := make(map[string]struct{}, len(a))
+	for i := range a {
+		accAddr := a[i]
+		if accAddr.Empty() {
+			return sdkerrors.Wrap(ErrEmpty, "address")
+		}
+		if err := sdk.VerifyAddressFormat(accAddr); err != nil {
+			return sdkerrors.Wrap(err, "address")
+		}
```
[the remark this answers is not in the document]
blushi: Indeed, just added some doc to `AccAddresses.ValidateBasic` to point out that individual address validation should be done separately.

x/group/server/msg_server.go
blushi: It can take as input multiple voters with the same choice per the initial design. I agree this is not ideal and it would be good to be able to define different choice for each voter, or rather to have something similar to `x/gov` where only one voter can vote at a time, wdyt?

x/group/server/msg_server.go
blushi: see https://github.com/regen-network/regen-ledger/pull/154#discussion_r562493715

x/group/server/msg_server.go
blushi: this is already handled by the ORM which will return a `unique constraint violation` error if it already exists, there's also a test that covers this case but I'll add a comment to make it clearer.

proto/regen/group/v1alpha1/types.proto
blushi: sounds good!

x/group/server/msg_server.go
anilCSE: Yeah, to keep it simple we can just allow one vote for now

orm/iterator.go
```diff
@@ -92,6 +95,91 @@ func First(it Iterator, dest Persistent) (RowID, error) {
...
+
+		count++
+
+		if count <= offset {
+			continue
+		}
+
+		if err != nil {
+			return nil, err
+		}
```
robert-zaremba, line 162: so we are ignoring the errors before the offset, but we are not ignoring the errors after `limit` if `countTotal` is set. Personally, I would not ignore any error.

orm/iterator.go
```diff
@@ -92,6 +95,91 @@ func First(it Iterator, dest Persistent) (RowID, error) {
...
+		if count <= end {
+			t = reflect.Append(t, val)
+		} else if count == end+1 {
+			nextKey = binKey
```
robert-zaremba, line 166: should we set the `nextKey` when the iterator is exhausted? Otherwise it will have empty value.

orm/iterator.go
```diff
@@ -92,6 +95,91 @@ func First(it Iterator, dest Persistent) (RowID, error) {
...
+			slice.Set(t)
```
robert-zaremba, line 167: let's move the destination setting after the loop. No need to repeat that code.
```
destRef.Set(tmpSlice)
```

orm/iterator_test.go
```diff
@@ -90,48 +94,148 @@ func TestReadAll(t *testing.T) {
...
+			} else {
+				require.NoError(t, err)
+				assert.EqualValues(t, spec.exp, loaded)
+				assert.EqualValues(t, spec.expPageRes.Total, res.Total)
+				assert.EqualValues(t, spec.expPageRes.NextKey, res.NextKey)
+			}
+
+		})
+	}
+}
```
robert-zaremba, line 235: we need to add more tests to cover edge cases we were discussing while reviewing the `Paginate` function.

orm/uint64_index.go
```diff
@@ -40,6 +44,11 @@ func (i UInt64Index) Get(ctx HasKVStore, searchKey uint64) (Iterator, error) {
 	return i.multiKeyIndex.Get(ctx, EncodeSequence(searchKey))
 }
 
+// GetPaginated returns a result iterator for the searchKey based on pageRequest.
```
robert-zaremba, line 47:
```suggestion
// GetPaginated creates an iterator, starting from the searchKey and based on pageRequest.
```

types/module/configurator.go
blushi: You mean using the new `Configurator` interface from `types/module/server` instead? That would mean temporarily adding some public `NewConfigurator` function so that we can use it in `app.go` `app.mm.RegisterServices(servermodule.NewConfigurator(...))` which does not seem to be the primary purpose of this new `configurator` type but could work for now, are we ok with this?

orm/uint64_index.go
```diff
@@ -40,6 +44,11 @@ func (i UInt64Index) Get(ctx HasKVStore, searchKey uint64) (Iterator, error) {
...
+func (i UInt64Index) GetPaginated(ctx HasKVStore, searchKey uint64, pageRequest *query.PageRequest) (Iterator, error) {
+	return i.multiKeyIndex.GetPaginated(ctx, EncodeSequence(searchKey), pageRequest)
+}
```
robert-zaremba, line 50: Should we add a test for this function?

robert-zaremba (changes requested):
Only `orm` part now. We had an on screen review of the Paginate functions. I left the comments we were discussing. 

There are few other things as well, notably adding more tests for edge cases.

orm/auto_uint64.go
robert-zaremba: https://github.com/cosmos/cosmos-sdk/issues/8412

proto/regen/group/v1alpha1/types.proto
blushi: The `Params` `MaxCommentLength` doesn't seem to be relevant anymore in this context, should we just get rid of it?

orm/iterator.go
```diff
@@ -92,6 +95,91 @@ func First(it Iterator, dest Persistent) (RowID, error) {
...
+		if ErrIteratorDone.Is(err) {
+			slice.Set(t)
+			break
+		}
+
+		count++
+
+		if count <= offset {
+			continue
+		}
```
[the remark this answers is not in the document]
anilCSE: it works for even `offset==0`, I don't see any potential issues?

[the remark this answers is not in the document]
blushi: yes actually `count` is always `>= 1` when doing this check, so if `offset == 0` then it'll just skip it.

orm/iterator.go
```diff
@@ -92,6 +95,91 @@ func First(it Iterator, dest Persistent) (RowID, error) {
...
+	it Iterator,
+	pageRequest *query.PageRequest,
+	dest ModelSlicePtr,
+) (*query.PageResponse, error) {
+	// if the PageRequest is nil, use default PageRequest
+	if pageRequest == nil {
+		pageRequest = &query.PageRequest{}
+	}
+
+	offset := pageRequest.Offset
```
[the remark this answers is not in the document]
blushi: That's just supported by `query.PageRequest` from cosmos-sdk.

orm/iterator.go
blushi: `nextKey` will just be nil which is expected in this case.

go.mod
```diff
@@ -15,10 +15,14 @@ require (
...
 	github.com/spf13/cast v1.3.1
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.6.1
+	golang.org/x/crypto v0.0.0-20201221181555-eec23a3978ad // indirect
```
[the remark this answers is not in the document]
blushi: not sure how it got there, got it cleaned up somehow.

testutil/server/configurator/configurator.go
blushi: Well the group module is kind of in an hybrid state at the moment since it doesn't really use the new server module manager and modules/services registration introduced by https://github.com/regen-network/regen-ledger/pull/157 because this doesn't handle genesis yet and we need to provide some params `Subspace` when registering the group services (in order to retrieve the max comment size) which is not handled either. That being said, we might just wanna get rid of this `MaxCommentLength` params (cf. https://github.com/regen-network/regen-ledger/pull/154#discussion_r562650879) so maybe we could move the group module to use #157 if we're ok with implementing genesis im/export later. Wdyt @aaronc?

x/group/server/proposal_executor.go
blushi: Something similar to what's in `BaseApp.runMsgs` instead? I remember we had previous discussions to keep using `sdk.Router` for now but happy to upgrade that if that makes sense.

proto/regen/group/v1alpha1/types.proto
aaronc: How about just `MaxMetadatLength`?

testutil/server/configurator/configurator.go
aaronc: We could just have `MaxMetadataLength` be fixed to 256 or 1024 bytes or something for now... What do you think? I don't really have a strong preference, but it seems like migrating to ADR 033 sooner rather than later would reduce code complexity a bit here no? And we'll need to address the `x/params` upgrade at some point soon anyway.

types/module/configurator.go
aaronc: I think that could work... but isn't ideal either. I would defer to whatever you think is the best path to be honest...

testutil/server/configurator/configurator.go
aaronc: Genesis isn't handled in ADR 033 either though right?

types/module/server/manager.go
aaronc: I see

x/group/server/proposal_executor.go
aaronc: I think it's fine for now, but can you add a follow up issue to address later?

testutil/server/configurator/configurator.go
blushi: I believe that just having some fixed `MaxMetadataLength` for now would be ok and it'd allow us to fully migrate the group module to ADR 033 and use `types/module/server/testutil.go` for tests then. We can deal with genesis later (https://github.com/regen-network/regen-ledger/issues/214).

types/module/configurator.go
blushi: Fully migrating the group module to ADR 033 will allow us to use the new module/service registration instead of `app.mm.RegisterServices(...)` (ref: https://github.com/regen-network/regen-ledger/pull/154#discussion_r564444402) so I'll just delete this `types/module/configurator.go` which won't be needed anymore.

x/group/server/proposal_executor.go
blushi: alright, added: https://github.com/regen-network/regen-ledger/issues/225

buf.yaml
```diff
@@ -12,9 +12,11 @@ lint:
   ignore:
     - cosmos
     - gogoproto
+    - cosmos_proto
```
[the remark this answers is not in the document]
aaronc: No it isn't

proto/regen/group/v1alpha1/types.proto
blushi: it's also used for the `groupMemberTable`, I think `GroupMember` is enough to show that it represents the actual relationship between a group and a member. But I guess we can embed `Member` inside it instead of repeating individual `Member`'s fields.

orm/iterator.go
```diff
@@ -92,6 +95,91 @@ func First(it Iterator, dest Persistent) (RowID, error) {
 	return binKey, nil
 }
 
+func Paginate(
```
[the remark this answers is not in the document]
blushi: Added some doc to the `Paginate` function, what do you think?

orm/uint64_index.go
blushi: Well it's not necessarily "starting" from the `searchKey`, but more the combined `searchKey` and `pageRequest.Key` if provided. I have updated the comment to make it clearer, wdyt?

orm/iterator.go
```diff
@@ -92,6 +95,91 @@ func First(it Iterator, dest Persistent) (RowID, error) {
...
+		if typ.Kind() == reflect.Ptr {
+			val.Set(reflect.New(typ.Elem()))
+			model = val
+		}
+
+		binKey, err := it.LoadNext(model.Interface().(codec.ProtoMarshaler))
+		if ErrIteratorDone.Is(err) {
+			slice.Set(t)
+			break
+		}
```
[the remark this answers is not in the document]
blushi: added.

orm/iterator.go
```diff
@@ -92,6 +95,91 @@ func First(it Iterator, dest Persistent) (RowID, error) {
...
+		countTotal = true
+	}
+
+	if it == nil {
+		return nil, errors.Wrap(ErrArgument, "iterator must not be nil")
+	}
+	defer it.Close()
+
+	var slice, t reflect.Value
+	typ, err := assertDest(dest, &slice, &t)
```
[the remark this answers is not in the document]
blushi: I kind of prefer `assertDest` in that it makes it more explicit that we're asserting that `dest` follows certain constraints...

orm/iterator.go
```diff
@@ -92,6 +95,91 @@ func First(it Iterator, dest Persistent) (RowID, error) {
...
+	}
+
+	end := offset + limit
+
+	var count uint64
+	var nextKey []byte
+	for {
+		obj := reflect.New(typ)
+		val := obj.Elem()
+		model := obj
```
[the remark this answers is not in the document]
blushi: if `elemType` is already a pointer (e.g. `dest` being some pointer to a slice of pointers, like `[]*GroupMember`), then `obj` is a pointer to a pointer which will cause issues when doing `obj.Interface().(codec.ProtoMarshaler)`. For that reason, we copy `obj` into `model` if we have a simple pointer but in case `elemType.Kind() == reflect.Ptr`, we overwrite it with `model = val`.

orm/iterator.go
```diff
@@ -158,3 +229,32 @@ func ReadAll(it Iterator, dest ModelSlicePtr) ([]RowID, error) {
 		rowIDs = append(rowIDs, binKey)
 	}
 }
+
+func assertDest(dest ModelSlicePtr, slice *reflect.Value, t *reflect.Value) (reflect.Type, error) {
```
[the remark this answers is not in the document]
blushi: done, let me know what you think, thanks.

orm/uint64_index.go
blushi: It was already tested through `TestPaginate` in `orm/iterator_test.go` but I have also added some additional tests in `orm/uint64_index_test.go` as well as `orm/index_test.go`.

orm/iterator.go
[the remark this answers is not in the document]
blushi: added some comments in the code.

orm/iterator.go
blushi: good point, fixed.

orm/iterator_test.go
blushi: Given https://github.com/regen-network/regen-ledger/pull/154#discussion_r562668406, I believe that what we were discussing was a bit off. But I have added additional tests anyway, please have a look, thanks.

blushi:
@aaronc @robert-zaremba @anilCSE could you have another look? Thanks.

proto/regen/group/v1alpha1/tx.proto
```diff
@@ -0,0 +1,260 @@
...
+    uint64 proposal_id = 1 [(gogoproto.casttype) = "ProposalID"];
+    
+    // signer is the account address used to execute the proposal.
+    string signer = 2;
+}
+
+// MsgExecResponse is the Msg/Exec request type.
+message MsgExecResponse { }
+
+
```
anilCSE, line 260:
```suggestion
```

proto/regen/group/v1alpha1/types.proto
```diff
@@ -0,0 +1,258 @@
...
+    
+    // metadata is any arbitrary metadata to attached to the member.
+    bytes metadata = 3;
+}
+
+// ThresholdDecisionPolicy implements the DecisionPolicy interface
+message ThresholdDecisionPolicy {
+    option (cosmos_proto.implements_interface) = "DecisionPolicy";
+
+    // threshold is a fix weight value that must be met or exceeded for a proposal to succeed.
```
anilCSE, line 31: does this refer to number of votes or total weight? may be being explicit here would be handy

anilCSE (approved):
lgtm

proto/regen/group/v1alpha1/types.proto
blushi: @anilCSE what do you think about this? 
```suggestion
    // threshold is the minimum weighted sum of yes votes that must be met or exceeded for a proposal to succeed.
```

anilCSE: Looks good.

testutil/server/fixture.go
```diff
@@ -42,4 +44,6 @@ type Fixture interface {
 
 	// Teardown performs any teardown actions for the fixture.
 	Teardown()
+
+	BankKeeper() bankkeeper.BaseKeeper
```
aaronc: What is the logic here?

x/group/server/msg_server.go
```diff
@@ -0,0 +1,599 @@
...
+	}
+
+	totalWeight := apd.New(0, 0)
+	for i := range members {
+		m := members[i]
+		if err := assertMetadataLength(m.Metadata, maxMetadataLength, "member metadata"); err != nil {
+			return nil, err
+		}
+
+		weight, err := math.ParseNonNegativeDecimal(m.Weight)
```
aaronc, line 39: Why would we accept a 0 weight when creating a group?

x/group/server/msg_server.go
```diff
@@ -0,0 +1,599 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		if !weight.IsZero() {
```
aaronc, line 44: ditto

x/group/server/msg_server.go
```diff
@@ -0,0 +1,599 @@
...
+			case orm.ErrNotFound.Is(err):
+				found = false
+			default:
+				return sdkerrors.Wrap(err, "get group member")
+			}
+
+			totalWeight, err := math.ParseNonNegativeDecimal(g.TotalWeight)
+			if err != nil {
+				return err
+			}
```
aaronc, line 112: This should be outside of the loop. No need to parse and serialize every iteration

x/group/server/msg_server.go
```diff
@@ -0,0 +1,599 @@
...
+			newMemberWeight, err := math.ParseNonNegativeDecimal(groupMember.Member.Weight)
+			if err != nil {
+				return err
+			}
+
+			// handle delete
+			if newMemberWeight.Cmp(apd.New(0, 0)) == 0 {
```
aaronc, line 119: Use `IsZero`

x/group/server/msg_server.go
```diff
@@ -0,0 +1,599 @@
...
+func (s serverImpl) UpdateGroupMembers(ctx types.Context, req *group.MsgUpdateGroupMembersRequest) (*group.MsgUpdateGroupMembersResponse, error) {
+	action := func(g *group.GroupInfo) error {
+		for i := range req.MemberUpdates {
+			groupMember := group.GroupMember{GroupId: req.GroupId,
+				Member: &group.Member{
+					Address:  req.MemberUpdates[i].Address,
+					Weight:   req.MemberUpdates[i].Weight,
+					Metadata: req.MemberUpdates[i].Metadata,
+				},
+			}
```
aaronc, line 97: let's move this to handle add + update

x/group/server/msg_server.go
```diff
@@ -0,0 +1,599 @@
...
+					return sdkerrors.Wrap(err, "add member")
+				}
+			} else if err := s.groupMemberTable.Create(ctx, &groupMember); err != nil {
+				return sdkerrors.Wrap(err, "add member")
+			}
+			err = math.Add(totalWeight, totalWeight, newMemberWeight)
+			if err != nil {
+				return err
+			}
+			g.TotalWeight = math.DecimalString(totalWeight)
```
aaronc, line 160: let's move this outside of the loop and use a var `totalWeight inside the loop

x/group/server/msg_server.go
```diff
@@ -0,0 +1,599 @@
...
+		1,
+		policy,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := s.groupAccountTable.Create(ctx, &groupAccount); err != nil {
+		return nil, sdkerrors.Wrap(err, "could not create group account")
+	}
```
aaronc, line 243: We also need to make sure that there is a module account created in `x/auth`. When we update with ADR 028 let's do this. Can you add a TODO @blushi ?

aaronc (changes requested):

x/group/server/msg_server.go
blushi: we need it before in "handle delete" or did you mean something else?

x/group/server/msg_server.go
blushi: added

x/group/server/msg_server.go
```diff
@@ -106,17 +114,14 @@ func (s serverImpl) UpdateGroupMembers(ctx types.Context, req *group.MsgUpdateGr
...
 				return err
 			}
 
-			// handle delete
-			if newMemberWeight.Cmp(apd.New(0, 0)) == 0 {
+			// Handle delete for members with zero weight.
+			if newMemberWeight.IsZero() {
+				// We can't delete a group member that doesn't already exist.
 				if !found {
 					return sdkerrors.Wrap(orm.ErrNotFound, "unknown member")
```
aaronc, line 126: There's no harm in ignoring this case right?

x/group/server/msg_server.go
```diff
@@ -238,6 +253,8 @@ func (s serverImpl) CreateGroupAccount(ctx types.Context, req *group.MsgCreateGr
 		return nil, err
 	}
 
+	// TODO Once we update to use ADR 028, we'll also need to make sure
+	// that there is a module account created in x/auth (#211).
```
aaronc, line 257: I'm thinking maybe as part of ADR 033 we actually had some sort of method `EnsureModuleAccountExists()` because this sort of functionality will be needed generically by ADR 033 users and we don't necessarily need to introduce an `x/auth` dependency. Wdyt @blushi ?

x/group/server/msg_server.go
```diff
@@ -376,8 +394,9 @@ func (s serverImpl) Vote(ctx types.Context, req *group.MsgVoteRequest) (*group.M
 	if err != nil {
 		return nil, err
 	}
+	// Ensure that we can still accept votes for this proposal.
 	if proposal.Status != group.ProposalStatusSubmitted {
-		return nil, sdkerrors.Wrap(group.ErrInvalid, "proposal not open")
+		return nil, sdkerrors.Wrap(group.ErrInvalid, "proposal not opened")
```
aaronc, line 399:
```suggestion
		return nil, sdkerrors.Wrap(group.ErrInvalid, "proposal not open for voting")
```

x/group/server/msg_server.go
```diff
@@ -445,6 +466,7 @@ func (s serverImpl) Vote(ctx types.Context, req *group.MsgVoteRequest) (*group.M
 	return &group.MsgVoteResponse{}, nil
 }
 
+// doTally updates proposal if needed based on the group account's decision policy.
```
aaronc, line 469:
```suggestion
// doTally updates the proposal status and tally if necessary based on the group account's decision policy.
```

aaronc (approved):
ACK 🎉 

I left a few minor doc suggestions that I'll just go ahead and commit.

There's also a few comments that can be dealt with in follow-ups if needed.

But otherwise, ready to merge this!

x/group/server/msg_server.go
blushi: yeah I guess `groupMemberTable.Delete` below would just fail in this case, but that would be an unnecessary access to the store given that we already know that it should fail, wdyt?

x/group/server/msg_server.go
blushi: Sounds good, I've updated the comment to not mention `x/auth` in particular but ADR 033.

x/group/server/msg_server.go
aaronc: We just wouldn't even execute this block... but maybe let's keep it as is because we would need to adjust tests and I don't want to further churn on this PR.

x/group/server/msg_server.go
aaronc: just it seems like it's initialized too early in the code

orm/uint64_index.go
robert-zaremba: good, thanks.

proto/regen/group/v1alpha1/query.proto
```diff
@@ -0,0 +1,245 @@
...
+
+  // GroupAccountInfo queries group account info based on group account address.
+  rpc GroupAccountInfo(QueryGroupAccountInfoRequest) returns (QueryGroupAccountInfoResponse);
+  
+  // GroupMembers queries members for a group
```
robert-zaremba, line 20:
```suggestion
  // GroupMembers queries members of a group
```

proto/regen/group/v1alpha1/query.proto
```diff
@@ -0,0 +1,244 @@
...
+  rpc GroupInfo(QueryGroupInfoRequest) returns (QueryGroupInfoResponse);
+
+  // GroupAccountInfo queries group account info based on group account address.
+  rpc GroupAccountInfo(QueryGroupAccountInfoRequest) returns (QueryGroupAccountInfoResponse);
+  
+  // GroupMembers queries members of a group
+  rpc GroupMembers(QueryGroupMembersRequest) returns (QueryGroupMembersResponse);
+
+  // GroupsByAdmin queries groups by admin address.
+  rpc GroupsByAdmin(QueryGroupsByAdminRequest) returns (QueryGroupsByAdminResponse);
```
robert-zaremba, line 24: It seams we are trying to implement relational DB - this queries should be handled by an off-chain indexer, or only added later when needed for state machine operations.

proto/regen/group/v1alpha1/query.proto
```diff
@@ -0,0 +1,244 @@
...
+
+  // GroupAccountsByGroup queries group accounts by group id.
+  rpc GroupAccountsByGroup(QueryGroupAccountsByGroupRequest) returns (QueryGroupAccountsByGroupResponse);
```
robert-zaremba, line 27: IMHO, would be better to remove Group prefix to avoid name clashes:

```
groupClient.GroupAccountsByGroup
```

robert-zaremba (approved):
Looks good. I would remove methods we don't need to not "pollute" the blockchain

proto/regen/group/v1alpha1/query.proto
aaronc: That's a bigger question. This decision was made a long time ago

proto/regen/group/v1alpha1/query.proto
aaronc: we can deal with that later... can you open an issue or PR if you think there's something that can be improved

robert-zaremba: @blushi  - do you have an opinion about naming? If you also find that we could improve naming then I'm happy to create a PR.

blushi: @robert-zaremba you mean having `AccountsByGroup` instead for instance? I don't have a strong preference, although I think keeping the Group prefix makes it more explicit.

robert-zaremba: > I think keeping the Group prefix makes it more explicit.

This context is already provided by a client / server object, right?
```
groupClient.AccountsByGroup
vs
groupClient.GroupAccountsByGroup   // we mention group 3 times, and it also sounds that we are grouping by (like in SQL)
```

PR #157, opened 2020-11-10
Implement ADR 033

aaronc:
ref: https://github.com/cosmos/cosmos-sdk/pull/7459 https://github.com/cosmos/cosmos-sdk/issues/7093
also addresses https://github.com/cosmos/cosmos-sdk/issues/7125 and https://github.com/cosmos/cosmos-sdk/issues/8030


types/module/server/module.go
```diff
@@ -0,0 +1,20 @@
...
+	module.ModuleBase
+
+	RegisterServices(Configurator)
+}
+
+type Configurator interface {
+	sdkmodule.Configurator
+
+	ModuleKey() ModuleKey
+	BinaryMarshaler() codec.BinaryMarshaler
```
blushi, line 19: don't we also need `RequireServer(serverInterface interface{})`?

aaronc: Yeah we do but in the simplest PoC we can just fail at runtime

testutil/server/configurator/configurator.go
```diff
@@ -81,7 +81,7 @@ func (c Fixture) Setup() server.Fixture {
 var _ server.Fixture = Fixture{}
 
 // TxConn implements the Fixture.TxConn method
-func (c Fixture) TxConn() grpc.ClientConnInterface {
+func (c Fixture) TxConn(sdk.AccAddress) grpc.ClientConnInterface {
```
blushi, line 84: why do add `sdk.Address` param here? could that be used to define a module or sub module account?

aaronc: Yes, I think a TxConn should always be in reference to a specific account. Otherwise, how do we specify who does the signing?

types/module/server/module.go
aaronc: ended up adding it anyway

testutil/server/configurator/configurator.go
aaronc: ended up reverting this

types/module/server/router.go
```diff
@@ -0,0 +1,194 @@
...
+
+	if writeCondition != nil && handler.commitWrites {
+		// msg handler
+		return func(ctx context.Context, request interface{}, response interface{}, opts ...interface{}) error {
+			if _, alreadyOnCallStack := rtr.antiReentryMap[moduleName]; alreadyOnCallStack {
+				return fmt.Errorf("re-entrant module calls not allowed for security reasons! module %s is already on the call stack", moduleName)
+			}
+
+			rtr.antiReentryMap[moduleName] = true
+			defer delete(rtr.antiReentryMap, moduleName)
```
robert-zaremba, line 83: This only works when we won't have async calls in `hander` (if handler is able to make a async call for other module , then this protection breaks apart.

As an alternative, we never delete from the map. Instead we create a new one once the app will start processing a new transaction.

aaronc: Yeah async currently was isn't allowed because there is no consensus way to do ordering.

types/module/server/router.go
```diff
@@ -0,0 +1,194 @@
...
+	if !found {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("cannot find method named %s", methodName))
+	}
+
+	moduleName := handler.moduleName
+
+	if writeCondition != nil && handler.commitWrites {
+		// msg handler
+		return func(ctx context.Context, request interface{}, response interface{}, opts ...interface{}) error {
+			if _, alreadyOnCallStack := rtr.antiReentryMap[moduleName]; alreadyOnCallStack {
```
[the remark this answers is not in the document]
aaronc: and it won't panic when the key isn't found?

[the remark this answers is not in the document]
robert-zaremba: No, maps return default values.
_An attempt to fetch a map value with a key that is not present in the map will return the zero value for the type of the entries in the map. For instance, if the map contains integers, looking up a non-existent key will return 0._

aaronc:
Any comments? @robert-zaremba @clevinson ?

Does this maybe look good enough to merge?

I know there are glaring omissions in terms of docs and unit tests (although most methods are tested by the x/data and x/ecocredit tests).

types/module/server/manager.go
```diff
@@ -0,0 +1,149 @@
...
+
+		serverMod.RegisterInterfaces(mm.cdc.InterfaceRegistry())
+	}
+
+	for _, mod := range modules {
+		// check if we actually have a server module, otherwise skip
+		serverMod, ok := mod.(Module)
+		if !ok {
+			continue
+		}
```
clevinson, line 54: Maybe i'm missing the rationale.. but it looks like this second for loop is uncessary, and should be combined with the for loop for lines 39-47? Lmk if i'm missing something and if there is a real need for running the `serverMod.RegisterInterfaces()` calls in a separate forloop before.

x/data/server/testsuite/suite.go
```diff
@@ -35,9 +35,7 @@ func (s *IntegrationTestSuite) SetupSuite() {
 	s.ctx = s.fixture.Context()
 	s.msgClient = data.NewMsgClient(s.fixture.TxConn())
 	s.queryClient = data.NewQueryClient(s.fixture.QueryConn())
-	if len(s.fixture.Signers()) < 2 {
-		s.FailNow("expected at least 2 signers, got %d", s.fixture.Signers())
-	}
+	s.Require().GreaterOrEqual(len(s.fixture.Signers()), 2)
```
clevinson, line 38: I know it's not something new as of this PR... but can you explain briefly why at least 2 signers are necessary in the x/data test suite?

clevinson (approved):
Great work on this @aaronc. I think i've got my head 90% wrapped around all the authorization logic (in `Invoker`, `InvokerFactory`, etc.)... but otherwise everything seems clear & makes sense.

Adding a bit more docs/comments would definitely help with some of the readability.

Few small other questions below, but approving as looks good to merge IMO.

types/module/server/manager.go
clevinson: Update: The separate forloop is prefered so that we can first propagate errors on interface registry failures before trying to do any other logic. A comment to explain this would be nice.

types/module/server/manager.go
```diff
@@ -0,0 +1,149 @@
...
+func (mm *Manager) CompleteInitialization() error {
+	for typ := range mm.requiredServices {
+		if _, found := mm.router.providedServices[typ]; !found {
+			return fmt.Errorf("initialization error, service %s was required, but not provided", typ)
+		}
+
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
aaronc: I have in mind this might do some more things later that are more like completion. Maybe we can address later

types/module/server/manager.go
```diff
@@ -0,0 +1,149 @@
...
+			router:       mm.router,
+			baseServer:   mm.baseApp.MsgServiceRouter(),
+			commitWrites: true,
+			moduleName:   name,
+		}
+
+		queryRegistrar := registrar{
+			router:       mm.router,
+			baseServer:   mm.baseApp.GRPCQueryRouter(),
+			commitWrites: true,
```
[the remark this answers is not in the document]
aaronc: Oh yes!!

x/data/server/testsuite/suite.go
aaronc: It's just the number of signers we use in test cases. If the test cases were written differently this would be different.

Issue #168, opened 2020-11-12
Query signatures by signer in x/data


aaronc:
It would be nice to be able to query all the pieces of data that an account is signed.

To do this we should likely store signatures as separate keys formed by `concat(cid, signer)` with a reverse index key `concat(signer, cid)`.

PR #171, opened 2020-11-12
Refactor data storage

aaronc:
ref: #168 

Refactor data storage so that signatures can be queried by both CID and signer by using prefix keys and reverse prefix keys.

All usage of ORM was removed for this module because it is a bit superfluous for this very simple case.


x/data/server/msg_server.go
```diff
@@ -38,62 +40,67 @@ func (s serverImpl) AnchorData(goCtx context.Context, request *data.MsgAnchorDat
...
 	}
 
+	store := ctx.KVStore(s.storeKey)
+	key := AnchorKey(cidBytes)
+	store.Set(key, bz)
+
 	return ctx.EventManager().EmitTypedEvent(&data.EventAnchorData{Cid: cidBytes})
 }
 
+var emptyBz = []byte{0}
```
clevinson, line 75: Any particular reason why this is set outside the function body?

clevinson (approved):
One small style nit. Otherwise lgtm.

x/data/server/msg_server.go
aaronc: Because it doesn't need to be allocated for every call

Issue #133, opened 2020-11-05
[x/ecocredit]: Add CLI commands


clevinson:
Create cli commands for ecocredit module functionality (both tx and query functionality, mirroring MsgServer and QueryServers)

clevinson:
Are there more CLI commands planned to be added to this? Seems like only `create credit_class` exists so far?

robert-zaremba:
Last week I finished on adding all query commands and `create credit_class`. @aaronc was coding yesterday - helped with interface registration and some other cleanups. So, I will add other commands tomorrow.

PR #172, opened 2020-11-13
ecocredits: add MsgSrv CLI

robert-zaremba:
closes: #133


robert-zaremba:
merge conflict solved.

x/ecocredit/client/tx.go
```diff
@@ -52,14 +57,139 @@ func txCreateClass() *cobra.Command {
 				return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "base64_metadata is malformed, proper base64 string is required")
```
clevinson, line 57:
```suggestion
				return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "metadata is malformed, proper base64 string is required")
```

Note to update this if we indeed end up going with "metadata" instead of "base64_metdata" for the param name. (same with line 53).

x/ecocredit/client/tx.go
```diff
@@ -52,14 +57,139 @@ func txCreateClass() *cobra.Command {
...
+func txCreateBatch() *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "create_batch [issuer] [class_id] [metadata] [issuance]",
+		Short: `Issues new credit batach.
+Parameters:
+  issuer:    issuer address
+  class_id:  credit class
+  metadata:  base64 encoded issuance metadata
+  issuance:  JSON encode issuance list,
+             eg: '[{"recipient": "a1", "tradeableUnits": "10", "retiredUnits": "2"}]'`,
```
clevinson, line 84: Feels like we could have a cleaner interface for this. Maybe something to explore after the collabathon.

For one- we may want to handle not setting tradableUnits or retiredUnits to be set to a default value of 0 there's no json value provided.

x/ecocredit/client/tx.go
```diff
@@ -52,14 +57,139 @@ func txCreateClass() *cobra.Command {
...
+		},
+	}
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
+
+func txSend() *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "send [recipient] [credits]",
+		Short: `Sends given credits to from the transaction author (--from) to the recipient.
```
clevinson, line 114:
```suggestion
		Short: `Sends credits from the transaction author (--from) to the recipient.
```

x/ecocredit/client/tx.go
```diff
@@ -52,14 +57,139 @@ func txCreateClass() *cobra.Command {
...
+	}
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
+
+func txRetire() *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "retire [holder] [credits]",
+		Short: `Retires specified amount of a holder credits. Transaction author (--from) must
+must be an authorized issuer.
```
clevinson, line 145: Do we actually verify that tx author is an issuer? It's not clear to me where this check happens.

Also- for user requirements I think we should also allow for users to retire their own credits (e.g. if --from == holder)

aaronc: Holders retire their own credits that is correct. Issuers shouldn't be mentioned here.

x/ecocredit/client/tx.go
```diff
@@ -23,18 +25,21 @@ func TxCmd() *cobra.Command {
...
+		txRetire(),
+		txSetPrecision(),
 	)
-
 	return cmd
 }
 
 func txCreateClass() *cobra.Command {
 	cmd := &cobra.Command{
 		Use: "create_class [designer] [issuer[,issuer]*] [base64_metadata]",
```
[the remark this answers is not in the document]
robert-zaremba: I also prefer to avoid the prefix.

x/ecocredit/client/tx.go
robert-zaremba: This is what happens. If the value is not set, it will go to default (0, "", nil...)

go.mod
```diff
@@ -11,10 +11,12 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/grpc-ecosystem/grpc-gateway v1.15.2
 	github.com/ipfs/go-cid v0.0.7
+	github.com/lib/pq v1.8.0 // indirect
```
aaronc, line 14: revert

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -128,11 +128,14 @@ message MsgSendResponse { }
 // MsgRetireRequest is the Msg/Retire request type.
 message MsgRetireRequest {
 
+    // issuer is the address of the batch issuer.
+    string issuer = 1;
+
     // holder is the credit holder address.
-    string holder = 1;
```
aaronc, line 132: revert

x/ecocredit/server/msg_server.go
```diff
@@ -249,9 +235,15 @@ func (s serverImpl) Retire(goCtx context.Context, req *ecocredit.MsgRetireReques
...
-
 	for _, credit := range req.Credits {
 		denom := batchDenomT(credit.BatchDenom)
+		classID, err := ecocredit.BatchID2ClassID(credit.BatchDenom)
+		if err != nil {
+			return nil, err
+		}
+		if err = s.assertIssuer(ctx, classID, req.Issuer); err != nil {
+			return nil, err
+		}
```
aaronc, line 246: revert

x/ecocredit/server/msg_server.go
```diff
@@ -295,30 +287,25 @@ func (s serverImpl) Retire(goCtx context.Context, req *ecocredit.MsgRetireReques
...
-	err := getAddAndSetDecimal(store, RetiredBalanceKey(recipient, batchDenom), retired)
+func (s serverImpl) SetPrecision(goCtx context.Context, req *ecocredit.MsgSetPrecisionRequest) (*ecocredit.MsgSetPrecisionResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	classID, err := ecocredit.BatchID2ClassID(req.BatchDenom)
 	if err != nil {
-		return err
+		return nil, err
+	}
+	if err = s.assertIssuer(ctx, classID, req.Issuer); err != nil {
+		return nil, err
```
aaronc, line 297: We just want to check that the issuer is the issuer of the batch. It should not be possible for any issuer to change the precision. Just the one who issued the batch. Based on this I don't see a need for `BatchID2ClassID` and `s.assertIssuer`. The only time this sort of check is needed is in `CreateBatch` and it doesn't need to be factored out into a separate function.

aaronc (changes requested):

go.mod
robert-zaremba: it has been added automatically. I've noticed it already several times.

robert-zaremba: removing.

x/ecocredit/server/msg_server.go
robert-zaremba: reverted

x/ecocredit/server/msg_server.go
robert-zaremba: I removed `BatchID2ClassID`, but I keep the `assertIssuer` - it's easier to read the code if we can break down some functions.

robert-zaremba: And I've updated the code as per your the spec here.

x/ecocredit/client/tx.go
robert-zaremba: BTW, I was thinking to use YAML instead of JSON. What do you think?

x/ecocredit/client/tx.go
```diff
@@ -49,17 +54,143 @@ func txCreateClass() *cobra.Command {
...
...It is an experimental feature. The number of decimal places allowed for a credit batch is determined by the original number of decimal places used with calling CreatBatch. SetPrecision allows the number of allowed decimal places to be increased, effectively making the supply more granular without actually changing any balances. It allows asset issuers to be able to issue an asset without needing to think about how many subdivisions are needed upfront. While it may not be relevant for credits which likely have a fairly stable market value, I wanted to experiment a bit and this serves as a proof of concept for a broader bank redesign where say for instance a coin like the ATOM or XRN could be issued in its own units rather than micro or nano-units. Instead an operation like SetPrecision would allow trading in micro, nano or pico in the future based on market demand. Arbitrary, unbounded precision is not desirable because this can lead to spam attacks (like sending 0.000000000000000000000000000001 coins). This is effectively fixed precision so under the hood it is still basically an integer, but the fixed precision can be increased so its more adaptable long term than just an integer.
```
clevinson, line 174:
```suggestion
		Short: "Allows an issuer to increase the decimal precision of a credit batch",
		Long: `Allows an issuer to increase the decimal precision of a credit batch. It is an experimental feature.

```

x/ecocredit/client/tx.go
```diff
@@ -49,17 +54,143 @@ func txCreateClass() *cobra.Command {
...
+	}
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
+
+func txRetire() *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "retire [credits]",
+		Short: `Retires specified amount of a holder credits. Transaction author (--from) must
+must be an authorized issuer.
```
clevinson, line 145:
```suggestion
		Short: "Retires a specified amounts of credits from the account of the transaction author (--from)",
		Long: `Retires a specified amounts of credits from the account of the transaction author (--from)

Parameters:
```

x/ecocredit/client/tx.go
```diff
@@ -49,17 +54,143 @@ func txCreateClass() *cobra.Command {
...
+		},
+	}
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
+
+func txSend() *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "send [recipient] [credits]",
+		Short: `Sends credits from the transaction author (--from) to the recipient.
```
clevinson, line 114:
```suggestion
		Short: "Sends credits from the transaction author (--from) to the recipient",
		Long: `Sends credits from the transaction author (--from) to the recipient.

```

x/ecocredit/client/tx.go
```diff
@@ -49,17 +54,143 @@ func txCreateClass() *cobra.Command {
...
+		},
+	}
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
+
+func txCreateBatch() *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "create_batch [issuer] [class_id] [metadata] [issuance]",
+		Short: `Issues new credit batach.
```
clevinson, line 78:
```suggestion
		Short: "Issues a new credit batch",
		Long: `Issues a new credit batch.

```

x/ecocredit/client/tx.go
```diff
@@ -23,18 +25,21 @@ func TxCmd() *cobra.Command {
...
 	)
-
 	return cmd
 }
 
 func txCreateClass() *cobra.Command {
 	cmd := &cobra.Command{
-		Use: "create_class [designer] [issuer[,issuer]*] [base64_metadata]",
+		Use: "create_class [designer] [issuer[,issuer]*] [metadata]",
 		Short: `Creates a new credit class.
```
clevinson, line 41:
```suggestion
		Short: "Creates a new credit class",
		Long: `Creates a new credit class.

```

clevinson (commented):
We should be using single line `--help` text for `Short`, and reserve longer explanations for `Long`

x/ecocredit/client/tx.go
```diff
@@ -49,17 +56,151 @@ func txCreateClass() *cobra.Command {
...
+		Use:   "create_batch [issuer] [class_id] [metadata] [issuance]",
+		Short: "Issues a new credit batch",
+		Long: `Issues a new credit batch.
+
+Parameters:
+  issuer:    issuer address
+  class_id:  credit class
+  metadata:  base64 encoded issuance metadata
+  issuance:  JSON encode issuance list,
+             eg: '[{"recipient": "a1", "tradeableUnits": "10", "retiredUnits": "2"}]'`,
```
clevinson, line 88:
```suggestion
  issuance:  JSON encode issuance list,
             eg: '[{"recipient": "a1", "tradableUnits": "10", "retiredUnits": "2"}]'
             Note: "tradableUnits" and "retiredUnits" default to 0.`,
```

clevinson: Add default note, and "tradeableUnits" should be "tradableUnits" (as is implemented in proto field name)

x/ecocredit/client/tx.go
```diff
@@ -49,17 +56,151 @@ func txCreateClass() *cobra.Command {
...
+func txSend() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "send [recipient] [credits]",
+		Short: "Sends credits from the transaction author (--from) to the recipient",
+		Long: `Sends credits from the transaction author (--from) to the recipient.
+
+Parameters:
+  recipient: recipient address
+  credits:   JSON encoded credit list
+             eg: '[{"batchDenom": "100/2", "tradeableUnits": "5", "retiredUnits": "0"}]'`,
```
clevinson, line 124:
```suggestion
             eg: '[{"batchDenom": "100/2", "tradableUnits": "5", "retiredUnits": "0"}]'`,
```

x/ecocredit/client/tx.go
```diff
@@ -23,18 +25,23 @@ func TxCmd() *cobra.Command {
...
 	)
-
 	return cmd
 }
 
 func txCreateClass() *cobra.Command {
 	cmd := &cobra.Command{
-		Use: "create_class [designer] [issuer[,issuer]*] [base64_metadata]",
-		Short: `Creates a new credit class.
+		Use:   "create_class [designer] [issuer[,issuer]*] [metadata]",
```
aaronc, line 40:
```suggestion
		Use:   "create-class [designer] [issuer[,issuer]*] [metadata]",
```

x/ecocredit/client/tx.go
```diff
@@ -49,17 +56,152 @@ func txCreateClass() *cobra.Command {
...
+			return c.send(err)
+		},
+	}
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
+
+func txCreateBatch() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create_batch [issuer] [class_id] [metadata] [issuance]",
```
aaronc, line 79:
```suggestion
		Use:   "create-batch [issuer] [class_id] [metadata] [issuance]",
```

x/ecocredit/client/tx.go
```diff
@@ -23,18 +25,23 @@ func TxCmd() *cobra.Command {
...
-		Use: "create_class [designer] [issuer[,issuer]*] [base64_metadata]",
-		Short: `Creates a new credit class.
+		Use:   "create_class [designer] [issuer[,issuer]*] [metadata]",
+		Short: "Creates a new credit class",
+		Long: `Creates a new credit class.
+
+Parameters:
   designer:  address of the account which designed the credit class
   issuer:    comma separated (no spaces) list of issuer account addresses. Example: "addr1,addr2"
   metadata:  base64 encoded metadata - arbitrary data attached to the credit class info`,
```
aaronc, line 47: In general I'd prefer if we had `--text-metadata` and `--binary-metadata` flags. JSON might be pretty common especially at this stage.

aaronc: But this can be in a separate PR

x/ecocredit/client/tx.go
```diff
@@ -49,17 +56,152 @@ func txCreateClass() *cobra.Command {
...
+func txCreateBatch() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create_batch [issuer] [class_id] [metadata] [issuance]",
+		Short: "Issues a new credit batch",
+		Long: `Issues a new credit batch.
+
+Parameters:
+  issuer:    issuer address
+  class_id:  credit class
+  metadata:  base64 encoded issuance metadata
```
aaronc, line 86: same as above regarding text and binary

x/ecocredit/client/tx.go
```diff
@@ -49,17 +56,152 @@ func txCreateClass() *cobra.Command {
...
+  issuance:  JSON encode issuance list,
```
aaronc, line 87: this is a little awkward in the CLI no? hopefully we get some actual user interfaces for this soon

x/ecocredit/client/tx.go
```diff
@@ -49,17 +56,152 @@ func txCreateClass() *cobra.Command {
...
+             eg: '[{"recipient": "a1", "tradableUnits": "10", "retiredUnits": "2"}]'
```
aaronc, line 88:
```suggestion
             eg: '[{"recipient": "xrn:sdgkjhs2345u79ghisodg", "tradableUnits": "10", "retiredUnits": "2"}]'
```

x/ecocredit/client/tx.go
```diff
@@ -49,17 +56,152 @@ func txCreateClass() *cobra.Command {
...
+}
+
+func txRetire() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "retire [credits]",
+		Short: "Retires a specified amounts of credits from the account of the transaction author (--from)",
+		Long: `Retires a specified amounts of credits from the account of the transaction author (--from)
+
+Parameters:
+  recipient: recipient address
```
aaronc, line 155:
```suggestion
```
this is unneeded right?

x/ecocredit/client/util.go
```diff
@@ -1,36 +1,54 @@
...
+	ctx, err := sdkclient.ReadQueryCommandFlags(ctx, cmd.Flags())
+	return ecocredit.NewQueryClient(ctx), ctx, err
 }
 
-func mkQueryClient(cmd *cobra.Command) (ecocredit.QueryClient, client.Context, error) {
-	ctx := client.GetClientContextFromCmd(cmd)
-	ctx, err := client.ReadQueryCommandFlags(ctx, cmd.Flags())
-	return ecocredit.NewQueryClient(ctx), ctx, err
+type msgSrvClient struct {
+	Cctx   *sdkclient.Context
```
aaronc, line 29:
```suggestion
	ClientContext   *sdkclient.Context
```

x/ecocredit/init_test.go
```diff
@@ -0,0 +1,9 @@
+package ecocredit
+
+import (
+	"testing"
+)
+
+func TestSuites(t *testing.T) {
+	// suite.Run(t, &ModelSuite{})
```
aaronc, line 8: ??

x/ecocredit/server/testsuite/suite.go
```diff
@@ -309,4 +308,47 @@ func (s *IntegrationTestSuite) TestScenario() {
 			}
 		})
 	}
+
+	/****   TEST SEST PRECISION   ****/
```
aaronc, line 312:
```suggestion
	/****   TEST SET PRECISION   ****/
```

x/ecocredit/server/testsuite/suite.go
```diff
@@ -309,4 +308,47 @@ func (s *IntegrationTestSuite) TestScenario() {
...
+			false,
+		}, {
+			"can increase", ecocredit.MsgSetPrecisionRequest{
+				Issuer: issuer1, BatchDenom: batchDenom, MaxDecimalPlaces: 8},
+			true,
+		}, {
+			"can NOT change precision of not existing denom", ecocredit.MsgSetPrecisionRequest{
+				Issuer: issuer1, BatchDenom: "not/existing", MaxDecimalPlaces: 1},
+			false,
+		},
```
aaronc, line 334: thank you 🙏

aaronc (approved):
Mostly LGTM.

Suggested some UX tweaks to the CLI cmd's. Those can be dealt with here or in separate PRs.

Pre-approving

aaronc:
Ideally we should use jsonpb @robert-zaremba. Which would mean decoding yaml to a map and then re-encoding as json then jsonpb decoding. Maybe not using jsonpb is okay for this simple case

x/ecocredit/client/tx.go
robert-zaremba: OK, let's discuss it in a separate task: https://github.com/regen-network/regen-ledger/issues/179

x/ecocredit/init_test.go
robert-zaremba: Yes, this is from the test I removed. I will remove this file.

x/ecocredit/client/tx.go
robert-zaremba: Yes, I the previous implementation I had this parameter. Thanks for catching.

x/ecocredit/client/util.go
robert-zaremba: The type name already has the full meaning: `ClientContext sdk.clientContext`. I can change to `ClientCtx` if you prefer.

x/ecocredit/client/tx.go
robert-zaremba: Why it's awkward? IMHO using a readable and widely used text format is the best choice for a list of complex object.

(btw, I've changed it to YAML)

clevinson: I think what @aaronc means is that stepping back, this kind of functionality is quite cumbersome to do via CLI and we long term should be prioritizing non-cli based tx interfaces for functionality like this with so many parameters.

x/ecocredit/client/tx.go
```diff
@@ -49,19 +61,143 @@ func txCreateClass() *cobra.Command {
...
+		Use:   "create-batch [issuer] [class_id] [metadata] [issuance]",
+		Short: "Issues a new credit batch",
+		Long: `Issues a new credit batch.
+
+Parameters:
+  issuer:    issuer address
+  class_id:  credit class
+  metadata:  base64 encoded issuance metadata
+  issuance:  YAML encode issuance list. Note: numerical values must be written in strings.
+             eg: '[{recipient: "xrn:sdgkjhs2345u79ghisodg", tradableUnits: "10", retiredUnits: "2"}]'
```
clevinson, line 91:
```suggestion
             eg: '[{recipient: "xrn:sdgkjhs2345u79ghisodg", tradableunits: "10", retiredunits: "2"}]'
```
Fixing to accurately reflect what the `yaml.Unmarshal()` expects

x/ecocredit/client/tx.go
```diff
@@ -49,19 +61,143 @@ func txCreateClass() *cobra.Command {
...
+
+func txRetire() *cobra.Command {
+	return &cobra.Command{
+		Use:   "retire [credits]",
+		Short: "Retires a specified amounts of credits from the account of the transaction author (--from)",
+		Long: `Retires a specified amounts of credits from the account of the transaction author (--from)
+
+Parameters:
+  credits:  YAML encoded credit list. Note: numerical values must be written in strings.
+            eg: '[{batchDenom: "100/2", units: "5"}]'`,
```
clevinson, line 155:
```suggestion
            eg: '[{batchdenom: "100/2", units: "5"}]'`,
```
Fixing to accurately reflect what the `yaml.Unmarshal()` expects

x/ecocredit/client/tx.go
```diff
@@ -49,19 +61,143 @@ func txCreateClass() *cobra.Command {
...
+func txSend() *cobra.Command {
+	return &cobra.Command{
+		Use:   "send [recipient] [credits]",
+		Short: "Sends credits from the transaction author (--from) to the recipient",
+		Long: `Sends credits from the transaction author (--from) to the recipient.
+
+Parameters:
+  recipient: recipient address
+  credits:   YAML encoded credit list. Note: numerical values must be written in strings.
+             eg: '[{batchDenom: "100/2", tradableUnits: "5", retiredUnits: "0"}]'`,
```
clevinson, line 126:
```suggestion
             eg: '[{batchdenom: "100/2", tradableUnits: "5", retiredUnits: "0"}]'`,
```
ditto

PR #173, opened 2020-11-13
Add Query/BySigner to x/data

### LINKED ISSUE #168: Query signatures by signer in x/data

aaronc:
Closes #168


x/data/server/keys.go
```diff
@@ -21,19 +20,29 @@ func CIDBase64String(cid []byte) string {
...
+	key := SignerCIDIndexPrefix(signer)
+	key = append(key, cid...)
+	return key
 }
 
 func SignerCIDIndexPrefix(signer string) []byte {
-	return append([]byte{SignerCIDPrefix}, []byte(fmt.Sprintf("%s/", signer))...)
+	key := []byte{SignerCIDPrefix}
+	key = append(key, signer...)
+	key = append(key, 0)
```
clevinson, line 44: Any particular reason for switching to `0` instead of `/` as a delimeter? If we expect users to ever be directly querying KVStore via tendermint RPC, then it does probably make sense for us to think of this as an API surface.

If we do switch here, should we also change other prefix deliminters (e.g. ecocredit's batchDenom prefixes also use `/` [here](https://github.com/regen-network/regen-ledger/blob/master/x/ecocredit/server/msg_server.go#L65))?

x/data/server/msg_server.go
```diff
@@ -100,7 +100,7 @@ func (s serverImpl) SignData(goCtx context.Context, request *data.MsgSignDataReq
 
 		store.Set(key, emptyBz)
 		// set reverse lookup key
-		store.Set(SignerCIDKey(signer, cidStr), emptyBz)
+		store.Set(SignerCIDKey(signer, cidBz), emptyBz)
```
clevinson, line 103: Why is this changed here? Shouldn't we choose one way, and keep it consistent across `SignerCIDKey()` and `CIDSignerKey()`? Currently `CIDSignerKey()` still takes `cidStr string`...

clevinson (commented):
Few small qs.. otherwise looks good.

x/data/server/keys.go
aaronc: Since both are strings, `0` is guaranteed to not be a string as it's the null terminator. `'/'` is a valid ASCII character

x/data/server/msg_server.go
aaronc: A string is not needed since this is the second part of the index key. It's an optimization. Maybe better to be consistent, but also good to optimize where possible no?

x/data/server/keys.go
aaronc: So this is a more generic way to separate two strings in an index key

third_party/proto/cosmos/base/query/v1beta1/pagination.proto
```diff
@@ -0,0 +1,50 @@
...
+message PageRequest {
+  // key is a value returned in PageResponse.next_key to begin
+  // querying the next page most efficiently. Only one of offset or key
+  // should be set.
+  bytes key = 1;
+
+  // offset is a numeric offset that can be used when key is unavailable.
+  // It is less efficient than using key. Only one of offset or key should
+  // be set.
+  uint64 offset = 2;
```
robert-zaremba, line 22: Is it possible to get just the next key instead of using `offset`?

x/data/server/keys.go
robert-zaremba: I saw somewhere a note to use `0` (probably in an ADR). It will be worth to add a documentation to this function and pointing to a specification if available.

x/data/server/msg_server.go
robert-zaremba: Yes!

x/data/server/testsuite/suite.go
```diff
@@ -92,6 +92,14 @@ func (s *IntegrationTestSuite) TestScenario() {
 	s.Require().Equal([]string{s.addr1.String()}, queryRes.Signers)
 	s.Require().Empty(queryRes.Content)
 
+	// query data by signer
+	bySignerRes, err := s.queryClient.BySigner(s.ctx, &data.QueryBySignerRequest{
+		Signer: s.addr1.String(),
+	})
+	s.Require().NoError(err)
+	s.Require().NotNil(bySignerRes)
+	s.Require().Contains(bySignerRes.Cids, cidBz)
```
robert-zaremba, line 101: I wonder, why do we want to use the long calls, rather than declaring a variable at the top of the function:
```
require := s.Require()
```

third_party/proto/cosmos/base/query/v1beta1/pagination.proto
robert-zaremba: Also, why don't we use `oneof` here?

proto/regen/data/v1alpha1/query.proto
```diff
@@ -32,4 +36,25 @@ message QueryByCidResponse {
...
+
+// QueryBySignerRequest is the Query/BySigner request type.
+message QueryBySignerRequest {
+
+    // signer is the address of the signer to query by.
+    string signer = 1;
+
+    // pagination is the PageRequest to use for pagination.
+    cosmos.base.query.v1beta1.PageRequest pagination = 2;
+}
```
robert-zaremba, line 50: NOTE: It's good to have it for verification, but in most of the production use-cases we will probably like to limit the blockchain load, and use off-chain DB for queries and analytics.

third_party/proto/cosmos/base/query/v1beta1/pagination.proto
aaronc: This was how the cosmos SDK pagination got implemented. My design I admit, but open to changing it

x/data/server/testsuite/suite.go
aaronc: 🤷🏻‍♂️

third_party/proto/cosmos/base/query/v1beta1/pagination.proto
aaronc: https://github.com/cosmos/cosmos-sdk/issues/6191

clevinson (approved):
🎉

PR #180, opened 2020-11-19
Use snake case for YAML encoding of proto fields

clevinson:
This PR adds `gogoproto.moretags` to appropriate `snake_case` proto fields, which enables better UX when interacting with the CLI. I've chosen `snake_case` over `camelCase` as that's what we use in other SDK proto files for yaml formatting.

UX Before:
```
regen tx ecocredit send $TO '[{batchdenom: "1Lb4WV1/1UnSkkA", tradableunits: "3", retiredunits: "5"}]' --from $FROM
```
UX After:
```
regen tx ecocredit send $TO '[{batch_denom: "1Lb4WV1/1UnSkkA", tradable_units: "3", retired_units: "5"}]' --from $FROM
```


amaurymartiny:
Could we use `OrigName: true` in jsonpb? https://godoc.org/github.com/golang/protobuf/jsonpb#Marshaler

amaurymartiny (approved):

robert-zaremba (approved):
good improvement. :+1:

PR #203, opened 2020-12-03
WIP on ADR 033 x/bank

aaronc:


types/module/server/manager.go
```diff
@@ -0,0 +1,147 @@
...
+		// check if we actually have a server module, otherwise skip
+		serverMod, ok := mod.(Module)
+		if !ok {
+			continue
+		}
+
+		serverMod.RegisterInterfaces(mm.cdc.InterfaceRegistry())
+	}
+
+	for _, mod := range modules {
```
jgimeno, line 47: you can use the same loop as before I think :)

aaronc: I want to make sure all interface types are registered first

clevinson:
@aaronc is this something that should be closed? Or are you planning to continue working on this?

aaronc:
We can close. Future work will happen in the SDK when we have enough dev bandwidth

PR #207, opened 2020-12-04
Implement custom protobuf service code generator

aaronc:


aaronc:
Would be great to get a review of this. I moved the WIP `Client` codegen to #213 so that this PR is more or less ready.

testutil/server/fixture.go
```diff
@@ -27,7 +27,7 @@ type FixtureFactory interface {
 // independent of the backend.
 type Fixture interface {
 
-	// Context is the context.Context to be used with gRPC generated client code.
+	// Context is the HasContext to be used with generated service client code.
```
amaurymartiny, line 30: What is HasContext?

amaurymartiny (approved):
lgtm

testutil/server/fixture.go
aaronc: Oops

Issue #209, opened 2020-12-07
[x/group] Add CLI support


blushi:
ref: https://github.com/regen-network/cosmos-modules/issues/61

Issue #197, opened 2020-11-25
[x/group] Add integration tests


blushi:
ref: [integration tests from cosmos-modules](https://github.com/regen-network/cosmos-modules/blob/master/incubator/group/integration_test.go).

Issue #215, opened 2021-01-06
[x/group] Events


blushi:
- Add remaining event types for proposals and votes
- We could make the ids (group, proposal...) just `uint64` (ref: https://github.com/regen-network/regen-ledger/pull/220#discussion_r576875675)
- We could also add more info to the existing events which are pretty minimalist right now (ref: https://github.com/regen-network/regen-ledger/pull/154#discussion_r552025354)

ref: https://github.com/regen-network/cosmos-modules/issues/43

aaronc:
In general I'm in favor of minimalist events which only contain indexable/observable fields, because the other info is elsewhere. So the question should really be - what is a client going to care to observe?

PR #220, opened 2021-01-15
[x/group] Add amino JSON signing support and CLI

blushi:
closes: #218
closes: #209 (not using ADR 031 approach for amino support)
closes: #197


x/group/client/testsuite/cli_test.go
```diff
@@ -0,0 +1,2047 @@
...
+	s.group = &group.GroupInfo{GroupId: group.ID(1), Admin: val.Address.String(), Metadata: []byte{1}, TotalWeight: "1", Version: 1}
+
+	// create 3 group accounts
+	for i := 0; i < 3; i++ {
+		out, err = cli.ExecTestCLICmd(val.ClientCtx, client.MsgCreateGroupAccountCmd(),
+			append(
+				[]string{
+					val.Address.String(),
+					"1",
+					"AQ==",
```
amaurym, line 108: what's AQ==? It's used multiple times, would it be useful to put it in a var?

x/group/module/module.go
```diff
@@ -43,26 +46,35 @@ func (a Module) DefaultGenesis(marshaler codec.JSONMarshaler) json.RawMessage {
...
-func (a Module) RegisterLegacyAminoCodec(*codec.LegacyAmino)    {}
+func (a Module) RegisterRESTRoutes(sdkclient.Context, *mux.Router) {}
+func (a Module) RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
+	group.RegisterLegacyAminoCodec(cdc)
+}
+
+// Route returns the message routing key for the group module.
+// func (a AppModule) Route() sdk.Route {
+// 	return sdk.NewRoute(group.RouterKey, server.NewHandler(a.storeKey, a.paramSpace, a.router, a.cdc))
+// }
```
amaurym, line 76: should be removed?

types/module/server/manager.go
```diff
@@ -98,6 +98,15 @@ func (mm *Manager) RegisterModules(modules []module.Module) error {
 
 		serverMod.RegisterServices(cfg)
 
+		// If mod implements LegacyRouteModule, registry module route.
```
amaurym, line 101:
```suggestion
		// If mod implements LegacyRouteModule, register module route.
```
```suggestion
		// If mod implements LegacyRouteModule, registry module route.
```

x/group/client/util.go
```diff
@@ -0,0 +1,28 @@
...
+	if membersFile == "" {
+		return members, nil
+	}
+
+	contents, err := ioutil.ReadFile(membersFile)
+	if err != nil {
+		return nil, err
+	}
+
+	err = json.Unmarshal(contents, &members)
```
amaurym, line 22: Given that `group.Member` only has string subfields, encoding/json works here, but technically it should be clientCtx.JSONMarshaler.UnmarshalJSON, right?

x/group/server/msg_server.go
```diff
@@ -82,8 +82,7 @@ func (s serverImpl) CreateGroup(ctx types.Context, req *group.MsgCreateGroupRequ
 		}
 	}
 
-	groupIDStr := util.Uint64ToBase58Check(groupID.Uint64())
-	err = ctx.EventManager().EmitTypedEvent(&group.EventCreateGroup{GroupId: groupIDStr})
+	err = ctx.EventManager().EmitTypedEvent(&group.EventCreateGroup{GroupId: strconv.FormatUint(groupID.Uint64(), 10)})
```
amaurym, line 85: Why don't we use GroupID uint64 in the typed event?

x/group/server/msg_server.go
```diff
@@ -98,6 +97,9 @@ func (s serverImpl) UpdateGroupMembers(ctx types.Context, req *group.MsgUpdateGr
 			return err
 		}
 		for i := range req.MemberUpdates {
+			if err := assertMetadataLength(req.MemberUpdates[i].Metadata, s.maxMetadataLength(ctx), "group member metadata"); err != nil {
```
amaurym, line 100: Not related to this PR, just curious: why is `maxMetadataLength` a function which takes sdk.Context, and not just a const?

x/group/client/testsuite/cli_test.go
```diff
@@ -0,0 +1,2047 @@
+package testsuite
```
amaurym, line 1: 2k lines for a file is becoming hard to navigate. How about splitting into 2 tests files: txs and queries?

x/group/client/testsuite/cli_test.go
```diff
@@ -0,0 +1,2047 @@
...
+
+func getTxSendFileName(s *IntegrationTestSuite, from string, to string) string {
+	tx := fmt.Sprintf(
+		`{"body":{"messages":[{"@type":"/cosmos.bank.v1beta1.MsgSend","from_address":"%s","to_address":"%s","amount":[{"denom":"%s","amount":"10"}]}],"memo":"","timeout_height":"0","extension_options":[],"non_critical_extension_options":[]},"auth_info":{"signer_infos":[],"fee":{"amount":[],"gas_limit":"200000","payer":"","granter":""}},"signatures":[]}`,
+		from, to, s.cfg.BondDenom,
+	)
+	return testutil.WriteToNewTempFile(s.T(), tx).Name()
+}
+
+func TestIntegrationTestSuite(t *testing.T) {
```
amaurym, line 2045: I think it could be useful to add one test using `--sign-mode=amino-json`, but to make sure the wiring works.

amaurym (commented):
Looks good, great job! Amino wiring looks good, but i'd like to see 1 test for peace of mind.

x/group/server/msg_server.go
blushi: this max metadata length used to be part of the `params` and this hasn't been cleaned up, will do that.

x/group/client/testsuite/cli_test.go
blushi: this is just some sample metadata, will put it in a var

x/group/server/msg_server.go
blushi: yes I guess we could do that, I'll file that in #215

x/group/client/util.go
blushi: good point

blushi: the thing is that `members` a just a regular slice, not `proto.Message` so I guess we might need some `Members` proto type in this case

anilCSE (approved):
@blushi @amaurym I am going ahead and merge this. Let's have follow-up PRs for the review comments

x/group/client/util.go
amaurym: ah yeah. That's annoying about protobuf, if you want an array of something you need to create a struct with an array subfield.

PR #221, opened 2021-01-18
Data module alpha2: Full schema

aaronc:
This PR removes support for IPFS CIDs and replaces it with a custom hash-based content identifiers that better suit our use cases.

I find IPFS CID's suboptimal and am removing them because:
* CID's are supposed to cover both the content format and hash but if you look at the [multicodec table](https://github.com/multiformats/multicodec/blob/master/table.csv) there are only a few data formats supported (very few of which apply to our use case) and lots of hash algorithms of which we will only support a few (for some reason `skein` hashes take up most of the table). Yes I could open PRs for every format we want, but the project just isn't that active, and also it feels wrong to list so many things in the codec which we have zero intention of supporting. Instead, I would rather choose the formats and hashes we think we'll use and not have a bunch of random extraneous stuff.
* I actually don't see a good way to support canonicalization algorithms given this specification, in particular RDF graphs. With canonicalization, a hash represents a canonical representation of some data which isn't tied to a specific serialization. So for RDF data we wouldn't really say "json-ld" or "n-quads" as the format, but rather we would say the canonicalization algorithm is URDNA2015 with a SHA-256 digest and json-ld or n-quads is something that can be used at the transport or storage layer but that's really up to specific implementations. Without going into too many details, it feels like this mental model is just a bad match for the way CID v1 is setup. Maybe they'll change it later, but that doesn't feel like _our_ problem given limited bandwidth.

What I done instead is define a custom `ID` message type with:
* an `IDType` enum specifying either:
  * "raw" data which has a corresponding `MediaType`,
  * RDF graph data, or
  * canonicalized geographic data
* a hash `DigestAlgorithm`
* a configurable `GraphCanonicalizationAlgorithm` for RDF data 

**High-level intentions of this design are to:**
* use the RDF graph data model as the basis for structured on-chain data to
  * align with W3C oriented work like DIDs, JSON-LD, Verifiable Credentials, and Linked Data Proofs
  * have a data format which already talks about canonicalization and multiple data representations rather than having to do that work on our own like we did with protobuf
  * ease coordination of on and off-chain data using RDF/"Linked Data" which is about distributed data to begin with
  * have flexible, extensible schemas and an easy mapping to database analytics via SPARQL/quad stores
* have canonical hashed representations of geographic data to ease coordination of on and off-chain data
* allow for a useful set of raw data formats to be hashed including PDFs, media files that may be associated with projects, and basic stuff like CSV and JSON files (although hopefully RDF will be preferred for that data)

Follow-ups after this:
* update implementation and remove `v1alpha1`
* specify URI mappings for IDs
* support on-chain graph data storage


proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,110 @@
...
+
+    // geospatial raster formats
+    // TODO
+
+    // geospatial vector formats
+    // TODO
+}
+
+enum GraphCanonicalizationAlgorithm {
+    GRAPH_CANONICALIZATION_ALGORITHM_URDNA2015_UNSPECIFIED = 0;
```
blushi, line 103: Shouldn't we have this instead?
```
GRAPH_CANONICALIZATION_ALGORITHM_UNSPECIFIED = 0;
GRAPH_CANONICALIZATION_ALGORITHM_URDNA2015 = 1;
```

aaronc: I was going for a sensible default rather than 0 being an error. I don't really have a strong opinion

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,110 @@
...
+    // ## **General Rules**
+    //
+    // - only the WGS 84 reference coordinate system is supported and all coordinates should be assumed to be in WGS 84
+    // - only 2D and 3D coordinates are supported, coordinates with linear referencing systems or an “M” component are unsupported
+    //
+    // ## **`POLYGON` Rules**
+    //
+    // - exterior linear rings must be defined in counterclockwise direction
+    // - interior linear rings must be defined in clockwise direction
+    // - where there is more than one ring, the first MUST be the exterior ring and other rings MUST be linear rings - this means that the area of a polygon MUST be continuous and unbroken. To represent multiple non-contiguous shapes, a `MULTIPOLYGON` should be used instead. Ex: in the case where inside of a hole in a polygon, there is an island that we want to include. Because the island is disjoint, this shape must be represented as a `MULTIPOLYGON` instead of a `POLYGON` with a hole and an island in the hole
```
clevinson, line 54:
```suggestion
    // - where there is more than one ring, the first MUST be the exterior ring and other rings MUST be interior rings - this means that the area of a polygon MUST be continuous and unbroken. To represent multiple non-contiguous shapes, a `MULTIPOLYGON` should be used instead. Ex: in the case where inside of a hole in a polygon, there is an island that we want to include. Because the island is disjoint, this shape must be represented as a `MULTIPOLYGON` instead of a `POLYGON` with a hole and an island in the hole
```

I think this should this read that non exterior rings must be interior rings

clevinson (commented):
Thanks for this @aaronc ! In general i'm very much in favor of us being more explicit about data types when thinking about anchoring/signing/storing functionality as described here. A few questions I have about the specific approach though:

1. What is our expectation of offchain data storage? With "CID"s from IPFS we get with the hash a mechanism to look up the actual content of the data on an offchain network. That is no longer the case here. @aaronc and I briefly discussed this on a call yesterday, and it seems like the strategy we might pursue for this is having a "RegisterDataResolver" type of msg that allows one to register a URL / IRI / etc. that can be used to retrieve the contents of a given `ContentID`.
2. For RDF data and polygon data, are there use cases in which we expect end users to only anchor some geometric data / RDF data, and not actually store it on chain? How would our approach differ if all RDF & geometric data were always expected to be stored (i imagine if we enforce this, it may be to our benefit). Would that change any of our design choices here? Perhaps the reason for still having a `ContentID` associated with a geopolygon and RDF graph is that it can then allow RDF's to reference other RDF graphs via the `ContentID` ?
3. If we look to concrete use cases, I imagine that "Geography" data is actually not data that lives in isolation, but rather it is more likely to exist as metadata associated with a piece of media data, other kind of raw data. Can you elaborate on a concrete use case for geography data being "signed" in isolation? What is that signing / timestamping of a geography object attesting to?

proto/regen/data/v1alpha2/types.proto
clevinson: One argument for having `UNSPECIFIED` be an error is that defaults may change over time. Thinking about the conversation you bring up [here](https://github.com/regen-network/regen-ledger/discussions/222), it seems like we're still evaluating what a sensible default may be?

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,110 @@
...
+    // choose from a variety of alternative formats for transport and encoding while maintaining the guarantee
+    // that the canonical hash will not change. The media type for "raw" data used with ID_TYPE_RAW_UNSPECIFIED
+    // is defined by the MediaType enum.
+    ID_TYPE_RAW_UNSPECIFIED = 0;
+
+    // ID_TYPE_GRAPH specifies graph data that conforms to the RDF data model.
+    // The canonicalization algorithm used for an RDF graph is specified by GraphCanonicalizationAlgorithm.
+    ID_TYPE_GRAPH = 1;
+
+    // ID_TYPE_GEOGRAPHY specifies geographic data conforming to this canonicalization algorithm:
```
blushi, line 41: why not defining the canonicalization algorithm for `ID_TYPE_GEOGRAPHY` types as part of the `ContentID` too? It could be relevant if we wanna use other canonical representations.

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,110 @@
...
+    MEDIA_TYPE_MPEG = 32;
+    MEDIA_TYPE_MP4 = 33;
+    MEDIA_TYPE_WEBM = 34;
+    MEDIA_TYPE_OGG = 35;
+
+    // geospatial raster formats
+    // TODO
+
+    // geospatial vector formats
+    // TODO
```
blushi, line 99: This means that we could support both geo data as `ID_TYPE_RAW_UNSPECIFIED` but also as `ID_TYPE_GEOGRAPHY`. Could you think of use cases where we'd use one over another? Could that be actually relevant to use both in some cases? That also relates a bit to point 3. of @clevinson's comment: https://github.com/regen-network/regen-ledger/pull/221#pullrequestreview-572931219.

blushi (commented):
I’m also in favor of having such a flexible architecture that handles a variety of data formats that are relevant to our use cases and provides a better support for RDF graphs and coordination between on and off-chain data.
Thinking more about user-facing features, what do we wanna provide in addition to anchor/sign/store to actually create the `ContentID`s in the first place? I guess that for RDF, we can move to [this discussion](https://github.com/regen-network/regen-ledger/discussions/222), what about the other types (although for those types, it seems less complex and computationally intensive)?

proto/regen/data/v1alpha2/types.proto
aaronc: I am thinking of `ID_TYPE_GEOGRAPHY` to represent a concrete polygon representing a project area.

But, we will likely also want annotated map data using a format like https://docs.mapbox.com/vector-tiles/reference/. I think we'd just want to store the raw hashed tile data rather than worrying about canonicalization.

I'm more thinking of `ID_TYPE_GEOGRAPHY` objects for use in signed claims, whereas map tile data, photos, videos, etc.  are more informational and less involved in verifying claims.

Whta do you think?

proto/regen/data/v1alpha2/types.proto
aaronc: I guess it could have its own canonicalization enum too...

proto/regen/data/v1alpha2/types.proto
aaronc: > Thinking more about user-facing features, what do we wanna provide in addition to anchor/sign/store to actually create the `ContentID`s in the first place? I guess that for RDF, we can move to [this discussion](https://github.com/regen-network/regen-ledger/discussions/222), what about the other types (although for those types, it seems less complex and computationally intensive)?

I'm thinking that signing is mainly relevant to RDF data and not raw/graph data. We generally just care about integrity and availability of that data. But all structure claims that would be signed would be RDF... So maybe we limit sign with that in mind? Also most raw data like PDFs, images, etc. would be too big to store on-chain (I'm thinking we'd have a 1kb on-chain storage limit or something).

aaronc:
> Thanks for this @aaronc ! In general i'm very much in favor of us being more explicit about data types when thinking about anchoring/signing/storing functionality as described here. A few questions I have about the specific approach though:
> 
> 1. What is our expectation of offchain data storage? With "CID"s from IPFS we get with the hash a mechanism to look up the actual content of the data on an offchain network. That is no longer the case here. @aaronc and I briefly discussed this on a call yesterday, and it seems like the strategy we might pursue for this is having a "RegisterDataResolver" type of msg that allows one to register a URL / IRI / etc. that can be used to retrieve the contents of a given `ContentID`.

So I think we could still use IPFS, we'd just maybe specify that all content types map to `raw` in IPFS expect for maybe JSON.

We'd also want to specify an API (maybe in gRPC) for services which store content-addressable data, possibly with proofs for querying parts of graphs.

> 2. For RDF data and polygon data, are there use cases in which we expect end users to only anchor some geometric data / RDF data, and not actually store it on chain? How would our approach differ if all RDF & geometric data were always expected to be stored (i imagine if we enforce this, it may be to our benefit). Would that change any of our design choices here? Perhaps the reason for still having a `ContentID` associated with a geopolygon and RDF graph is that it can then allow RDF's to reference other RDF graphs via the `ContentID` ?

Yeah, basically this design allows graph data to refer to other graph or geography data whether this lives on or off chain. Credit classes might require data to live on-chain in their metadata schemas.

> 3. If we look to concrete use cases, I imagine that "Geography" data is actually not data that lives in isolation, but rather it is more likely to exist as metadata associated with a piece of media data, other kind of raw data. Can you elaborate on a concrete use case for geography data being "signed" in isolation? What is that signing / timestamping of a geography object attesting to?

Geographic data generally wouldn't be signed in isolation. Its a separate type because its large (compared to graphs) and could be reused between multiple graphs... A graph might be ~1kb vs a polygon that's 100-200kb depending on the size of land and sampling resolution.

aaronc:
So one thing I'm thinking is where `Msg/Sign` should really only apply to graphs... because other formats can't necessarily be implied to have implicit meaning. Except maybe PDFs?

proto/regen/data/v1alpha2/types.proto
clevinson: I'm in favor of liminting signing capabilities to graph data.  Signing always represents attesting a claim, and in RDF land this just means claims are always represented as triples. If someone wants to attest to the veracity of a PDF, they might as well attest to the connection of that PDF to a given project or credit class, which would be stored onchain as an rdf graph.

clevinson:
> So one thing I'm thinking is where Msg/Sign should really only apply to graphs... because other formats can't necessarily be implied to have implicit meaning. Except maybe PDFs?

Makes sense. I think I agree with this logic. I'd rather not have exceptions for specific media types (e.g. PDF), but rather just choose one strategy & stick with it.

In general my inclination is to pick something that satisfies our most basic needs & use cases, and iterate later as our needs evolves.

@aaronc would you like to include `MsgStoreRDFData` and `MsgStoreGraphData` definitions in this PR as well? Or tackle those in a follow up ?

aaronc:
> @aaronc would you like to include `MsgStoreRDFData` and `MsgStoreGraphData` definitions in this PR as well? Or tackle those in a follow up ?

Is it helpful to see the full picture? Is so then yes I can

proto/regen/data/v1alpha2/types.proto
blushi: Yeah it makes sense. There might be cases where some data could be used both for verifying claims and for informational purposes, e.g. a project boundaries. In this case, should we just use `ID_TYPE_GEOGRAPHY`?

aaronc:
I made a few updates to show the bigger picture. There are a few more things that I have in mind which I'll try to get to later today

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,209 @@
+syntax = "proto3";
+
+package regen.data.v1alpha2;
+
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+
+option go_package = "github.com/regen-network/regen-ledger/x/data";
+
+// ContentID specifies a hash based content identifier for a piece of data
```
clevinson, line 10:
```suggestion
// ContentHash specifies a hash based content identifier for a piece of data
```

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    message Literal {
+        Datatype datatype = 1;
+
+        oneof value {
+            string str_value = 3;
+            google.protobuf.Timestamp timestamp_value = 4;
+            google.protobuf.Duration duration_value = 5;
+            uint64 geo_id = 6;
```
clevinson, line 170: What do you mean by `geo_id` here?

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,209 @@
...
+    message Geo {
+        // hash represents the hash of the data based on the specified digest_algorithm
+        bytes hash = 1;
+
+        // digest_algorithm represents the hash digest algorithm.
+        DigestAlgorithm digest_algorithm = 2;
+    }
+}
+
+enum IDType {
```
clevinson, line 53: Is this enum still necessary?

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,209 @@
...
+        bytes hash = 1;
+
+        // digest_algorithm represents the hash digest algorithm.
+        DigestAlgorithm digest_algorithm = 2;
+
+        // graph_canonicalization_algorithm represents the RDF graph canonicalization algorithm. It should
+        // be left unset if type is not ID_TYPE_GRAPH.
+        GraphCanonicalizationAlgorithm canonicalization_algorithm = 3;
+
+        GraphMerkleTree merkle_tree = 4;
```
clevinson, line 41: What does this value actually tell us about the hash? Does the hash then represent the merkle tree root? I think i'm a little confused as to what different values of this enum actually mean in practice for how we handle a given Graph's content hash.

clevinson (commented):
Few questions I have on your most recent updates. These are all probably things that we can discuss quickly on our call tomorrow.

aaronc:
@clevinson @blushi so the thing I was trying to explain there at the end is that right now I have three ways of representing each piece of data:
* IRIs - these exist in the actual graphs when users see them and when they're canonicalized
* `uint64` IDs - these would be auto-generated for all IRIs including ones like `https://regen.network/schema#...` and are used to save space when storing graphs on chain and sending `Msg`s
* `ContentHash` this is really a descriptor for what gets serialized into the IRIs

We could maybe get rid of `ContentHash` from most APIs and just have users construct IRIs themselves, although I think it's useful to have `ContentHash` to explicitly construct IRIs with the right fields set.

The `uint64` IDs feel useful to reduce storage space and just having fixed 8-byte index keys will make managing the KV store a lot simpler. We could maybe mostly remove them from user-facing APIs and just use IRIs and `ContentHash`s, but I do like the idea of this built-in compression which is why this API uses `uint64`s pretty often.

The workflow for constructing one fo these compressed `Graph` data structures which uses `uint64` IDs shouldn't be so bad. Here's what someone would do:
1. create a dataset in JSON-LD and compute the URDNA2015 hash on the client side
2. call `Query/ConvertToCompactDataset` with the JSON-LD content and this will return a `Graph.Dataset` with `uint64` IDs filled in
3. (if needed) If some IRIs are not registered, `Query/ConvertToCompactDataset` will return a list of those IRIs and instruct the user to call `MsgRegisterIRIs` with those IRIs. Then the user will be able to call `Query/ConvertToCompactDataset` without errors
4. call `Msg/StoreGraphData` with the compact `Graph.Dataset` and the URDNA2015 hash generated client side. This way there is efficiency at the storage and transport layer with no need for the user to fully trust a node because of the client-side URDNA2015 hash

Does that seem reasonable? Thoughts on this API design and usage of these three types of identifiers in different places?

aaronc:
Also in terms of URI construction, we could also just always use an extension and hide everything else beneath base58 encoding, so for a dataset we could actually have `regen:6EqBg1puBvfGDxMSbxah9R.rdf` instead of `regen:g/6EqBg1puBvfGDxMSbxah9R`. We could maybe use a `.geo` or `.wkt` extension for geography and the regular extension for other raw data types... Then maybe IRIs would be a bit more consistent. What do you think?

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,209 @@
...
+        // hash represents the hash of the data based on the specified digest_algorithm
+        bytes hash = 1;
+
+        // digest_algorithm represents the hash digest algorithm.
+        DigestAlgorithm digest_algorithm = 2;
+
+        // graph_canonicalization_algorithm represents the RDF graph canonicalization algorithm. It should
+        // be left unset if type is not ID_TYPE_GRAPH.
+        GraphCanonicalizationAlgorithm canonicalization_algorithm = 3;
+
```
clevinson, line 40:
```suggestion
        // merkle_tree represents the optional merkle tree algorithm used for proving that
        // a given RDF triple exists in this graph. If set, the merkle root is stored in the `hash` field
```

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,209 @@
...
+        // digest_algorithm represents the hash digest algorithm.
+        DigestAlgorithm digest_algorithm = 2;
+
+        // media_type represents the MediaType for raw data.
+        MediaType media_type = 3;
+
+    }
+
+    message Graph {
+        // hash represents the hash of the data based on the specified digest_algorithm
```
clevinson, line 31:
```suggestion
        // hash represents the hash of the data based on the specified digest_algorithm.
        // if merkle_tree is set, then `hash` represents the merkle root of the Graph as
        // calculated using the specified merkle tree algorithm
```

proto/regen/data/v1alpha2/types.proto
clevinson: Added some comment suggestions to clarify this.

clevinson (commented):
Added some clarifying comments for GraphMerkleTree.

clevinson:
> The uint64 IDs feel useful to reduce storage space and just having fixed 8-byte index keys will make managing the KV store a lot simpler. We could maybe mostly remove them from user-facing APIs and just use IRIs and ContentHashs, but I do like the idea of this built-in compression which is why this API uses uint64s pretty often.

Having users manage the cognitive overhead of two separate IDs as well as manually registering all IRIs feels like a lot of complexity to be introducing to end users. If the concern is just about storage, why can't we just use something under the hood that doesn't get exposures to users at all (similar to how we handle protobuf any compression in sdk). 

I feel like we're already asking a lot of end users to be learning the world of RDF / SHACL (and for good reason), and would love to keep additional complexity / DevX burdon to a minimum.

Can we get most of the performance / storage benefits you desire by taking a more under-the-hood approach to compression?

aaronc:
Well we can't compress the transactions that are in the tendermint block store. And compared to protobuf, these IRIs make up the primary content of RDF. And even if we got rid of the uint64 IDs, I wouldn't feel comfortable directly parsing json-ld or turtle in the state machine. I mean maybe we could, but it wouldn't necessarily be less work to implement and I want to limit on chain state machine logic to the minimum. A shacl subset already feels relatively complex. So I think there would be a conversion step from json ld to another format either way.

aaronc:
One thing we _could_ do is only use uint64 IDs in the compact graph format for StoreGraphData and other methods would only use ContentHash or the IRIs. Maybe as a starting point I can make that change. Wdyt?

aaronc:
Actually I think I've thought of a way to hide most of the complexity of the uint64 IDs. I'll update this PR with the new design soon.

aaronc:
> 1. create a dataset in JSON-LD and compute the URDNA2015 hash on the client side
> 2. call `Query/ConvertToCompactDataset` with the JSON-LD content and this will return a `Graph.Dataset` with `uint64` IDs filled in
> 3. (if needed) If some IRIs are not registered, `Query/ConvertToCompactDataset` will return a list of those IRIs and instruct the user to call `MsgRegisterIRIs` with those IRIs. Then the user will be able to call `Query/ConvertToCompactDataset` without errors
> 4. call `Msg/StoreGraphData` with the compact `Graph.Dataset` and the URDNA2015 hash generated client side encoded into `ContentHash.Graph`. This way there is efficiency at the storage and transport layer with no need for the user to fully trust a node because of the client-side URDNA2015 hash
> 5. `Msg/StoreGraphData` will return a `uint64` ID which the user can use for other calls like `Msg/ValidateGraph` if needed. The IRI internally representing this graph can also be retrieved with `Query/IRIsByID`.

So now I've simplified this to the follow:

1. create a dataset in JSON-LD and compute the URDNA2015 hash on the client side
2. call `Query/ConvertToCompactDataset` with the JSON-LD content and this will return a `CompactDataset`
3. call `Msg/StoreGraphData` with the compact `CompactDataset` and the URDNA2015 hash generated client side encoded into `ContentHash.Graph`. This way there is efficiency at the storage and transport layer with no need for the user to fully trust a node because of the client-side URDNA2015 hash
4. `Msg/ValidateGraph` now uses `ContentHash.Graph` to identify the data and shapes graphs, with optional named graph strings in case we're pointing to a graphs within datasets

So now:
* all of the `Msg` APIs now use `ContentHash.Graph` to refer to graphs
* the `uint64` IDs are hidden behind `CompactDataset` for most use cases, with a few utility methods which most users wouldn't need to use
* IRIs only get exposed on the query side

Let me know what you think @clevinson.

One thought I have is whether the IRIs should play a role at all in the `Msg` APIs - maybe in the return types, but maybe it's fine as it is and they mostly just get exposed on the query side of things.

proto/regen/data/v1alpha2/query.proto
```diff
@@ -0,0 +1,152 @@
...
+service Query {
+  // ByContentHash queries data based on its ContentHash.
+  rpc ByContentHash (QueryByContentHashRequest) returns (QueryByContentHashResponse);
+
+  // BySigner queries data based on signers.
+  rpc BySigner (QueryBySignerRequest) returns (QueryBySignerResponse);
+
+  rpc IDsByIRI (QueryIDsByIRIRequest) returns (QueryIDsByIRIResponse);
+
+  rpc IRIsByID (QueryIRIsByIDRequest) returns (QueryIRIsByIDResponse);
```
blushi, line 21: in which context those might be useful given https://github.com/regen-network/regen-ledger/pull/221#issuecomment-768730796?

aaronc: They're utility methods is you want to build a CompactGraph in client code, which generally isn't needed but should be possible

clevinson:
This is on hold until we revisit on-chain storage

PR #229, opened 2021-01-28
Data module alpha2: Part 1

aaronc:
Starting from #221:

* `StoreData` -> `StoreRawData`
* just update `AnchorData`, `SignData` and `StoreRawData` to use `ContentHash`
* `Query/ByCid` -> `QueryByHash`
* added genesis
* commented out all implementation and tests

Did not add:
* SMT merkle tree because the format should be better specified up front in the docs
* `ContentHash.Geo` in case more discussion is needed


proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,184 @@
...
+    // Graph is the content hash type used for RDF graph data
+    message Graph {
+        // hash represents the hash of the data based on the specified digest_algorithm
+        bytes hash = 1;
+
+        // digest_algorithm represents the hash digest algorithm.
+        DigestAlgorithm digest_algorithm = 2;
+
+        // graph_canonicalization_algorithm represents the RDF graph canonicalization algorithm. It should
+        // be left unset if type is not ID_TYPE_GRAPH.
```
clevinson, line 48:
```suggestion
        // graph_canonicalization_algorithm represents the RDF graph canonicalization algorithm.
```

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,184 @@
...
+        GraphCanonicalizationAlgorithm canonicalization_algorithm = 3;
+
+        // merkle_tree is the merkle tree type used for the graph hash, if any
+        GraphMerkleTree merkle_tree = 4;
+    }
+}
+
+// MediaType defines MIME media types to be used with ID_TYPE_RAW_UNSPECIFIED.
```
clevinson, line 56:
```suggestion
// MediaType defines MIME media types to be used with a ContentHash.Raw hash.
```

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,184 @@
...
+    DIGEST_ALGORITHM_BLAKE2B_256 = 1;
+}
+
+// ContentEntry describes data referenced and possibly stored on chain
+message ContentEntry {
+    // hash is the content hash
+    ContentHash hash = 1;
+
+    // iri is the content IRI
+    string iri = 2;
```
clevinson, line 155: Does every ContentEntry have an IRI separate from something that can be deterministically generated from the ContentHash? What's the rational for having iri stored separately at the protobuf message layer?

proto/regen/data/v1alpha2/tx.proto
```diff
@@ -36,35 +36,30 @@ service Msg {
...
   // sender is the address of the sender of the transaction.
   // The sender in StoreData is not attesting to the veracity of the underlying
   // data. They can simply be a intermediary providing services.
   string sender = 1;
 
-  // cid is a Content Identifier for the data corresponding to the IPFS CID
-  // specification: https://github.com/multiformats/cid.
-  bytes cid = 2;
+  // id is the hash-based identifier for the anchored content.
+  ContentHash id = 2;
```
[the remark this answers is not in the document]
clevinson: Or is it because this is a `oneof` (descriptor) that you don't actually want to call field "hash" ?

clevinson (approved):
Handful of small comment suggestions. Approving as all are quite minor and non-blocking IMO.

Only substantial question i had was the relationship btw ContentHash / IRIs, Should that be made explicit anywhere in hte protobuf comments as well?

proto/regen/data/v1alpha2/events.proto
```diff
@@ -0,0 +1,28 @@
...
+package regen.data.v1alpha2;
+
+import "regen/data/v1alpha2/types.proto";
+
+option go_package = "github.com/regen-network/regen-ledger/x/data";
+
+// EventAnchorData is an event emitted when data is anchored on-chain.
+message EventAnchorData {
+    // iri is the data IRI
+    string iri = 1;
```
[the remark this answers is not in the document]
aaronc: The details will come in the next PR tomorrow. We will need to update the docs though...

proto/regen/data/v1alpha2/types.proto
aaronc: This type is for queries - maybe I should move it. Yes it is deterministic but the idea is to save the user trouble. Only one will get stored in state.

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,182 @@
...
+    // geospatial raster formats
+    // TODO
+
+    // geospatial vector formats
+    // TODO
+}
+
+// GraphCanonicalizationAlgorithm is the graph canonicalization algorithm
+enum GraphCanonicalizationAlgorithm {
+    // unspecified
```
blushi, line 126: maybe add in the comment that this is invalid.

proto/regen/data/v1alpha2/types.proto
```diff
@@ -0,0 +1,182 @@
...
+
+// GraphMerkleTree is the graph merkle tree type used for hashing, if any
+enum GraphMerkleTree {
+    // no merkle tree
+    GRAPH_MERKLE_TREE_NONE_UNSPECIFIED = 0;
+}
+
+// DigestAlgorithm is the hash digest algorithm
+enum DigestAlgorithm {
+    // unspecified
```
blushi, line 141: ditto

blushi (approved):
Pre-approving as I just have minor comments.

PR #230, opened 2021-01-29
Data module alpha 2: Part 2

aaronc:
Adds:

* validation for `MsgRequest` types, `ContentHash`, etc.
* IRI generation for `ContentHash` and `AccAddress`es

with tests


x/data/types_test.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			chr := ContentHash_Raw{
+				Hash:            tt.fields.Hash,
+				DigestAlgorithm: tt.fields.DigestAlgorithm,
+				MediaType:       tt.fields.MediaType,
+			}
+			if err := chr.Validate(); (err != nil) != tt.wantErr {
+				t.Errorf("Validate() error = %v, wantErr %v", err, tt.wantErr)
```
clevinson, line 96: For my own understanding- what's the reason for using `t.Errorf()` here, but `require.Error()` / `require.NoError()` in other tests below?

clevinson (approved):
ACK 🎉 

Few minor clarifying q's (one suggestion). Nothing major. Great to see the solid test coverage here.

blushi (approved):
Looks good to me overall. Just some small comments.
Although their implementation is pretty straightforward, I would also add some docs to the `ToIRI` methods to describe how IRIs are constructed.

x/data/iri.go
```diff
@@ -0,0 +1,103 @@
...
+	bz[1] = byte(chr.DigestAlgorithm)
+	copy(bz[2:], chr.Hash)
+	hashStr := base58.CheckEncode(bz, iriVersion0)
+
+	ext, err := chr.MediaType.ToExtension()
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("regen:%s.%s", hashStr, ext), nil
```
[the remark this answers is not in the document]
aaronc: well in this case I think we actually want the full format string because it gets compiled to a constant whereas doing something like `RegenConstant + "%s.%s"` leads to extra memory at runtime. does that make sense? or were you talking about something different?

x/data/requests.go
```diff
@@ -38,7 +46,29 @@ func (m *MsgSignDataRequest) GetSigners() []sdk.AccAddress {
...
+	switch digestAlgorithm {
+	case DigestAlgorithm_DIGEST_ALGORITHM_BLAKE2B_256:
+		hash := crypto.BLAKE2b_256.New()
+		_, err = hash.Write(m.Content)
+		if err != nil {
+			return sdkerrors.Wrap(ErrHashVerificationFailed, err.Error())
+		}
+
+		digest := hash.Sum(nil)
+		if !bytes.Equal(m.Hash.Hash, digest) {
```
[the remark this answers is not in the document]
aaronc: done

x/data/types_test.go
aaronc: `t.Errorf` just fails the test with a formatted error message

`require.NoError` fails the test if and only if there is an error

Does that make sense?

aaronc:
Added docs on `ToIRI` @blushi. Let me know if that looks okay.

Going ahead and merging

x/data/iri.go
```diff
@@ -0,0 +1,113 @@
...
+func (ch ContentHash) ToIRI() (string, error) {
+	switch hash := ch.Sum.(type) {
+	case *ContentHash_Raw_:
+		return hash.Raw.ToIRI()
+	case *ContentHash_Graph_:
+		return hash.Graph.ToIRI()
+	default:
+		return "", fmt.Errorf("invalid %T type %T", ch, hash)
+	}
+}
```
robert-zaremba, line 31: Why do we call it `IRI` instead of `URI`? We use only ASCII characters, so the output is always `URI`.

x/data/types_test.go
robert-zaremba: I think Cory question was more about an approach:
```
if wantErr {
   assert.Error(err)
} else {
   assert.NoError(err)
}
```
vs 
```
if (err != nil) != tt.wantErr {
   t.Errorf("Validate() error = %v, wantErr %v", err, tt.wantErr)
}
```

x/data/iri.go
aaronc: Because IRI is the terminology used in all RDF contexts

x/data/types_test.go
aaronc: Ah sorry...

PR #236, opened 2021-02-04
feat: data module alpha2: part 4

aaronc:
* implements `MsgServer` and `QueryServer` with test suites
* adds `ParseIRI`


x/data/server/query_server.go
```diff
@@ -10,55 +15,106 @@ import (
...
+
+	for iterator.Valid() {
+		signer := sdk.AccAddress(iterator.Key())
+		var timestamp gogotypes.Timestamp
+		err = timestamp.Unmarshal(iterator.Value())
+		if err != nil {
+			return nil, err
+		}
+
+		signerEntries = append(signerEntries, &data.SignerEntry{
```
clevinson, line 61: Just thinking about this now – is there an attack vector in someone appending an arbitrary number of signatures to a given ContentHash, and this query getting bloated / slow? especially without pagination of signers...

x/data/server/query_server.go
```diff
@@ -10,55 +15,106 @@ import (
...
+	signerIDStore := prefix.NewStore(store, SignerIDIndexPrefix(addr))
+
+	var entries []*data.ContentEntry
+	pageRes, err := query.Paginate(signerIDStore, request.Pagination, func(key []byte, value []byte) error {
+		entry, err := s.getEntry(store, key)
+		if err != nil {
+			return err
+		}
+
+		entries = append(entries, entry)
```
clevinson, line 109: So this looks like this query response will return a paginated list of entries, but each entry could have an unbounded number of signers inside the ContentEntry payload?  Maybe that's fine... but just flagging incase you see that as a potential problem down the line. Not sure what alternatives we have... maybe the `BySigner()` call shouldn't be expected to return ContentEntry's with full lists of all signers inside?

Just a thought.. ignore if you see this as a non-issue.

x/data/server/testsuite/suite.go
```diff
@@ -41,106 +42,168 @@ func (s *IntegrationTestSuite) TearDownSuite() {
...
+
+	// anchoring same data twice is a no-op
+	_, err = s.msgClient.AnchorData(s.ctx, &data.MsgAnchorDataRequest{
+		Sender: s.addr1.String(),
+		Hash:   contentHash,
+	})
+	s.Require().NoError(err)
+
+	// can query data and get timestamp
+	queryRes, err := s.queryClient.ByHash(s.ctx, &data.QueryByHashRequest{
```
clevinson, line 74: Later do we expect to add a querier for `ByIRI`? Not for now, but perhaps in a follow-up. Seems like that would be useful, unless you don't expect app developers to be interacting with IRIs much themselves...

x/data/server/testsuite/suite.go
```diff
@@ -41,106 +42,168 @@ func (s *IntegrationTestSuite) TearDownSuite() {
...
+		Hash: contentHash,
+	})
+	s.Require().NoError(err)
+	s.Require().NotNil(queryRes)
+	s.Require().NotNil(queryRes.Entry)
+	ts := queryRes.Entry.Timestamp
```
clevinson, line 80: It would be nice in this test to actually verify the timestamp is accurate (especially as we are calling `AnchorData()` multiple times and asserting the second is a no-op.

x/data/server/keys.go
```diff
@@ -1,50 +1,56 @@
...
 const (
-	AnchorTablePrefix byte = 0x0
-	CIDSignerPrefix   byte = 0x1
-	SignerCIDPrefix   byte = 0x2
-	DataTablePrefix   byte = 0x3
+	IriIDTablePrefix      byte = 0x0
+	AnchorTimestampPrefix byte = 0x1
+	IDSignerPrefix        byte = 0x2
+	SignerIDPrefix        byte = 0x2
+	RawDataPrefix         byte = 0x3
```
clevinson, line 10: Shouldn't `IDSignerPrefix` and `SignerIDPrefix` be distinct ?
```suggestion
	IDSignerPrefix        byte = 0x2
	SignerIDPrefix        byte = 0x3
	RawDataPrefix         byte = 0x4
```

clevinson (changes requested):
Nice! I'm seeing the pieces really come together here.

Some small non-blocking nits & clarifying q's, but there are also a few pieces in particular that i'd like to see responses on before approving:
- store key prefix bytes are not all distinct (i think this is a type-o)
- adding some clarification to why ParseIRI takes a generic `methodPrefix` when handling DID's, yet only seems to be able to return sdk.AccAddresses if the IRI is a DID
- i don't understand why `getIRIAndAnchor` always returns current blocktime, can you clarify if this is intentional, and if so why?

Otherwise, all looks good~!

x/data/server/testsuite/suite.go
```diff
@@ -41,106 +42,168 @@ func (s *IntegrationTestSuite) TearDownSuite() {
...
+	s.Require().NotNil(queryRes)
+	s.Require().Equal(ts, queryRes.Entry.Timestamp)
+	s.Require().Len(queryRes.Entry.Signers, 2)
+	signers := make([]string, len(queryRes.Entry.Signers))
+	for _, signer := range queryRes.Entry.Signers {
+		signers = append(signers, signer.Signer)
+	}
+	s.Require().Contains(signers, s.addr1.String())
+	s.Require().Contains(signers, s.addr2.String())
+	s.Require().Empty(queryRes.Entry.Content)
```
robert-zaremba, line 142: why we don't declare 
```
require := s.Require()
```

at the beginning of the function - it would make the code easier to read.

x/data/iri.go
```diff
@@ -111,3 +115,128 @@ var mediaTypeExtensions = map[MediaType]string{
...
+
+var mediaTypeExtensionsReverse = map[string]MediaType{}
+
+func init() {
+	for mt, ext := range mediaTypeExtensions {
+		mediaTypeExtensionsReverse[ext] = mt
+	}
+}
+
+func ParseIRI(methodPrefix string, iri string) (*ContentHash, sdk.AccAddress, error) {
```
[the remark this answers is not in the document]
robert-zaremba: I second that. Would be good to explain what's the expected iri, eg:
```
regen:<hashPart>.<ext>
```
What's `hashaPart` represents? It' seams it's more than a hash of something  - it's base58 with IRI version and a type. And we validate the last part of the hash: `digestAlg.Validate(hash)` - normally we validate hash against some other value. So, maybe we should rename it.

robert-zaremba (commented):
Left few comments, also Cory has few valid points re readability and 2 places with potential bugs.

aaronc:
Thanks for your reviews @robert-zaremba and @clevinson.

On a high-level I'd actually like some review of the IRI format we're using. Of all things related to this data module, we need to more or less look the IRI format in by mainnet launch because that will be used in documents and shouldn't change too much.

For instance, what do we even think about `regen:` as an IRI scheme? Should it be something more generic like a `hash:` IRI scheme? What about the way it's structured using extensions and specifically the `.rdf` extension for graph data?

Any thoughts @robert-zaremba @clevinson @blushi ?

blushi (commented):
> For instance, what do we even think about regen: as an IRI scheme? Should it be something more generic like a hash: IRI scheme? What about the way it's structured using extensions and specifically the .rdf extension for graph data?

Using `regen` as IRI scheme basically means that we reference something that leaves on regen ledger and per https://tools.ietf.org/html/rfc3986#section-3.1 refers to a specification for assigning identifiers within that scheme. So I guess it makes sense to use `regen:` because that clearly specifies restrictions on identifiers. What would be the use case for having a more generic IRI scheme? Or did you mean that we could have multiple hash-based schemes?

x/data/server/query_server.go
```diff
@@ -10,55 +15,106 @@ import (
...
-	//	Signers:   signers,
-	//	Content:   content,
-	//}, err
+	iri, err := request.Hash.ToIRI()
+	if err != nil {
+		return nil, err
+	}
+
+	store := ctx.KVStore(s.storeKey)
+	id := s.iriIDTable.GetOrCreateID(store, []byte(iri))
```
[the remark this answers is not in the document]
aaronc: Oops! No we don't... we just want to get an existing ID or return `nil`... Good catch!

clevinson:
@aaronc I'm a fan of adding extensions to IRIs when they reference binary data, but now that i'm looking a bit more into examples of IRIs, i'm trying to think of how we actually expect this to be used. In looking at other [examples online](https://www.stardog.com/tutorials/data-model/#iris), I'm seeing a lot of use cases where the default namespace (e.g. `:Band` without a prefix, as it is referring to the local namespace) often consists of human readable text making the graph much easier to parse for end users.

Are there examples of how we expect to use human readable IRIs alongside hash based ones? I think it'd be nice to have some level of support for human readable IRIs in the default namespace. Perhaps we could provide support for this with some kind of name service for registering human readable IRIs? Alternatively (and probably the better option) would actually be to start with some base level schema that's defined in governance.

So if we want to support human readable IRIs, i'm wondering if we should differentiate btw prefixes of IRIs that correspond with `ContentHash`s and IRIs that are human-readable (potentially registered via a name service or managed by governance)?

Could we do something like `regen:hash:{CONTENT_HASH}.{EXT}` for all the ones we've previously discussed and maybe `regen:entity:{ENTITY_NAME}`? Is there best practicies for these? It seems like most examples I can find just use a raw "blank prefix" for this type of use case.

aaronc:
Fixed some of the most egregious errors here and removed `StoreRawData` (as per my concerns about GDPR and also unnecessarily bloating the ledger). There's still a lot of comments here for me to address, but at least I did find a bit of time to pick this up!

x/data/server/query_server.go
technicallyty: i agree here, i think it would be better to omit signers from the ByHash query and add a new query for a paginated list of signers on a certain content entry

x/data/server/testsuite/suite.go
aaronc: Let's handle this in a follow-up

x/data/server/msg_server.go
```diff
@@ -11,156 +17,139 @@ import (
...
+	if err != nil {
+		return nil, sdkerrors.Wrap(err, "invalid block time")
+	}
+
+	return timestamp, err
+}
+
+var trueBz = []byte{1}
 
 func (s serverImpl) SignData(ctx types.Context, request *data.MsgSignDataRequest) (*data.MsgSignDataResponse, error) {
```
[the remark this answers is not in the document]
aaronc: Let's maybe consider renaming to `Attest` in a follow-up

x/data/server/msg_server.go
```diff
@@ -2,164 +2,119 @@ package server
...
-//
-//	store := ctx.KVStore(s.storeKey)
-//	key := AnchorKey(cidBytes)
-//	store.Set(key, bz)
-//
-//	return ctx.EventManager().EmitTypedEvent(&data.EventAnchorData{Cid: cidBytes})
-//}
-
-//var emptyBz = []byte{0}
+type ToIRI interface {
```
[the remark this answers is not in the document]
aaronc: it's not really a hash... let's leave as is for now and if we come up with something better we rename

x/data/server/msg_server.go
```diff
@@ -11,156 +17,139 @@ import (
...
-	//err = ctx.EventManager().EmitTypedEvent(&data.EventSignData{
-	//	Cid:     cidBz,
-	//	Signers: request.Signers,
-	//})
-	//if err != nil {
-	//	return nil, err
-	//}
-	//
-	//return &data.MsgSignDataResponse{}, nil
+	iri, id, timestamp, err := s.getIRIAndAnchor(ctx, request.Hash)
```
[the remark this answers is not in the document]
aaronc: we did some renaming to make it clearer and it looks like tests confirm it is a no-op

x/data/server/query_server.go
```diff
@@ -2,63 +2,120 @@ package server
...
+
+	return &data.QueryByHashResponse{
+		Entry: entry,
+	}, nil
+}
+
+func (s serverImpl) getEntry(store sdk.KVStore, id []byte) (*data.ContentEntry, error) {
+	bz := store.Get(AnchorTimestampKey(id))
+	if len(bz) == 0 {
+		return nil, status.Error(codes.NotFound, "not found")
```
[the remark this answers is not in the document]
aaronc: ID is mainly for internal purposes right now (not sent by user) so probably best not to leak it

proto/regen/data/v1alpha2/query.proto
```diff
@@ -62,10 +65,24 @@ message ContentEntry {
...
+// QuerySignersResponse is the Query/QuerySigners response type.
+message QuerySignersResponse {
 
-  // content is the actual content if stored on-chain
-  Content content = 5;
+  // signers are the addresses of the signers.
+  repeated string signers = 1;
+
+  // pagination is the pagination PageResponse.
+  cosmos.base.query.v1beta1.PageResponse pagination = 3;
```
clevinson, line 87:
```suggestion
  cosmos.base.query.v1beta1.PageResponse pagination = 2;
```

x/data/iri.go
```diff
@@ -111,3 +106,133 @@ var mediaTypeExtensions = map[MediaType]string{
...
+}
+
+// ParseIRI parses an IRI string representation of a ContentHash into a ContentHash struct
+// Currently IRIs must have a "regen:" prefix, and only ContentHash_Graph and ContentHash_Raw
+// are supported.
+func ParseIRI(iri string) (*ContentHash, error) {
+	const regenPrefix = "regen:"
+
+	if !strings.HasPrefix(iri, regenPrefix) {
+		return nil, sdkerrors.ErrInvalidRequest.Wrap(fmt.Sprintf("can't parse IRI %s without %s prefix", iri, regenPrefix))
```
aaronc, line 125: @technicallyty we can simplify even further and just use wrapf. 

It would also be a good idea to add data module errors like InvalidIRI

ryanchristo (commented):
Looks good to me. A few questions/comments.

proto/regen/data/v1alpha2/query.proto
```diff
@@ -20,6 +20,9 @@ service Query {
   rpc BySigner (QueryBySignerRequest) returns (QueryBySignerResponse) {
     option (google.api.http).get = "/regen/data/v1alpha2/signers/{signer}";
   }
+
+  // Signers queries by IRI.
+  rpc Signers(QuerySignersRequest) returns (QuerySignersResponse);
```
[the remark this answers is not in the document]
technicallyty: yep, that makes the most sense to me, changed here https://github.com/regen-network/regen-ledger/pull/236/commits/47ef05a9dd5d3beb471e82a15c50ba4d7f552e19

ryanchristo (approved):
It might be nice to add a few more test cases but otherwise looks good to me.

proto/regen/data/v1alpha2/query.proto
```diff
@@ -12,8 +12,8 @@ option go_package = "github.com/regen-network/regen-ledger/x/data";
 // Query is the regen.data.v1alpha2 Query service
 service Query {
   // ByHash queries data based on its ContentHash.
-  rpc ByHash (QueryByHashRequest) returns (QueryByHashResponse) {
-    option (google.api.http).get = "/regen/data/v1alpha2/content/by_hash";
+  rpc ByIRI (QueryByIRIRequest) returns (QueryByIRIResponse) {
+    option (google.api.http).get = "/regen/data/v1alpha2/content/{iri}";
```
technicallyty, line 16: after a call with @clevinson we thought querying by iri makes more sense here. 

thoughts on this design? @aaronc @ryanchristo

ryanchristo: Makes sense to me. More aligned with the signers query.

aaronc: Sure but ideally we have both. Or maybe we add query methods to convert between IRIs and hashes but always query with IRIs by default?

aaronc:
@technicallyty @clevinson @ryanchristo what do we see remaining here to merge this? Are there some things that can be dealt with in follow-ups? Maybe we can break those out into issues if so...

ryanchristo:
> Are there some things that can be dealt with in follow-ups? Maybe we can break those out into issues if so...

I think we should merge and further discuss and address remaining items in follow-ups. Here are some of the items that I see as unresolved:

- [ ] add support for human readable IRIs
- [ ] use attestation terminology (e.g. rename `SignData` to `Attest`)
- [ ] add query methods to convert between IRIs and hashes but always query with IRIs by default
- [ ] improve test coverage (i.e. add more test cases for parsing IRI)

aaronc:
Thanks @ryanchristo. What do you mean by " add support for human readable IRIs"?

ryanchristo:
> What do you mean by "add support for human readable IRIs"?

This was something @clevinson brought up in https://github.com/regen-network/regen-ledger/pull/236#issuecomment-776411378. If we want to continue exploring this idea, we should open an issue or discussion to do so but I don't see it as a priority.

Also, I would be happy to open issues for the items listed following a second approval.

aaronc:
Sure please feel free to open issues 👍

clevinson (approved):
utACK.

Thanks @technicallyty for wrapping this up!

PR #238, opened 2021-02-04
Add logs to root command

akhilkumarpilli:


app/regen/cmd/root.go
```diff
@@ -79,6 +81,9 @@ func Execute(rootCmd *cobra.Command) error {
 	ctx = context.WithValue(ctx, client.ClientContextKey, &client.Context{})
 	ctx = context.WithValue(ctx, server.ServerContextKey, server.NewDefaultContext())
 
+	rootCmd.PersistentFlags().String(flags.FlagLogLevel, zerolog.InfoLevel.String(), "The logging level (trace|debug|info|warn|error|fatal|panic)")
+	rootCmd.PersistentFlags().String(flags.FlagLogFormat, tmcfg.LogFormatPlain, "The logging format (json|plain)")
```
amaurymartiny, line 85: The reason why we cannot just use `github.com/cosmos/cosmos-sdk/server/cmd` is because of the REGEN env prefix, is that right?

amaurymartiny (approved):
Code lgtm, it looks like what we have on the SDK (and simapp has correct logs). However I did NOT test it manually.

app/regen/cmd/root.go
akhilkumarpilli: Yes, we cannot use REGEN env prefix.

anilCSE (approved):

PR #245, opened 2021-02-04
Group account creation

aaronc:
Adds a hack to allow us to use auth.AccountKeeper in the group module with ADR 033 and uses this to create group module accounts.


aaronc:
There are some failing tests here that I'm not too sure about. Can you check @blushi ? Looks like otherwise things are mostly wired up.

app/app.go
```diff
@@ -359,10 +359,19 @@ func NewRegenApp(logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest
...
 
-	err := newModuleManager.RegisterModules(NewModules)
+	// BEGIN HACK: this is a total, ugly hack until x/auth supports ADR 033 or we have a suitable alternative
+	groupModule := group.Module{AccountKeeper: authkeeper.AccountKeeper{}}
+	// use a separate newModules from the global NewModules here because we need to pass state into the group module
+	newModules := []moduletypes.Module{
+		ecocredit.Module{},
+		data.Module{},
+		groupModule,
+	}
```
robert-zaremba, line 369: In line 137 we have:
```
	NewModules = []moduletypes.Module{
		ecocredit.Module{},
		data.Module{},
		group.Module{},
	}
```.
It seams we should remove that.

File: app/app.go [hunk 2]
```diff
@@ -359,10 +359,19 @@ func NewRegenApp(logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest
 	/* New Module Wiring START */
 	newModuleManager := servermodule.NewManager(app.BaseApp, codec.NewProtoCodec(interfaceRegistry))
 
-	err := newModuleManager.RegisterModules(NewModules)
+	// BEGIN HACK: this is a total, ugly hack until x/auth supports ADR 033 or we have a suitable alternative
```
robert-zaremba, line 362: Is this a hack or a TODO? 

Why we can't use a `MsgServer` client here?

types/module/server/module_key.go
```diff
@@ -1,14 +1,15 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/regen-network/regen-ledger/types"
 )
 
 type ModuleKey interface {
 	types.InvokerConn
 
 	ModuleID() types.ModuleID
-	Address() []byte
+	Address() sdk.AccAddress
```
robert-zaremba, line 12: :+1:

x/group/conditions.go
```diff
@@ -1,130 +0,0 @@
-package group
```
robert-zaremba, line 1: why removing this file?

aaronc: It's replaced by ADR 028/033

app/app.go
aaronc: It's a brutal hack

app/app.go
aaronc: Because cli code still uses it. This is strictly a hack until x/auth supports ADR 033

app/app.go
aaronc: x/auth has no MsgServer

aaronc:
Found a slightly cleaner way I think, but still a bunch of test failures...

blushi:
> Found a slightly cleaner way I think, but still a bunch of test failures...

@aaronc I've fixed the tests as part of https://github.com/regen-network/regen-ledger/pull/252

aaronc:
Thanks @blushi !

aaronc:
What do you think otherwise @blushi ?

blushi (approved):
Pre-approving, hacky but does the job for now.

app/app.go
aaronc: Because they are used for setting up cli commands

aaronc:
Super hacky. Please let's try to think of better solutions 😉

PR #246, opened 2021-02-04
Network based test fixtures

aaronc:


clevinson:
@aaronc are you planning to pick this work up again at some point, or should we close this PR?

aaronc:
It's really something somebody should do in some way, but in the SDK when we do app wiring

Issue #240, opened 2021-02-04
[x/group] Add simulations


blushi:

clevinson:
@aleem1314 working on this, and @blushi to meet & address some blockers.

PR #251, opened 2021-02-05
feat: add x/group simulations

aleem1314:
closes: #240


x/group/module/module.go
```diff
@@ -63,6 +72,29 @@ func (a Module) GetQueryCmd() *cobra.Command {
...
+
+func (a Module) RegisterStoreDecoder(registry sdk.StoreDecoderRegistry) {
+}
+
+func (a Module) WeightedOperations(simState module.SimulationState) []simtypes.WeightedOperation {
+	protoCdc := codec.NewProtoCodec(a.Registry)
+	// TODO: WeightedOperations requires group module query-server
+	return simulation.WeightedOperations(
+		simState.AppParams, simState.Cdc,
+		a.AccountKeeper, a.BankKeeper, protoCdc,
```
aleem1314, line 94: Is there any way to access group module msgServer here @aaronc?

aaronc: We can use the MsgClient with text fixtures. Do you need private state here?

aleem1314: Each simulation operation can depend on the previous one. In the case of the group module, to create a group account 
I don't think . We need a way to pass `group.QueryClient` [here](https://github.com/regen-network/regen-ledger/pull/251/files#diff-e972a3eebdba55ab7bc12235f56652e406f5bc2229c8b4bbe573a2479e9416cdR92).

aleem1314: The whole simulation state is initialized in `app/sim_test.go`. Mostly we are querying the group module state, we need a way to pass `group.QueryClient` from [here](https://github.com/regen-network/regen-ledger/pull/251/files#diff-e972a3eebdba55ab7bc12235f56652e406f5bc2229c8b4bbe573a2479e9416cdR92).
Some context:  Each simulation operation can depend on the previous one. In the case of the group module, to create a group account we need [group-info](https://github.com/regen-network/regen-ledger/pull/251/files#diff-c7b2e4ef3c1b56b3e1be6efbd7d9ed108b8f3457089b7344692be5fe82733608R145) (which is created in the previous operation [SimulateMsgCreateGroup](https://github.com/regen-network/regen-ledger/pull/251/files#diff-c7b2e4ef3c1b56b3e1be6efbd7d9ed108b8f3457089b7344692be5fe82733608R79)).

x/group/module/module.go
```diff
@@ -64,6 +73,27 @@ func (a Module) GetQueryCmd() *cobra.Command {
...
+
+func (a Module) RandomizedParams(r *rand.Rand) []simtypes.ParamChange {
+	return nil
+}
+
+func (a Module) RegisterStoreDecoder(registry sdk.StoreDecoderRegistry) {
+}
+
+func (a Module) WeightedOperations(simState module.SimulationState) []simtypes.WeightedOperation {
+	return simulation.WeightedOperations(
```
aleem1314, line 91:
```suggestion
        // TODO: simulation.WeightedOperations requires group module's `QueryServer` here.
	return simulation.WeightedOperations(
```

app/test_helpers.go
```diff
@@ -0,0 +1,65 @@
...
+	abci "github.com/tendermint/tendermint/abci/types"
+	"github.com/tendermint/tendermint/libs/log"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	tmtypes "github.com/tendermint/tendermint/types"
+	dbm "github.com/tendermint/tm-db"
+)
+
+// DefaultConsensusParams defines the default Tendermint consensus params used in
+// SimApp testing.
+var DefaultConsensusParams = &abci.ConsensusParams{
```
blushi, line 17: can't we export this variable directly from the sdk?

app/utils.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	simState.ParamChanges = app.SimulationManager().GenerateParamChanges(config.Seed)
+	simState.Contents = app.SimulationManager().GetProposalContents(simState)
+	return app.nm.WeightedOperations(simState)
+}
+
+// CheckExportSimulation exports the app state and simulation parameters to JSON
+// if the export paths are defined.
+func CheckExportSimulation(
```
blushi, line 40: we could just re-use the `CheckExportSimulation` from the sdk which uses `App` interface, doesn't seem like `RegenApp` is needed here.

proto/regen/group/v1alpha1/query.proto
```diff
@@ -45,6 +48,17 @@ service Query {
...
+message QueryGroupsRequest {
+}
+
+// QueryGroupsResponse is the Query/Groups response type.
+message QueryGroupsResponse {
+
+    // info is the list of GroupInfo for the group.
+    repeated GroupInfo info = 1;
+}
+
```
blushi, line 61: we should really add pagination request and response fields.

types/module/simulation/simulation.go
```diff
@@ -0,0 +1,97 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	"github.com/cosmos/cosmos-sdk/types/simulation"
+	"github.com/regen-network/regen-ledger/types/module/server"
+)
+
+// AppModuleSimulation defines the standard functions that every module should expose
+// for the SDK blockchain simulator
+type AppModuleSimulation interface {
```
blushi, line 14: no need to redefine this interface which is already part of the sdk as is.

types/module/simulation/simulation.go
```diff
@@ -0,0 +1,97 @@
...
+	for _, module := range sm.Modules {
+		paramChanges = append(paramChanges, module.RandomizedParams(r)...)
+	}
+
+	return
+}
+
+// WeightedOperations returns all the modules' weighted operations of an application
+func (sm *AppSimulationManager) WeightedOperations(simState module.SimulationState) []simulation.WeightedOperation {
+	// TODO: change it to use New module manager
```
blushi, line 89: I guess this can be removed?
```suggestion
```

types/module/simulation/simulation.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// NewAppSimulationManager creates a new SimulationManager object
+//
+// CONTRACT: All the modules provided must be also registered on the module Manager
+func NewAppSimulationManager(serverManager *server.Manager, modules ...module.AppModuleSimulation) *AppSimulationManager {
+	return &AppSimulationManager{
+		Modules:       modules,
+		StoreDecoders: make(sdk.StoreDecoderRegistry),
+		manager:       serverManager,
```
blushi, line 46: could we just pass in the result of `serverManager.GetWeightedOperationsHandlers()` instead of the whole `serverManager`?

types/module/simulation/simulation.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return
+}
+
+// WeightedOperations returns all the modules' weighted operations of an application
+func (sm *AppSimulationManager) WeightedOperations(simState module.SimulationState) []simulation.WeightedOperation {
+	// TODO: change it to use New module manager
+	wOps := make([]simulation.WeightedOperation, 0, len(sm.Modules))
+	modules := sm.manager.GetWeightedOperationsHandlers()
```
blushi, line 91: In `sm.Modules`, we have the regular modules (from cosmos-sdk) while in `modules`, we'll have the `WeightedOperations` methods from the new regen modules (ecocredit, data, group) so this seems wrong to me.

x/group/server/query_server.go
```diff
@@ -9,6 +9,27 @@ import (
...
+	var infos []*group.GroupInfo
+	var info group.GroupInfo
+	defer it.Close()
+	for {
+		_, err := it.LoadNext(&info)
+		if err != nil {
+			break
+		}
+		infos = append(infos, &info)
+	}
```
blushi, line 26: ditto, use pagination request and `Paginate` function from `orm`.

proto/regen/group/v1alpha1/query.proto
```diff
@@ -45,6 +48,17 @@ service Query {
...
 
+// QueryGroupsRequest is the Query/Groups request type.
+message QueryGroupsRequest {
+}
+
+// QueryGroupsResponse is the Query/Groups response type.
+message QueryGroupsResponse {
+
+    // info is the list of GroupInfo for the group.
+    repeated GroupInfo info = 1;
```
blushi, line 59: I'd rather call it this way:
```suggestion
    repeated GroupInfo groups = 1;
```

x/group/simulation/operations.go
```diff
@@ -0,0 +1,748 @@
...
+		_, _, err = app.Deliver(txGen.TxEncoder(), tx)
+		if err != nil {
+			return simtypes.NoOpMsg(group.ModuleName, msg.Type(), "unable to deliver tx"), nil, err
+		}
+
+		return simtypes.NewOperationMsg(&msg, true, ""), nil, err
+	}
+}
+
+// TODO: Add WeightedOperations for MsgVote & MsgExec. Waiting on genesis PR
```
blushi, line 748: why are you waiting on genesis PR? If it's just about adding `proposal_id` to a `Proposal`, you can go ahead directly here.

blushi (changes requested):
Thanks @aleem1314 for pulling this up!
Generally, I believe we still need to remove some amount of code directly copied from cosmos-sdk.
Also, this doesn't feel right to me to make the bank keeper a dependency of the group server just for simulations, do you see any other possible approach?

aleem1314:
> Also, this doesn't feel right to me to make the bank keeper a dependency of the group server just for simulations, do you see any other possible approach?

I think we can use `bank/QueryClient`, similar to `group/QueryClient`. the issue here is bank module's `QueryClient` doesn't have all the queries, which are required in simulations.

x/group/server/query_server.go
```diff
@@ -9,6 +9,24 @@ import (
 	"github.com/regen-network/regen-ledger/x/group"
 )
 
+func (s serverImpl) Groups(ctx types.Context, request *group.QueryGroupsRequest) (*group.QueryGroupsResponse, error) {
+	it, err := s.groupTable.PrefixScan(ctx, nil, nil)
```
blushi, line 13: You have to use `request.Pagination.Key` in `PrefixScan` `start` parameter. Check out `Paginate` function documentation for more details on how to instantiate iterators to pass into `Paginate`.

types/module/simulation/simulation.go
```diff
@@ -0,0 +1,67 @@
...
+	"math/rand"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	"github.com/cosmos/cosmos-sdk/types/simulation"
+)
+
+// AppSimulationManager defines a simulation manager that provides the high level utility
+// for managing and executing simulation functionalities for a group of modules
+type AppSimulationManager struct {
```
blushi, line 13: why do we need a different simulation manager than the one already defined in cosmos-sdk?

x/group/server/msg_server.go
```diff
@@ -425,6 +425,7 @@ func (s serverImpl) CreateProposal(ctx types.Context, req *group.MsgCreatePropos
 		return nil, sdkerrors.Wrap(err, "end time conversion")
 	}
 
+	proposalID := s.proposalTable.Sequence().NextVal(ctx)
```
blushi, line 428: you can just use `s.proposalTable.Sequence().PeekNextVal(ctx)` instead and keep using `s.proposalTable.Create(ctx, m)` which will increment and persist the counter.

proto/regen/group/v1alpha1/query.proto
```diff
@@ -45,6 +48,23 @@ service Query {
   rpc VotesByVoter(QueryVotesByVoterRequest) returns (QueryVotesByVoterResponse);
 }
 
+// QueryGroupsRequest is the Query/Groups request type.
+message QueryGroupsRequest {
+
+  // pagination defines an optional pagination for the request.
+  cosmos.base.query.v1beta1.PageRequest pagination = 2;
```
blushi, line 55:
```suggestion
  cosmos.base.query.v1beta1.PageRequest pagination = 1;
```

blushi (changes requested):

x/group/simulation/operations.go
```diff
@@ -0,0 +1,941 @@
...
+
+		_, _, err = app.Deliver(txGen.TxEncoder(), tx)
+		if err != nil {
+			return simtypes.NoOpMsg(group.ModuleName, msg.Type(), "unable to deliver tx"), nil, err
+		}
+		return simtypes.NewOperationMsg(msg, true, ""), nil, err
+	}
+}
+
+func getGroupDetails(sdkCtx sdk.Context, qryClient group.QueryClient, acc simtypes.Account) (groupAdmin string, groupID uint64, op simtypes.OperationMsg, err error) {
```
blushi, line 276: you could have just used `GroupsByAdmin` query here to query groups by admin account.

blushi (changes requested):

types/module/server/manager.go
```diff
@@ -115,6 +128,22 @@ func (mm *Manager) RegisterModules(modules []module.Module) error {
 	return nil
 }
 
+// WeightedOperations returns all the modules' weighted operations of an application
+func (mm *Manager) WeightedOperations(state sdkmodule.SimulationState, modules []sdkmodule.AppModuleSimulation) []simulation.WeightedOperation {
+	wOps := make([]simulation.WeightedOperation, 0, len(modules))
```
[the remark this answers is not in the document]
aleem1314: Yeah, I'm getting a lint issue.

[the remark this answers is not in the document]
blushi: well in this case, let's keep using `make` but with a capacity that makes sense ie `len(modules) + len(mm.weightedOperationsHandlers)` and that will avoid any reallocation.

blushi (commented):
@aleem1314 Do we still need `Groups` query?

 Not required. Shall I remove it?

blushi:
> @aleem1314 Do we still need `Groups` query?

If it's not required, let's just remove it.

types/module/server/manager.go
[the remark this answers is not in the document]
blushi: unresolving this as this hasn't been addressed

x/group/simulation/genesis.go
```diff
@@ -0,0 +1,3 @@
+package simulation
+
+// TODO: #270
```
blushi, line 3: If genesis PR #270 is merged before this one, then let's do it as part of this PR, else we'll have to create a follow-up issue for adding genesis simulation.

blushi (commented):
Looks good overall but I'd like this to be addressed before approving: https://github.com/regen-network/regen-ledger/pull/251#discussion_r593036370

types/module/server/manager.go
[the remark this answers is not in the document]
aleem1314: We cannot use `len(modules) + len(mm.weightedOperationsHandlers)` . Each module returns [`[]simulation.WeightedOperation` ](https://github.com/regen-network/regen-ledger/pull/251/files#diff-a3144c816ac58cb8fbf744d53aa6cc93d3601da586fb23a432db1909b3c4eff8R134-R142).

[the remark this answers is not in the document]
blushi: oh I see, they might actually be more elements in `wOps` than `len(modules) + len(mm.weightedOperationsHandlers)` but still, pre-allocating would avoid reallocation until we reach `len(modules) + len(mm.weightedOperationsHandlers)` (whereas using `var wOps []simulation.WeightedOperation`, it will realloc from the very first `append`)

app/experimental_appconfig.go
```diff
@@ -79,7 +79,7 @@ func setCustomModules(app *RegenApp, interfaceRegistry types.InterfaceRegistry)
 	newModuleManager := server.NewManager(app.BaseApp, codec.NewProtoCodec(interfaceRegistry))
 
 	// BEGIN HACK: this is a total, ugly hack until x/auth supports ADR 033 or we have a suitable alternative
```
blushi, line 81: maybe update the comment to explain why we're adding bank keeper too now...

x/group/simulation/genesis.go
blushi: now that #270 is merged, could you address this TODO?

app/app.go
```diff
@@ -538,6 +542,11 @@ func (app *RegenApp) SimulationManager() *module.SimulationManager {
 	return app.sm
```
blushi, line 542: do you still need this sim manager `sm *module.SimulationManager`?

blushi (changes requested):

app/app.go
aleem1314: Yes, to initialize SDK modules simulations.

app/app.go
```diff
@@ -538,6 +541,11 @@ func (app *RegenApp) SimulationManager() *module.SimulationManager {
...
 }
 
+// NewManager implements the SimulationApp interface
+func (app *RegenApp) NewManager() *server.Manager {
```
blushi, line 545: why is this needed?

app/app.go
blushi: So we're using the regular sdk `SimulationManager` for all except `WeightedOperation` which needs the new server module manager `smm`, right? Probably better to clarify that in some comments. We will likely need to revisit this as part of #301

blushi (changes requested):

app/app.go
aleem1314: Removing it. Not using anywhere.

x/group/simulation/genesis.go
```diff
@@ -0,0 +1,186 @@
...
+	GroupProposals   = "group-proposals"
+	GroupVote        = "group-vote"
+)
+
+func getGroups(r *rand.Rand, accounts []simtypes.Account) []*group.GroupInfo {
+	groups := make([]*group.GroupInfo, 3)
+	for i := 0; i < 3; i++ {
+		acc, _ := simtypes.RandomAcc(r, accounts)
+		groups[i] = &group.GroupInfo{
+			GroupId:     1,
```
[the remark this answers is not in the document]
aleem1314: Yeah, make sense.

.github/workflows/sims.yml
```diff
@@ -0,0 +1,155 @@
...
+            go.mod
+            go.sum
+      - uses: actions/cache@v2.1.3
+        with:
+          path: ~/go/bin
+          key: ${{ runner.os }}-go-runsim-binary
+        if: env.GIT_DIFF
+      - name: test-test-nondeterminism
+        run: |
+          make sim-regen-nondeterminism
```
anilCSE, line 70: Should we have two jobs. One for `stable` and one for `experimental` builds?

sims.mk
```diff
@@ -6,40 +6,25 @@
 sim-regen-nondeterminism:
 	@echo "Running nondeterminism test..."
 	@go test -mod=readonly $(APP_DIR) -run TestAppStateDeterminism -Enabled=true \
-		-NumBlocks=100 -BlockSize=200 -Commit=true -Period=0 -v -timeout 24h
+		-NumBlocks=50 -BlockSize=100 -Commit=true -Period=0 -v -timeout 24h -tags=experimental
```
anilCSE, line 9: May be we should have 2 simulations. One for stable as well.

anilCSE: Also is there any reason for changing the NumBlocks and BlockSize?

x/group/simulation/genesis.go
```diff
@@ -0,0 +1,189 @@
...
+			Metadata:       []byte(simtypes.RandStringOfLength(r, 10)),
+		}
+	}
+	return groupMembers
+}
+
+func getProposals(r *rand.Rand, simState *module.SimulationState) []*group.Proposal {
+	proposals := make([]*group.Proposal, 3)
+	for i := 0; i < 3; i++ {
+
```
anilCSE, line 79:
```suggestion

	for i := 0; i < 3; i++ {
```

x/group/simulation/genesis.go
```diff
@@ -0,0 +1,189 @@
...
+		from, _ := simtypes.RandomAcc(r, simState.Accounts)
+		to, _ := simtypes.RandomAcc(r, simState.Accounts)
+
+		proposal := &group.Proposal{
+			ProposalId:          uint64(i + 1),
+			Proposers:           []string{simState.Accounts[0].Address.String(), simState.Accounts[1].Address.String()},
```
anilCSE, line 85: move this `simState.Accounts[0].Address.String()` and `simState.Accounts[1].Address.String()` out of loop.

x/group/simulation/genesis.go
```diff
@@ -0,0 +1,189 @@
...
+				AbstainCount: "0",
+				VetoCount:    "0",
+			},
+			ExecutorResult: group.ProposalExecutorResultNotRun,
+			Metadata:       []byte(simtypes.RandStringOfLength(r, 50)),
+			SubmittedAt:    gogotypes.Timestamp{Seconds: 1},
+			Timeout:        gogotypes.Timestamp{Seconds: 1000},
+		}
+		err := proposal.SetMsgs([]sdk.Msg{&banktypes.MsgSend{
+			FromAddress: from.Address.String(),
```
anilCSE, line 103: `from.Address.String()` is used multiple times. Use a variable.

x/group/simulation/genesis.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func getVotes(r *rand.Rand, simState *module.SimulationState) []*group.Vote {
+	votes := make([]*group.Vote, 3)
+
+	for i := 0; i < 3; i++ {
+		votes[i] = &group.Vote{
+			ProposalId:  uint64(i + 1),
+			Voter:       simState.Accounts[i].Address.String(),
+			Choice:      group.Choice_CHOICE_YES,
```
anilCSE, line 123: Can we randomize the vote choice or do we need `YES` always?

x/group/simulation/genesis.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	bz, err := simState.Cdc.MarshalJSON(&groupGenesis)
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Printf("Selected randomly generated %s parameters:\n%s\n", group.ModuleName, bz)
+	simState.GenState[group.ModuleName] = simState.Cdc.MustMarshalJSON(&groupGenesis)
+
```
anilCSE, line 188:
```suggestion
```

x/group/simulation/genesis.go
```diff
@@ -0,0 +1,189 @@
...
+		Proposals:       proposals,
+		Votes:           votes,
+	}
+
+	bz, err := simState.Cdc.MarshalJSON(&groupGenesis)
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Printf("Selected randomly generated %s parameters:\n%s\n", group.ModuleName, bz)
```
anilCSE, line 186:
```suggestion
	fmt.Printf("selected randomly generated %s parameters:\n%s\n", group.ModuleName, bz)
```

anilCSE: This is genesis state and not just parameters right?
```suggestion
	fmt.Printf("selected randomly generated %s genesis state:\n%s\n", group.ModuleName, bz)
```

x/group/simulation/operations.go
```diff
@@ -0,0 +1,935 @@
...
+	OpMsgCreateProposal                   = "op_weight_msg_create_proposal"
+	OpMsgVote                             = "op_weight_msg_vote"
+	OpMsgExec                             = "ops_weight_msg_exec"
+)
+
+const (
+	WeightCreateGroup    = 150
+	WeightCreateProposal = 100
+	WeightUpdateGroup    = 5
+	WeightGroupMember    = 40
```
anilCSE, line 42: Any rationale for these weights?

x/group/simulation/operations.go
```diff
@@ -0,0 +1,935 @@
...
+}
+
+// SimulateMsgCreateGroup generates a MsgCreateGroupRequest with random values
+func SimulateMsgCreateGroup(ak exported.AccountKeeper, bk exported.BankKeeper) simtypes.Operation {
+	return func(
+		r *rand.Rand, app *baseapp.BaseApp, ctx sdk.Context, accounts []simtypes.Account, chainID string) (simtypes.OperationMsg, []simtypes.FutureOperation, error) {
+		acc := accounts[0]
+		account := ak.GetAccount(ctx, acc.Address)
+
+		spendableCoins := bk.SpendableCoins(ctx, account.GetAddress())
```
anilCSE, line 174:
```suggestion
		acc := accounts[0]
		accAddr := acc.Address
		account := ak.GetAccount(ctx, accAddr)

		spendableCoins := bk.SpendableCoins(ctx, accAddr)
```

x/group/simulation/operations.go
```diff
@@ -0,0 +1,935 @@
...
+		fees, err := simtypes.RandomFees(r, ctx, spendableCoins)
+		if err != nil {
+			return simtypes.NoOpMsg(group.ModuleName, group.TypeMsgCreateGroup, "fee error"), nil, err
+		}
+
+		members := []group.Member{
+			{
+				Address:  acc.Address.String(),
```
anilCSE, line 182:
```suggestion
				Address:  accAddr.String(),
```

or even
```go
accAddrStr := accAddr.String()
```
```suggestion
				Address:  accAddrStr,
```

anilCSE (commented):
overall lgtm. A couple of nits and questions

sims.mk
aleem1314: Updated to original values. I was testing simulations. `-NumBlocks=100` `-BlockSize=200` taking so much time to finish simulation jobs.

x/group/simulation/operations.go
aleem1314: Yes, If the group or group account modified `MsgVote` & `MsgExec` operations return `noOp`, that's why we have less weight for group & group account update.

blushi: could add a comment in the code to explain this then?

blushi (approved):
LGTM

app/experimental_appconfig.go
```diff
@@ -99,6 +99,7 @@ func setCustomModules(app *RegenApp, interfaceRegistry types.InterfaceRegistry)
 
 	return newModuleManager
 	/* New Module Wiring END */
+	return newModuleManager
```
atheeshp, line 102:
```
```

atheeshp, line 102:
```suggestion
```

x/group/simulation/operations.go
```diff
@@ -0,0 +1,938 @@
...
+		if err != nil {
+			return simtypes.NoOpMsg(group.ModuleName, group.TypeMsgVote, "fail to query group info"), nil, err
+		}
+
+		groupAccounts := result.GetGroupAccounts()
+		if len(groupAccounts) == 0 {
+			return simtypes.NoOpMsg(group.ModuleName, group.TypeMsgVote, "no group account found"), nil, nil
+		}
+
+		proposalsResult, err := queryClient.ProposalsByGroupAccount(ctx, &group.QueryProposalsByGroupAccountRequest{Address: groupAccounts[0].Address})
```
atheeshp, line 886: Can we add helper for `groupAccountAddress`, Same code present in other functions.

atheeshp (approved):
lgtm overall, nice one @aleem1314, thanks!

blushi:
@anilCSE could you have a final look at this when you have time so we can eventually merge it? Thanks.

x/group/module/module.go
```diff
@@ -71,5 +81,34 @@ func (a Module) RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
 }
 
 func (a Module) Route(configurator servermodule.Configurator) sdk.Route {
-	return sdk.NewRoute(group.RouterKey, server.NewHandler(configurator, a.AccountKeeper))
+	return sdk.NewRoute(group.RouterKey, server.NewHandler(configurator, a.AccountKeeper, a.BankKeeper))
+}
+
+// AppModuleSimulation functions
+
+// GenerateGenesisState creates a randomized GenState of the group module.
```
anilCSE, line 89:
```suggestion
// GenerateGenesisState creates a randomized GenesisState of the group module.
```

x/group/simulation/operations.go
```diff
@@ -0,0 +1,911 @@
...
+	OpMsgExec                             = "ops_weight_msg_exec"
+)
+
+//  If update group or group account txn's executed, `SimulateMsgVote` & `SimulateMsgExec` txn's returns `noOp`.
+//  That's why we have less weight for update group & group-account txn's.
+const (
+	WeightCreateGroup    = 150
+	WeightCreateProposal = 100
+	WeightUpdateGroup    = 5
+	WeightGroupMember    = 40
```
anilCSE, line 44: We should declare weights for all the sdk messages. The SDK style of doing this is `DefaultWeightMsgMultiSend`, `DefaultWeightMsgDelegate` etc. Also, don't see a reason for keeping `150` for `WeightCreateGroup`. May be it can be just `100`. All sdk weights have either 100 or lesser. May be we should follow the same

anilCSE (commented):
implementation lgtm. left a few questions

blushi:
@aleem1314 could you resolve the conflicts so we can merge when you have time?

blushi:
@anilCSE are we good to merge this?

aleem1314:
> @anilCSE are we good to merge this?

Don't merge. One simulation job is failing.

blushi:
> > @anilCSE are we good to merge this?
> 
> Don't merge. One simulation job is failing.

It seems like it's due to group invariants being broken, could you have a look when you have time @aleem1314?

aleem1314:
@anilCSE can you take a final look?

anilCSE (approved):

blushi:
@aleem1314 are we good to merge this?

PR #256, opened 2021-02-10
make configurable builds

anilCSE:
This PR enables a configuration to build `stable` and `experimental` builds via ENV flags

To generate stable builds:
`make build`

To generate experimental builds:
`EXPERIMENTAL=true make build`


app/stable_appconfig.go
```diff
@@ -0,0 +1,34 @@
...
+
+package app
+
+import (
+	"github.com/cosmos/cosmos-sdk/types/module"
+)
+
+func setCustomModuleBasics() []module.AppModuleBasic {
+	return []module.AppModuleBasic{}
+}
```
anilCSE, line 11: basically in `stable` version, we'll not have any experimental modules integrated.

app/experimental_appconfig.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/regen-network/regen-ledger/x/ecocredit"
+	"github.com/regen-network/regen-ledger/x/group"
+)
+
+func setCustomModuleBasics() []module.AppModuleBasic {
+	return []module.AppModuleBasic{
+		data.AppModuleBasic{},
+		ecocredit.AppModuleBasic{},
+		group.AppModuleBasic{},
+	}
```
anilCSE, line 18: this is an experimental build, so we integrate new modules here

blushi: you will probably need another function to account for the new module wiring and `setCustomKeepers` might be left empty for regen modules since they don't actually have a keeper.

Makefile
```diff
@@ -17,6 +17,11 @@ export GO111MODULE = on
 # process build tags
 
 build_tags = netgo
+
+ifeq ($(STABLE),true)
```
amaurymartiny, line 21: I rarely see the `STABLE` flag out there, is it something known?

An alternative would be just to have 2 scripts:
- `make build` -> stable
- `make build-experimental` -> experimental

Don't want to block the PR on this, just a suggestion.

amaurymartiny (approved):

Makefile
anilCSE: Not really, I didn't see anyone using STABLE flag either.  My intention was not to introduce any extra commands to handle this. That's the reason to use an ENV setting. But later looks self-explanatory.
Also might need to have 2 install commands too if we follow separate commands to generate stable and experimental builds

clevinson: I'm fine with env variable, but my preference would be for having the name of the flag `EXPERIMENTAL`, `ENABLE_EXPERIMENTAL_FEATURES`, or `EXPERIMENTAL_BUILD`, etc. (with defualt to false), rather than `STABLE` with default to true. I like this better because "stable" usually implies things like API guarantees, and until we reach regen ledger v1.0, or v1 proto definitions, we don't really have such guarantees around stability of our API yet. "Experimental" speaks more about the featureset, which can include which modules are part of a build.

Not to bike shed, but thats my 2c.

aaronc: Agree about experimental vs stable. Otherwise looks good!

anilCSE: Agree, will change it to `EXPERIMENTAL`

Makefile
```diff
@@ -17,6 +17,11 @@ export GO111MODULE = on
 # process build tags
 
 build_tags = netgo
+
+ifeq ($(EXPERIMENTAL),true)
+	build_tags += stable
```
amaurymartiny, line 22: @anilCSE The logic is reversed here

maybe the build_tag should also be called `experimental`?

amaurymartiny (approved):

app/app.go
```diff
@@ -356,28 +342,8 @@ func NewRegenApp(logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest
...
-	}
-	// END HACK
-
-	err = newModuleManager.CompleteInitialization()
-	if err != nil {
-		panic(err)
-	}
-	/* New Module Wiring END */
+	// register experimental modules here
+	setCustomModules(app, interfaceRegistry)
```
robert-zaremba, line 346: :+1:

PR #260, opened 2021-02-16
Update getting-started.md

dpdanpittman:
Adding more detailed instructions on how to build and setup Regen ledger

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -4,26 +4,192 @@
 
 ### Building from source
 
-Prerequisites: The [Go](https://golang.org/doc/install) compiler **version 1.15** (we use
-go modules) or later and GNU make. The go bin path (usually `$HOME/go/bin`) should be
-in your system `$PATH`.
+#### Prerequisites
 
-Run `make install` to build the code. The `regen` binary serves as both the blockchain daemon, as well as
+**Minimum Hardware Requirements**
+
+- 8GB RAM
+- 4vCPUs
+- 200GB Disk space
+
+**Software Requirements** 
+
+Before beginning make sure you have the [Go](https://golang.org/doc/install) compiler **version 1.15** (we use
+go modules) or later and GNU make installed and add the go bin path (usually `$HOME/go/bin`) should be in your system `$PATH`. To learn more about adding commands to your path please visit [linuxize.com](https://linuxize.com/post/how-to-add-directory-to-path-in-linux/). We recommend using Ubuntu 18.04 or higher, the official images can be found at [Ubuntu.com](https://ubuntu.com/tutorials/install-ubuntu-desktop#1-overview)
+
+---
+
+#### Download and Install Regen Ledger
+
+Install git and build-essentials for ubuntu
+```sh
+sudo apt update
+```
+```sh
+sudo apt install git build-essential jq -y
+```
+
+Download go and extract tarball
+```sh
+wget https://dl.google.com/go/go1.15.2.linux-amd64.tar.gz
+```
+```sh
+tar -xvf go1.15.2.linux-amd64.tar.gz
+```
+```sh
+sudo mv go /usr/local
+```
+
+Add go to local path
+```sh
+echo "" >> ~/.profile
+```
+```sh
+echo 'export GOPATH=$HOME/go' >> ~/.profile
+```
+```sh
+echo 'export GOROOT=/usr/local/go' >> ~/.profile
+```
+```sh
+echo 'export GOBIN=$GOPATH/bin' >> ~/.profile
+```
+```sh
+echo 'export PATH=$PATH:/usr/local/go/bin:$GOBIN' >> ~/.profile
+```
+
+```sh
+. ~/.profile
+```
+
+Check to make sure you have the correct version of installed **version 1.15** or higher
+```sh
+go version
+```
+
+Next, we download the official Regen ledger github repository
+```sh
+cd $HOME
+```
+```sh
+git clone https://github.com/regen-network/regen-ledger $GOPATH/src/github.com/regen-network/regen-ledger
+```
+```sh
+cd $GOPATH/src/github.com/regen-network/regen-ledger
+```
+
+Run `make install` to build the code.
+```sh
+make install
+```
+
+Once installed the `regen` binary serves as both the blockchain daemon, as well as
 the command line blockchain client.
+
+## Creating Keys
+
+Run this command replacing `[moniker]` with the name of your key(you get to choose this). 
+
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -4,26 +4,192 @@
...
+```sh
+echo 'export GOBIN=$GOPATH/bin' >> ~/.profile
+```
+```sh
+echo 'export PATH=$PATH:/usr/local/go/bin:$GOBIN' >> ~/.profile
+```
+
+```sh
+. ~/.profile
+```
```
clevinson, line 62: Feels like a lot of lines for just getting someone to setup their gopath stuff. Can we consolidate these into one multiline code snippet?

docs/getting-started.md
```diff
@@ -4,26 +4,192 @@
...
+
+Check to make sure you have the correct version of installed **version 1.15** or higher
+```sh
+go version
+```
+
+Next, we download the official Regen ledger github repository
+```sh
+cd $HOME
+```
```
clevinson, line 72: I don't think this is necessary if we're just asking people immediately afterwards to clone & then cd into the `regen-ledger`

docs/getting-started.md
```diff
@@ -4,26 +4,192 @@
...
+
+```sh
+. ~/.profile
+```
+
+Check to make sure you have the correct version of installed **version 1.15** or higher
+```sh
+go version
+```
```
clevinson, line 67: All of the install golang, git, build-essentials instructions seem to be taking up a lot of real estate for the first thing people read on the "Getting Started" page.

I imagine many folks who will be downloading regen ledger will already have the core dependencies & golang installed, but I do like the idea of us doing a bit of hand-holding for those who need it.

What if instead we put this all on a new page called "Installing Go & Prerequisites", and link to it from here?

docs/getting-started.md
```diff
@@ -4,26 +4,192 @@
...
+
+Once installed the `regen` binary serves as both the blockchain daemon, as well as
 the command line blockchain client.
 
+## Creating Keys
+
+Run this command replacing `[moniker]` with the name of your key(you get to choose this). 
+
+::: **IMPORTANT!**
```
clevinson, line 92: This doesn't render the container properly. It should be something like: 
```suggestion
::: tip IMPORTANT
```

The first word is how vuepress selects the color of the container. Options are "tip", "warning", "danger". You can see previews of them all [here](https://v1.vuepress.vuejs.org/guide/markdown.html#custom-containers).

clevinson: btw- if you want to preview this all locally, you can cd into the docs directory and run `yarn install && yarn run dev`.

docs/getting-started.md
```diff
@@ -4,26 +4,192 @@
...
+
+:::
+
+```sh
+regen keys add [moniker]
+```
+
+During this process you will need to create a password to secure your private keys. **DO NOT LOSE PASSWORD**
+
+## Setting up Cosmovisor
```
clevinson, line 104: I think the whole cosmovisor thing should be an optional piece on a separate page. It's valuable, but only for a subset of users (production validators), and I think app developers or folks who are wanting to play around just want to know the basics of gettting a local chain going.

Maybe a high level we can split this out though, into different user groups?

docs/getting-started.md
```diff
@@ -4,26 +4,192 @@
...
+```sh
+git clone https://github.com/cosmos/cosmos-sdk $GOPATH/src/github.com/cosmos/cosmos-sdk
+```
+
+Enter cosmos-sdk and use correct version
+```sh
+cd $GOPATH/src/github.com/cosmos/cosmos-sdk
+```
+```sh
+git checkout v0.40.0
```
clevinson, line 125: This shouldn't be necessary, folks should be building cosmovisor from the same version that they are building regen ledger most of the time.

I think ppl should just be running `go get` as described in the cosmovisor readme:
https://github.com/cosmos/cosmos-sdk/tree/master/cosmovisor#installation

If we point folks to just `go get` cosmovisor and read the readme for usage guidelines, I think we can actually leave out most of this section. I think it would be nicer to point to existing documentation where possible rather that duplicating instructions incase they get updated later.

docs/getting-started.md
```diff
@@ -4,26 +4,192 @@
...
+mkdir -p ~/.regen/cosmovisor
+```
+```sh
+mkdir -p ~/.regen/cosmovisor/genesis/bin
+```
+```sh
+cp $GOBIN/regen ~/.regen/cosmovisor/genesis/bin
+```
+
+In the `cosmovisor` directory create the system files. Copy the entire code block below and paste in your shell and hit enter
```
clevinson, line 150:
```suggestion
In the `cosmovisor` directory create the systemd files. Copy the entire code block below and paste in your shell and hit enter
```

docs/getting-started.md
```diff
@@ -118,7 +279,7 @@ and your node should start producing blocks.
 
 ### Check the Node's Status
 
-Regardless of whether you're running a local node (as above), or connecting directly to a live network, you can run `regen status` to verify that the CLI is able to connect to the blockchain node. If connecting to a live network, make sure you provide a `--node` flag with the correct address of a live peer. For connecting to `regen-devnet-1`, use the following:
+Regardless of whether you're running a local node (as above), or connecting directly to a live network, you can run `regen status` to verify that the CLI is able to connect to the blockchain node. If connecting to a live network, make sure you provide a `--node` flag with the correct address of a live peer. For example if you are connecting to `regen-devnet-1`, use the following:
```
clevinson, line 282:
```suggestion
Regardless of whether you're running a local node (as above), or connecting directly to a live network, you can run `regen status` to verify that the CLI is able to connect to the blockchain node. If connecting to a live network, make sure you provide a `--node` flag with the correct address of a live peer. For example if you are connecting to `regen-devnet-3`, use the following:
```

clevinson (changes requested):
Nice work @dpdanpittman !

In general I think this looks a bit long now, and wondering if we can tighten up the non regen-ledger specific sections or either pull tehm out to a separate page, or just link to external documentation (like [here](https://github.com/cosmos/cosmos-sdk/tree/master/cosmovisor#installation) for cosmovisor).

Additionally, it'd be great to give a bit more context setting at the beginning about what regen network is (just a few short sentences), and start speaking to different user groups. If you'd like, we can also just merge this with a few cleanups and then tackle the user personas language in a follow-up PR later.

wdyt?

dpdanpittman:
I agree let me break this out to more user specific sections and I'll just link to the original docs for redundant info such as setting up go.

PR #261, opened 2021-02-17
remove casttype

likhita-809:
closes: #261


x/group/server/testsuite/suite.go
```diff
@@ -203,26 +203,27 @@ func (s *IntegrationTestSuite) TestCreateGroup() {
...
 			if spec.expErr {
 				s.Require().Error(err)
-				_, err := s.queryClient.GroupInfo(s.ctx, &group.QueryGroupInfoRequest{GroupId: group.ID(seq + 1)})
+				_, err := s.queryClient.GroupInfo(s.ctx, &group.QueryGroupInfoRequest{GroupId: group.ID(seq + 1).Uint64()})
 				s.Require().Error(err)
 				return
 			}
 			s.Require().NoError(err)
-			id := res.GroupId
+			// id := group.ID(res.GroupId)
```
blushi, line 211:
```suggestion
```

x/group/server/testsuite/suite.go
```diff
@@ -1947,7 +1948,7 @@ func createProposal(
 
 	proposalRes, err := s.msgClient.CreateProposal(ctx, proposalReq)
 	s.Require().NoError(err)
-	return proposalRes.ProposalId
+	return group.ProposalID(proposalRes.ProposalId)
```
blushi, line 1951: let's just return `uint64` instead of `group.ProposalID` here, this will avoid back and forth conversions in the rest of the file.

x/group/server/testsuite/suite.go
```diff
@@ -1993,5 +1994,5 @@ func createGroupAndGroupAccount(
 	groupAccountRes, err := s.msgClient.CreateGroupAccount(s.ctx, groupAccount)
 	s.Require().NoError(err)
 
-	return groupAccountRes.GroupAccount, myGroupID, policy
+	return groupAccountRes.GroupAccount, group.ID(myGroupID), policy
```
blushi, line 1997: ditto, let's just return `uint64` instead of `group.ID`

x/group/server/testsuite/suite.go
```diff
@@ -96,15 +96,15 @@ func (s *IntegrationTestSuite) SetupSuite() {
 		Metadata: nil,
 	})
 	s.Require().NoError(err)
-	s.groupID = groupRes.GroupId
+	s.groupID = group.ID(groupRes.GroupId)
```
blushi, line 99: I guess we could just simply make `s.groupID` an `uint64` to avoid back and forth conversions in the tests below.

x/group/types.go
```diff
@@ -165,7 +168,7 @@ var _ orm.Validateable = GroupAccountInfo{}
 func NewGroupAccountInfo(groupAccount sdk.AccAddress, group ID, admin sdk.AccAddress, metadata []byte, version uint64, decisionPolicy DecisionPolicy) (GroupAccountInfo, error) {
```
blushi, line 168: maybe let's just pass in `group` as `uint64` instead of `ID`

blushi (changes requested):
looks good overall but it would be good to adapt some existing functions to avoid unnecessary conversions.

PR #262, opened 2021-02-17
enable build with badgerdb

gotjoshua:


Makefile
```diff
@@ -87,6 +87,9 @@ endif
 ldflags += $(LDFLAGS)
 ldflags := $(strip $(ldflags))
 
+build_tags += $(BUILD_TAGS)
+build_tags := $(strip $(build_tags))
```
robert-zaremba, line 91: why changing the order?

gotjoshua: @robert-zaremba  so that the flags actually get in (instead of overwritten).

Would be great if this could get merged with 0.6.2

gotjoshua: the boltdb tags also get ignored in the old order...

robert-zaremba (approved):
Good, thanks for contribution!

PR #265, opened 2021-02-17
add gir upgrade handler

anilCSE:


blushi (approved):

app/experimental_appconfig.go
```diff
@@ -47,3 +49,9 @@ func setCustomModules(app *RegenApp, interfaceRegistry types.InterfaceRegistry)
 	}
 	/* New Module Wiring END */
 }
+
+func (app *RegenApp) registerUpgradeHandlers() {
+	app.UpgradeKeeper.SetUpgradeHandler("Gir", func(ctx sdk.Context, plan upgradetypes.Plan) {
+		// no-op handler, does nothing
```
amaurym, line 55: would be good to add a comment on why we are adding a no-op upgrade handler, for those who have no context (like me).

Issue #225, opened 2021-01-26
feat: Use ADR 031 when executing proposals


blushi:
Use ADR 031 approach for routing messages within the group module instead of `sdk.Router`.

ref: https://github.com/regen-network/regen-ledger/pull/154#discussion_r562226524

PR #269, opened 2021-02-18
add msg service router

atheeshp:
closes: #225


aaronc (commented):
We could potentially do this as a workaround for now, but what we want to get to is all messages getting routed through the invoker on the module key. Which means bridging the gap between the MsgServiceRouter and the new ADR 033 router. Does that make sense? Maybe we should have a chat together with @blushi to sort this out.

blushi:
@blushi to work on adding support for registering all modules with the ADR 033 router, passing them to the new manager.

Issue #223, opened 2021-01-22
[x/group] Implement AppModule.RegisterInvariants 


blushi:
ref: https://github.com/regen-network/regen-ledger/pull/154#discussion_r562065435

PR #273, opened 2021-02-22
Add RegisterInvariants to Configuration

likhita-809:
Closes: #223 

An Invariant is a function that checks for a particular invariant within a module. These functions are useful to detect bugs early on and act upon them to limit their potential consequences (e.g. by halting the chain). If the invariant has been broken, it should return an error containing a descriptive message about what happened and a boolean indicating whether the invariant has been broken.


x/group/server/invariant.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func resultVoteState(v group.Tally) resTally {
+	return resTally{
+		YesCount:     strToInt(v.YesCount),
+		NoCount:      strToInt(v.NoCount),
+		AbstainCount: strToInt(v.AbstainCount),
+		VetoCount:    strToInt(v.VetoCount),
+	}
+}
```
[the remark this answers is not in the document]
blushi: you should be using "github.com/cockroachdb/apd/v2" instead, see `(t Tally) GetYesCount() (*apd.Decimal, error)`, etc. methods in `x/group/types.go`

x/group/server/invariant.go
```diff
@@ -0,0 +1,91 @@
...
+		_, err = orm.Paginate(it2, &pageReq, &Proposals)
+		if err != nil {
+			return "", false
+		}
+		for i := 0; i < len(Proposals) && i < len(Proposals2); i++ {
+			if int32(Proposals[i].Status) == 1 && int32(Proposals2[i].Status) == 1 {
+				var voteState1 = resultVoteState(Proposals[i].VoteState)
+				var voteState2 = resultVoteState(Proposals2[i].VoteState)
+				if (voteState1.YesCount > voteState2.YesCount) || (voteState1.NoCount > voteState2.NoCount) || (voteState1.AbstainCount > voteState2.AbstainCount) || (voteState1.VetoCount > voteState2.VetoCount) {
+					return sdk.FormatInvariant(group.ModuleName, "Tally-Votes", "\tTallyVoteSums is failed"), true
```
blushi, line 84: could we add a more descriptive message? Describing where the problem is.

x/group/server/invariant.go
```diff
@@ -0,0 +1,91 @@
...
+		return sdk.FormatInvariant(group.ModuleName, "Tally-Votes", "\tTallyVoteSums is failed"), false
+	}
+}
+
+func (s serverImpl) TallyVotesInvariant() sdk.Invariant {
+	return func(sdkCtx sdk.Context) (string, bool) {
+		ctx := types.Context{Context: sdkCtx}
+		var pageReq = query.PageRequest{
+			Offset:     0,
+			Limit:      100,
```
blushi, line 55: This means that we won't verify the proposal votes after 100 proposals, I guess we should just iterate through all the proposals using directly `PrefixScan(ctx, nil, nil)` and not use pagination at all here.

x/group/server/query_server.go
```diff
@@ -176,6 +176,10 @@ func (s serverImpl) getProposalsByGroupAccount(ctx types.Context, account sdk.Ac
 	return s.proposalByGroupAccountIndex.GetPaginated(ctx, account.Bytes(), pageRequest)
 }
 
+func (s serverImpl) getAllProposals(ctx types.Context, pageRequest *query.PageRequest) (orm.Iterator, error) {
+	return s.proposalTable.PrefixScan(ctx, pageRequest.Offset, pageRequest.Limit)
```
blushi, line 180: `PrefixScan` is wrongly used here:
- `pageRequest.Key` is being ignored;
- if `pageRequest.Offset` > 0, then the `end` parameter of `PrefixScan` (`pageRequest.Limit`) is wrong (theoritically, it should be `pageRequest.Offset + pageRequest.Limit`)

I'm aware that it should work in the particular context of `TallyVotesInvariant()` (with default `pageRequest` values) but then let's not expose this `getAllProposals` function then and just use directly `s.proposalTable.PrefixScan(ctx, nil, nil)` in `TallyVotesInvariant()`

blushi (changes requested):

types/module/server/manager.go
```diff
@@ -14,19 +14,30 @@ import (
...
-	cdc              *codec.ProtoCodec
-	keys             map[string]ModuleKey
-	router           *router
-	requiredServices map[reflect.Type]bool
+	baseApp            *baseapp.BaseApp
+	cdc                *codec.ProtoCodec
+	keys               map[string]ModuleKey
+	router             *router
+	requiredServices   map[reflect.Type]bool
+	registerInvariants map[string]RegisterInvariantsHandler
```
blushi, line 22:
```suggestion
	registerInvariantsHandlers map[string]RegisterInvariantsHandler
```

types/module/server/manager.go
```diff
@@ -14,19 +14,30 @@ import (
...
 // NewManager creates a new Manager
 func NewManager(baseApp *baseapp.BaseApp, cdc *codec.ProtoCodec) *Manager {
 	return &Manager{
-		baseApp: baseApp,
-		cdc:     cdc,
-		keys:    map[string]ModuleKey{},
+		baseApp:            baseApp,
+		cdc:                cdc,
+		keys:               map[string]ModuleKey{},
+		registerInvariants: map[string]RegisterInvariantsHandler{},
```
blushi, line 40:
```suggestion
		registerInvariantsHandlers: map[string]RegisterInvariantsHandler{},
```

types/module/server/manager.go
```diff
@@ -97,6 +108,7 @@ func (mm *Manager) RegisterModules(modules []module.Module) error {
 		}
 
 		serverMod.RegisterServices(cfg)
+		mm.registerInvariants[name] = cfg.registerInvariants
```
blushi, line 111:
```suggestion
		mm.registerInvariantsHandlers[name] = cfg.registerInvariants
```

types/module/server/manager.go
```diff
@@ -137,13 +149,16 @@ func (mm *Manager) CompleteInitialization() error {
...
-	cdc              codec.Marshaler
-	requiredServices map[reflect.Type]bool
-	router           sdk.Router
+	msgServer          gogogrpc.Server
+	queryServer        gogogrpc.Server
+	key                *rootModuleKey
+	cdc                codec.Marshaler
+	requiredServices   map[reflect.Type]bool
+	router             sdk.Router
+	registerInvariants RegisterInvariantsHandler
```
blushi, line 161:
```suggestion
	registerInvariantsHandler RegisterInvariantsHandler
```

types/module/server/manager.go
```diff
@@ -156,6 +171,10 @@ func (c *configurator) QueryServer() gogogrpc.Server {
 	return c.queryServer
 }
 
+func (c *configurator) RegisterInvariants(registry RegisterInvariantsHandler) {
```
blushi, line 174:
```suggestion
func (c *configurator) RegisterInvariantsHandler(registry RegisterInvariantsHandler) {
```

types/module/server/manager.go
```diff
@@ -156,6 +171,10 @@ func (c *configurator) QueryServer() gogogrpc.Server {
...
+	c.registerInvariants = registry
```
blushi, line 175:
```suggestion
	c.registerInvariantsHandler = registry
```

types/module/server/module.go
```diff
@@ -20,7 +20,7 @@ type Configurator interface {
 	ModuleKey() RootModuleKey
 	Marshaler() codec.Marshaler
 	RequireServer(interface{})
-
+	RegisterInvariants(registry RegisterInvariantsHandler)
```
blushi, line 23:
```suggestion
	RegisterInvariantsHandler(registry RegisterInvariantsHandler)
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,96 @@
...
+	"math"
+
+	"github.com/regen-network/regen-ledger/orm"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/regen-network/regen-ledger/types"
+	"github.com/regen-network/regen-ledger/x/group"
+)
+
+func (s serverImpl) RegisterInvariants(ir sdk.InvariantRegistry) {
```
blushi, line 14:
```suggestion
func (s serverImpl) RegisterInvariantsHandler(ir sdk.InvariantRegistry) {
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,96 @@
...
+	return func(ctx sdk.Context) (string, bool) {
+		return s.TallyVotesInvariant()(ctx)
+	}
+}
+
+func (s serverImpl) TallyVotesInvariant() sdk.Invariant {
+	return func(sdkCtx sdk.Context) (string, bool) {
+		var msg string
+		var broken bool
+		var curProposal []*group.Proposal
```
blushi, line 28:
```suggestion
		var curProposals []*group.Proposal
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,96 @@
...
+		it1, err := s.proposalTable.PrefixScan(ctx, 1, math.MaxUint64)
+		if err != nil {
+			return "start value must be less than end value in iterator", false
+		}
+		curProposalRowID, err := orm.ReadAll(it1, &curProposal)
+		if err != nil {
+			return "cannot read all proposals in current block", false
+		}
+		_ = curProposalRowID
+		var prevProposal []*group.Proposal
```
blushi, line 39:
```suggestion
		var prevProposals []*group.Proposal
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (s serverImpl) TallyVotesInvariant() sdk.Invariant {
+	return func(sdkCtx sdk.Context) (string, bool) {
+		var msg string
+		var broken bool
+		var curProposal []*group.Proposal
+		ctx := types.Context{Context: sdkCtx}
+		it1, err := s.proposalTable.PrefixScan(ctx, 1, math.MaxUint64)
+		if err != nil {
+			return "start value must be less than end value in iterator", false
```
blushi, line 32: I'm wondering if we shouldn't actually just panic instead in case of error? See for instance in cosmos-sdk, `x/distribution/keeper/invariants.go`

x/group/server/invariants.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		curProposalRowID, err := orm.ReadAll(it1, &curProposal)
+		if err != nil {
+			return "cannot read all proposals in current block", false
```
blushi, line 36: ditto

x/group/server/server.go
```diff
@@ -165,4 +165,5 @@ func RegisterServices(configurator servermodule.Configurator, accountKeeper Acco
 	impl := newServer(configurator.ModuleKey(), configurator.Router(), accountKeeper, configurator.Marshaler())
 	group.RegisterMsgServer(configurator.MsgServer(), impl)
 	group.RegisterQueryServer(configurator.QueryServer(), impl)
+	configurator.RegisterInvariants(impl.RegisterInvariants)
```
blushi, line 168:
```suggestion
	configurator.RegisterInvariantsHandler(impl.RegisterInvariants)
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		_ = curProposalRowID
+		var prevProposal []*group.Proposal
+		if ctx.BlockHeight()-1 >= 0 {
+			sdkCtx = sdkCtx.WithBlockHeight(ctx.BlockHeight() - 1)
+		} else {
+			return "Not enough blocks to perform TallyVotesInvariant", false
```
blushi, line 43: I don't believe we should return false in this case, there's just nothing to compare so the invariant is not actually broken. Also we should just verify this at the very beginning of the function so we don't perform `PrefixScan` and `ReadAll` for nothing, ie:
```go
if ctx.BlockHeight() - 1 < 0 {
 return "Not enough blocks to perform TallyVotesInvariant", true
}
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,96 @@
...
+		if err != nil {
+			return "start value must be less than end value in iterator", false
+		}
+		prevProposalRowID, err := orm.ReadAll(it2, &prevProposal)
+		if err != nil {
+			return "cannot read all proposals from previous block", false
+		}
+		_ = prevProposalRowID
+		for i := 0; i < len(prevProposal) && i < len(curProposal); i++ {
+			if int32(prevProposal[i].Status) == 1 && int32(curProposal[i].Status) == 1 {
```
blushi, line 55: why only verifying for this proposal status value?

blushi (changes requested):
Thanks @likhita-809, I have added a few comments (nits and logic related). I would also feel more confident in accepting this if we could add some tests for `TallyVotesInvariant` if possible.

types/module/server/manager.go
```diff
@@ -14,19 +14,30 @@ import (
...
+	keys               map[string]ModuleKey
+	router             *router
+	requiredServices   map[reflect.Type]bool
+	registerInvariants map[string]RegisterInvariantsHandler
+}
+
+// RegisterInvariants registers all module routes and module querier routes
+func (mm *Manager) RegisterInvariants(ir sdk.InvariantRegistry) {
+	for _, moduleName := range mm.registerInvariants {
+		if moduleName != nil { //TODO: Remove if condition
```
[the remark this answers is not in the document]
likhita-809: Sorry for the inconvenience, I was just cross-checking and forgot to remove that.

x/group/server/invariants.go
likhita-809: Yeah, I'll do that

x/group/server/invariants.go
likhita-809: Okay, I'll fix that
Thanks for the info

x/group/server/invariants.go
anilCSE: @blushi I believe here we are checking for only active proposals. Do we need to check all the proposals? I think that is not necessary

blushi: @anilCSE yeah it makes sense since technically you cannot vote on a non active proposal (ie when `proposal.Status` different from `ProposalStatusSubmitted`) but I thought it would be more secure to double check everything.
In any case, if a proposal is closed at height `i` and it was still active at `i-1`, it's still relevant to compare them, since this means there has been a final tally for the proposal to get closed.

x/group/server/invariants.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+func (s serverImpl) TallyVotesInvariant() sdk.Invariant {
+	return func(sdkCtx sdk.Context) (string, bool) {
+		var msg string
+		var broken bool
+		ctx := types.Context{Context: sdkCtx}
+		if ctx.BlockHeight()-1 < 0 {
+			return sdk.FormatInvariant(group.ModuleName, "Tally-Votes", "Not enough blocks to perform TallyVotesInvariant"), false
```
blushi, line 31:
```suggestion
			return sdk.FormatInvariant(group.ModuleName, "Tally-Votes", "Not enough blocks to perform TallyVotesInvariant"), true
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		sdkCtx = sdkCtx.WithBlockHeight(ctx.BlockHeight() - 1)
+		it1, err := s.proposalTable.PrefixScan(ctx, 1, math.MaxUint64)
```
blushi, line 34: should we rename it to be more consistent with the proposals var naming (ie `curProposals`)?
```suggestion
		curIt, err := s.proposalTable.PrefixScan(ctx, 1, math.MaxUint64)
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,94 @@
...
+		if err != nil {
+			panic(err)
+		}
+		it2, err := s.proposalTable.PrefixScan(sdkCtx, 1, math.MaxUint64)
```
blushi, line 38: ditto
```suggestion
		prevIt, err := s.proposalTable.PrefixScan(sdkCtx, 1, math.MaxUint64)
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,94 @@
...
+		if err != nil {
+			panic(err)
+		}
+		var t require.TestingT
+		var curProposals []*group.Proposal
+		curProposalRowID, err := orm.ReadAll(it1, &curProposals)
```
blushi, line 44:
```suggestion
		_, err := orm.ReadAll(it1, &curProposals)
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,94 @@
...
+		require.NoError(t, err, &curProposals)
+
+		_ = curProposalRowID
+		var prevProposals []*group.Proposal
+		prevProposalRowID, err := orm.ReadAll(it2, &prevProposals)
```
blushi, line 49:
```suggestion
		_, err := orm.ReadAll(it2, &prevProposals)
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,94 @@
...
+		require.NoError(t, err, &curProposals)
+
+		_ = prevProposalRowID
```
blushi, line 52:
```suggestion
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,94 @@
...
+		it2, err := s.proposalTable.PrefixScan(sdkCtx, 1, math.MaxUint64)
+		if err != nil {
+			panic(err)
+		}
+		var t require.TestingT
+		var curProposals []*group.Proposal
+		curProposalRowID, err := orm.ReadAll(it1, &curProposals)
+		require.NoError(t, err, &curProposals)
+
+		_ = curProposalRowID
```
blushi, line 47:
```suggestion
```

types/module/server/testutil.go
```diff
@@ -28,9 +28,10 @@ type FixtureFactory struct {
 	signers []sdk.AccAddress
 	cdc     *codec.ProtoCodec
 	baseApp *baseapp.BaseApp
+	invar   *sdk.InvariantRegistry
 }
 
-func NewFixtureFactory(t *testing.T, numSigners int) *FixtureFactory {
+func NewFixtureFactory(t *testing.T, numSigners int, invariant sdk.InvariantRegistry) *FixtureFactory {
```
blushi, line 34: why adding `invariant` as param here? It's always set as `nil` in the tests so far. On that note, it would be good to add some tests for the group invariants.

blushi (changes requested):

x/group/server/invariants.go
anilCSE: @blushi if the height is <1, why we are calling it as `invariant` broken? I think it should be false here right?

blushi: Oh right, my mistake, thanks!

x/group/server/invariants.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/regen-network/regen-ledger/orm"
+	"github.com/regen-network/regen-ledger/types"
+	"github.com/regen-network/regen-ledger/x/group"
+)
+
+func (s serverImpl) RegisterInvariantsHandler(ir sdk.InvariantRegistry) {
+	ir.RegisterRoute(group.ModuleName, "Tally-Votes", s.TallyVotesInvariant())
+}
+
+func (s serverImpl) AllInvariants() sdk.Invariant {
```
blushi, line 19: is this method really needed?

likhita-809: Yeah, `AllInvariants` is the function that runs all the invariant's functions included in our module

x/group/server/invariants.go
```diff
@@ -0,0 +1,97 @@
...
+	ir.RegisterRoute(group.ModuleName, "Tally-Votes", tallyVotesInvariant(invar))
+}
+
+func AllInvariants(invar Invar) sdk.Invariant {
+	return func(ctx sdk.Context) (string, bool) {
+		return tallyVotesInvariant(invar)(ctx)
+	}
+}
+
+func tallyVotesInvariant(invar Invar) sdk.Invariant {
```
blushi, line 29: keep having some `func (s serverImpl) tallyVotesInvariant() sdk.Invariant`but then call some private function like this one in it

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		fmt.Println(err)
+		panic("create proposal")
+	}
+
+	msg, broken := tallyVotesInvariant(prevProposal, curProposal)
+	fmt.Println(msg, broken)
+	if broken == true {
+		panic("Invariant broken")
+	}
```
blushi, line 111: could you add table tests in order to cover multiple cases (eg invariant broken and not broken)?

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/tendermint/tendermint/libs/log"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	dbm "github.com/tendermint/tm-db"
+
+	"github.com/regen-network/regen-ledger/orm"
+	"github.com/regen-network/regen-ledger/x/group"
+)
+
+func TestTallyVotesInvariant(t *testing.T) {
+	var s serverImpl
```
blushi, line 22: you don't need any `serverImpl` here, just the `proposalTable` is sufficient to test the invariant.

blushi (changes requested):

x/group/server/invariants_test.go
likhita-809: Yeah, sure

x/group/server/invariants_test.go
likhita-809: Okay thanks, I'll do necessary changes

x/group/server/invariants.go
likhita-809: Yeah, sure

types/module/server/testutil.go
likhita-809: Thanks, we'll do necessary changes

types/module/server/manager.go
likhita-809: We should not change RegisterInvariants since this is a part of AppModule interface

x/group/server/invariant.go
likhita-809: Yeah

x/group/server/invariants.go
likhita-809: We should not change RegisterInvariants since this is a part of AppModule interface

x/group/server/invariants.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+func tallyVotesInvariant(prevProposal *group.Proposal, curProposal *group.Proposal) (string, bool) {
+
+	var msg string
+	var broken bool
+	prevYesCount, err := prevProposal.VoteState.GetYesCount()
+	if err != nil {
+		panic(err)
```
blushi, line 62: maybe just return an error instead of all those panics here and below
then just panic in the main function above

x/group/server/invariants.go
```diff
@@ -0,0 +1,98 @@
...
+		require.NoError(t, err, &curProposals)
+
+		for i := 0; i < len(prevProposals) && i < len(curProposals); i++ {
+			msg, broken = tallyVotesInvariant(prevProposals[i], curProposals[i])
+		}
+		return sdk.FormatInvariant(group.ModuleName, "Tally-Votes", msg), broken
+	}
+}
+
+func tallyVotesInvariant(prevProposal *group.Proposal, curProposal *group.Proposal) (string, bool) {
```
blushi, line 56: you could even extract the lines with `s.proposalTable.PrefixScan(...)` and `orm.ReadAll()` from above into this function and just provide the `ctx` and `proposalTable`

types/module/server/manager.go
blushi: I don't get it, `configurator` doesn't implement `AppModule` interface

x/group/server/invariants.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		curIt, err := s.proposalTable.PrefixScan(ctx, 1, math.MaxUint64)
+		if err != nil {
+			panic(err)
+		}
+
+		var t require.TestingT
+		var curProposals []*group.Proposal
+		_, err = orm.ReadAll(curIt, &curProposals)
+		require.NoError(t, err, &curProposals)
```
blushi, line 43: why do we have this test `require` in the middle of some regular code?

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+		prevProposals := spec.prevReq
+		curProposals := spec.curReq
+
+		for i := 0; i < len(prevProposals) && i < len(curProposals); i++ {
+			_, err = proposalTable.Create(prevCtx, prevProposals[i])
+			if err != nil {
+				fmt.Println(err)
+				panic("create proposal")
+			}
```
blushi, line 237:
```suggestion
			require.NoError(t, err)
```

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,249 @@
...
+			_, err = proposalTable.Create(curCtx, curProposals[i])
+			if err != nil {
+				fmt.Println(err)
+				panic("create proposal")
+			}
```
blushi, line 242:
```suggestion
			require.NoError(t, err)
```

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,249 @@
...
+	proposalTable := proposalTableBuilder.Build()
+
+	_, _, addr1 := testdata.KeyTestPubAddr()
+	_, _, addr2 := testdata.KeyTestPubAddr()
+
+	curBlockTime, err := gogotypes.TimestampProto(curCtx.BlockTime())
+	if err != nil {
+		fmt.Println("block time conversion")
+		panic(err)
+	}
```
blushi, line 48:
```suggestion
	require.NoError(t, err)
```

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,249 @@
...
+	interfaceRegistry := types.NewInterfaceRegistry()
+	cdc := codec.NewProtoCodec(interfaceRegistry)
+	key := sdk.NewKVStoreKey(group.ModuleName)
+	db := dbm.NewMemDB()
+	cms := store.NewCommitMultiStore(db)
+	cms.MountStoreWithDB(key, sdk.StoreTypeIAVL, db)
+	err := cms.LoadLatestVersion()
+	if err != nil {
+		panic(err)
+	}
```
blushi, line 32:
```suggestion
	require.NoError(t, err)
```

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,249 @@
...
+	curBlockTime, err := gogotypes.TimestampProto(curCtx.BlockTime())
+	if err != nil {
+		fmt.Println("block time conversion")
+		panic(err)
+	}
+	prevBlockTime, err := gogotypes.TimestampProto(prevCtx.BlockTime())
+	if err != nil {
+		fmt.Println("block time conversion")
+		panic(err)
+	}
```
blushi, line 53:
```suggestion
	require.NoError(t, err)
```

x/group/server/invariants.go
likhita-809: Yeah sure

x/group/server/invariants.go
likhita-809: okay

x/group/server/invariants.go
likhita-809: I'll remove that

app/experimental_appconfig.go
```diff
@@ -72,6 +72,7 @@ func (app *RegenApp) setCustomKeeprs(bApp *baseapp.BaseApp, keys map[string]*sdk
 	govRouter.AddRoute(wasm.RouterKey, wasm.NewWasmProposalHandler(app.wasmKeeper, wasm.EnableAllProposals))
 }
 
+// func setCustomModules(app *RegenApp, interfaceRegistry types.InterfaceRegistry) *servermodule.Manager {
```
blushi, line 75:
```suggestion
```

types/module/server/manager.go
```diff
@@ -232,6 +247,9 @@ func (c *configurator) QueryServer() gogogrpc.Server {
 	return c.queryServer
 }
 
+func (c *configurator) RegisterInvariantsHandler(registry RegisterInvariantsHandler) {
+	c.registerInvariantsHandler = registry
+}
```
blushi, line 252:
```suggestion
}

```

x/group/server/invariants.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func tallyVotesInvariant(ctx sdk.Context, prevCtx sdk.Context, proposalTable orm.AutoUInt64Table) (string, bool) {
+
+	var msg string
+	var broken bool
+
+	prevIt, err := proposalTable.PrefixScan(prevCtx, 1, math.MaxUint64)
+	if err != nil {
+		return err.Error(), true
```
blushi, line 41:
```suggestion
		return msg, broken, err
```
same for all error handling below

x/group/server/invariants.go
```diff
@@ -0,0 +1,99 @@
...
+			return sdk.FormatInvariant(group.ModuleName, "Tally-Votes", "Not enough blocks to perform TallyVotesInvariant"), false
+		}
+		prevCtx, _ := ctx.CacheContext()
+		prevCtx = prevCtx.WithBlockHeight(ctx.BlockHeight() - 1)
+		msg, broken := tallyVotesInvariant(ctx, prevCtx, s.proposalTable)
+		return sdk.FormatInvariant(group.ModuleName, "Tally-Votes", msg), broken
+	}
+}
+
+func tallyVotesInvariant(ctx sdk.Context, prevCtx sdk.Context, proposalTable orm.AutoUInt64Table) (string, bool) {
```
blushi, line 34:
```suggestion
func tallyVotesInvariant(ctx sdk.Context, prevCtx sdk.Context, proposalTable orm.AutoUInt64Table) (string, bool, error) {
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (s serverImpl) tallyVotesInvariant() sdk.Invariant {
+	return func(ctx sdk.Context) (string, bool) {
+		if ctx.BlockHeight()-1 < 0 {
+			return sdk.FormatInvariant(group.ModuleName, "Tally-Votes", "Not enough blocks to perform TallyVotesInvariant"), false
+		}
+		prevCtx, _ := ctx.CacheContext()
+		prevCtx = prevCtx.WithBlockHeight(ctx.BlockHeight() - 1)
+		msg, broken := tallyVotesInvariant(ctx, prevCtx, s.proposalTable)
```
blushi, line 29:
```suggestion
		msg, broken, err := tallyVotesInvariant(ctx, prevCtx, s.proposalTable)
		if err != nil {
		 panic(err)
		}
```

blushi (changes requested):

x/group/server/invariants.go
blushi: from where it's called?

x/group/server/invariants.go
```diff
@@ -0,0 +1,129 @@
...
+
+func tallyVotesInvariant(ctx sdk.Context, prevCtx sdk.Context, proposalTable orm.AutoUInt64Table) (string, bool, error) {
+
+	var msg string
+	var broken bool
+
+	prevIt, err := proposalTable.PrefixScan(prevCtx, 1, math.MaxUint64)
+	if err != nil {
+		msg = err.Error()
+		broken = true
```
blushi, line 46: not sure if that makes sense to set the invariants `broken` to `true` since we couldn't even verify whether or not it was broken because of the error, additionally it will just cause a panic.
Also you're already returning the `err` in itself so no need to add it as part of the returned `msg` too

x/group/server/invariants.go
```diff
@@ -0,0 +1,129 @@
...
+		return msg, broken, err
+	}
+	curIt, err := proposalTable.PrefixScan(ctx, 1, math.MaxUint64)
+	if err != nil {
+		msg = err.Error()
+		broken = true
+		return msg, broken, err
```
blushi, line 53: ditto for all the error handlings below

blushi (changes requested):

x/group/server/invariants.go
likhita-809: I don't think it's called anywhere, if `tallyVotesInvariant` is the only invariant in our module, then we can safely remove this function

likhita-809: I don't think it's called anywhere, if `tallyVotesInvariant` is the only invariant in our module, then we can safely remove this function

app/app.go
```diff
@@ -320,6 +324,8 @@ func NewRegenApp(logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest
 
 	// register experimental modules here
 	app.smm = setCustomModules(app, interfaceRegistry)
+	app.smm.RegisterInvariants(&app.CrisisKeeper)
+	// app.smm = setCustomModules(app, interfaceRegistry)
```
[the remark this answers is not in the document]
likhita-809: will do

types/module/server/manager.go
```diff
@@ -18,23 +18,34 @@ import (
...
+	router                    *router
+	requiredServices          map[reflect.Type]bool
+	initGenesisHandlers       map[string]module.InitGenesisHandler
+	exportGenesisHandlers     map[string]module.ExportGenesisHandler
+	registerInvariantsHandler map[string]RegisterInvariantsHandler
+}
+
+// RegisterInvariants registers all module routes and module querier routes
+func (mm *Manager) RegisterInvariants(ir sdk.InvariantRegistry) {
+	for _, moduleName := range mm.registerInvariantsHandler {
```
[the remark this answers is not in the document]
likhita-809: okay

x/data/server/testsuite/suite.go
```diff
@@ -43,105 +42,105 @@ func (s *IntegrationTestSuite) TearDownSuite() {
 }
 
 func (s *IntegrationTestSuite) TestScenario() {
-	//testContent := []byte("xyzabc123")
-	//mh, err := multihash.Sum(testContent, multihash.SHA2_256, -1)
-	//s.Require().NoError(err)
-	//cid := gocid.NewCidV1(gocid.Raw, mh)
+	// testContent := []byte("xyzabc123")
```
amaurym, line 45: is it still useful to keep commented code committed on master?

x/group/server/invariants.go
```diff
@@ -0,0 +1,97 @@
...
+	"math"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/regen-network/regen-ledger/orm"
+	"github.com/regen-network/regen-ledger/x/group"
+)
+
+func (s serverImpl) RegisterInvariants(ir sdk.InvariantRegistry) {
+	ir.RegisterRoute(group.ModuleName, "Tally-Votes", s.tallyVotesInvariant())
```
amaurym, line 13: I would put `"Tally-Votes"` in a private const in this file

x/group/server/invariants.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return msg, broken, err
+	}
+
+	for i := 0; i < len(prevProposals) && i < len(curProposals); i++ {
+		prevYesCount, err := prevProposals[i].VoteState.GetYesCount()
+		if err != nil {
+			return msg, broken, err
+		}
+		curYesCount, err := curProposals[i].VoteState.GetYesCount()
```
amaurym, line 62: Are we sure that `prevProposals[i] == curProposals[i]` always holds? (I guess it depends on the implementation of orm.ReadAll). Maybe an invariant `prevProposals[i].ProposalID == curProposals[i].ProposalID` could be useful, or at least a comment about this.

Also, we can replace `for i := 0; i < len(prevProposals) && i < len(curProposals)` by `for i := 0; i < len(prevProposals)` right? Since there's no proposal deletion mechanism.

amaurym (commented):
Invariant logic looks correct to me, except one small question

x/group/server/invariants.go
likhita-809: Okay, let's do it that way

x/group/server/invariants.go
likhita-809: Thanks @amaurym 
I too think that we can replace the loop

x/data/server/testsuite/suite.go
likhita-809: I don't really know about this 
Do you have any idea on this @blushi ?

blushi: this is related to the data module which is still WIP so let's keep the comments for now

x/group/server/invariants.go
```diff
@@ -0,0 +1,174 @@
...
+				}
+				membersWeight += curMemWeight
+			}
+		}
+		groupWeight, err := regenMath.ParseNonNegativeDecimal(groupInfo[i].GetTotalWeight())
+		if err != nil {
+			return msg, broken, err
+		}
+		totalMembersWeight := apd.New(membersWeight, 0)
+		if (groupWeight.Cmp(totalMembersWeight) == 1) || (groupWeight.Cmp(totalMembersWeight) == -1) {
```
blushi, line 167: why not just check that `groupWeight.Cmp(totalMembersWeight) != 0`

x/group/server/invariants.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	_, err = orm.ReadAll(memberIt, &groupMembers)
+	if err != nil {
+		return msg, broken, err
+	}
+
+	var membersWeight int64
+	for i := 0; i < len(groupInfo); i++ {
+		membersWeight = 0
+		for j := 0; j < len(groupMembers); j++ {
```
blushi, line 153: instead of looping through all the group members, could you just query group members of the current group instead? (you can use `s.groupMemberByGroupIndex` for that)

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,339 @@
...
+			require.NoError(t, err)
+			_, err = proposalTable.Create(cacheCurCtx, curProposals[i])
+			require.NoError(t, err)
+		}
+		_, broken, _ := tallyVotesInvariant(cacheCurCtx, cachePrevCtx, proposalTable)
+		require.Equal(t, spec.expErr, broken)
+	}
+}
+
+func TestTallyTotalWeightInvariant(t *testing.T) {
```
blushi, line 237:
```suggestion
func TestGroupTotalWeightInvariant(t *testing.T) {
```

blushi (changes requested):
Looks good overall, just some small improvements and renaming requested

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,343 @@
...
+	interfaceRegistry := types.NewInterfaceRegistry()
+	cdc := codec.NewProtoCodec(interfaceRegistry)
+	key := sdk.NewKVStoreKey(group.ModuleName)
+	db := dbm.NewMemDB()
+	cms := store.NewCommitMultiStore(db)
+	cms.MountStoreWithDB(key, sdk.StoreTypeIAVL, db)
+	err := cms.LoadLatestVersion()
+	require.NoError(t, err)
+	curCtx := sdk.NewContext(cms, tmproto.Header{}, false, log.NewNopLogger())
+	curCtx = curCtx.WithBlockHeight(10)
```
blushi, line 247: could you refactor this part since it's used in both tests?

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,343 @@
...
+	prevCtx, _ := curCtx.CacheContext()
+	prevCtx = prevCtx.WithBlockHeight(curCtx.BlockHeight() - 1)
+	// Proposal Table
```
blushi, line 34:
```suggestion

	// Proposal Table
```

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,343 @@
...
+	cdc := codec.NewProtoCodec(interfaceRegistry)
+	key := sdk.NewKVStoreKey(group.ModuleName)
+	db := dbm.NewMemDB()
+	cms := store.NewCommitMultiStore(db)
+	cms.MountStoreWithDB(key, sdk.StoreTypeIAVL, db)
+	err := cms.LoadLatestVersion()
+	require.NoError(t, err)
+	curCtx := sdk.NewContext(cms, tmproto.Header{}, false, log.NewNopLogger())
+	curCtx = curCtx.WithBlockHeight(10)
+	// Group Table
```
blushi, line 248:
```suggestion

	// Group Table
```

x/group/server/invariants_test.go
```diff
@@ -0,0 +1,343 @@
...
+	groupTableBuilder := orm.NewTableBuilder(GroupTablePrefix, key, &group.GroupInfo{}, orm.FixLengthIndexKeys(orm.EncodedSeqLength), cdc)
+	groupTable := groupTableBuilder.Build()
+
+	// Members Table
```
blushi, line 252:
```suggestion
	// Group Member Table
```

x/group/server/invariants.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		return msg, broken, err
+	}
+	_, err = orm.ReadAll(groupIt, &groupInfo)
+	if err != nil {
+		return msg, broken, err
+	}
+
+	var curGroupMembers []*group.GroupMember
+	memIt, err := groupMemberByGroupIndex.PrefixScan(ctx, 1, math.MaxUint64)
```
blushi, line 136: you should get the members iterator within the `groupInfo` loop below and use the group id of the current group:
`it, err := groupMemberByGroupIndex.Get(ctx, orm.EncodeSequence(groupInfo[i].GroupId))`
and then the `if groupInfo[i].GroupId == curGroupMembers[j].GroupId` can be removed

blushi (changes requested):
`groupTotalWeightInvariant` implementation is wrong

x/group/server/invariants.go
likhita-809: okay

x/group/server/invariants.go
likhita-809: okay

x/group/server/invariants.go
```diff
@@ -0,0 +1,174 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	regenMath "github.com/regen-network/regen-ledger/math"
+	"github.com/regen-network/regen-ledger/orm"
+	"github.com/regen-network/regen-ledger/x/group"
+)
+
+const (
+	votesInvariant  = "Tally-Votes"
+	weightInvariant = "Tally-TotalWeight"
```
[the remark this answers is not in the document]
likhita-809: I'll rename it

x/group/server/invariants.go
likhita-809: okay

x/group/server/invariants.go
likhita-809: Okay, I'll do the required changes

x/group/server/invariants.go
```diff
@@ -0,0 +1,171 @@
...
+		if err != nil {
+			return msg, broken, err
+		}
+		_, err = orm.ReadAll(memIt, &curGroupMembers)
+		if err != nil {
+			return msg, broken, err
+		}
+
+		for j := 0; j < len(curGroupMembers); j++ {
+			curMemWeight, err := regenMath.ParseNonNegativeDecimal(curGroupMembers[j].GetMember().GetWeight())
```
blushi, line 150: instead of using `ReadAll` then looping through all the `curGroupMembers` results, you could also directly use the `memIt` with `_, err := memIt.LoadNext(&member)` until there are no more items, when `ErrIteratorDone.Is(err)`
Also please for all the iterators within this file, don't forget to close them using `defer it.Close()` (there are lots of examples of this within the `orm` package)

blushi (changes requested):

x/group/server/invariants.go
likhita-809: Okay

x/group/server/invariants.go
```diff
@@ -0,0 +1,169 @@
...
+	membersWeight := apd.New(0, 0)
+	for i := 0; i < len(groupInfo); i++ {
+		memIt, err := groupMemberByGroupIndex.Get(ctx, groupInfo[i].GroupId)
+		if err != nil {
+			return msg, broken, err
+		}
+		defer memIt.Close()
+		for {
+			_, err = memIt.LoadNext(&groupMember)
+			if err != nil {
```
blushi, line 146: you need to `break` the loop only if `ErrIteratorDone.Is(err)` else return the error

x/group/server/invariants.go
```diff
@@ -0,0 +1,169 @@
...
+	var msg string
+	var broken bool
+
+	var groupInfo []*group.GroupInfo
+
+	groupIt, err := groupTable.PrefixScan(ctx, nil, nil)
+	if err != nil {
+		return msg, broken, err
+	}
+	_, err = orm.ReadAll(groupIt, &groupInfo)
```
blushi, line 130: I actually think now we could just use the same logic than we have for the members to iterate the groups, ie just use the `groupIt` with `LoadNext`

blushi (changes requested):
almost there! just a small fix and nit

x/group/server/invariants.go
likhita-809: Okay, thanks Marie

x/group/server/invariants.go
likhita-809: Okay

amaurym (approved):
I had a look at both invariants, and their tests, it lgtm. Thanks @likhita-809 and @blushi!

amaurym:
One invariant I'm thinking is: the Tally on a proposal should be equal to the sum of all weigthed voted. @blushi Do you think that's useful? (We can definitely add that in another PR).

blushi:
> One invariant I'm thinking is: the Tally on a proposal should be equal to the sum of all weigthed votes. @blushi Do you think that's useful? (If yes, we can definitely add that in another PR).

Indeed, created a new issue to tackle this https://github.com/regen-network/regen-ledger/issues/326

x/group/server/invariants.go
```diff
@@ -0,0 +1,170 @@
...
+	var broken bool
+
+	var groupInfo group.GroupInfo
+
+	groupIt, err := groupTable.PrefixScan(ctx, nil, nil)
+	if err != nil {
+		return msg, broken, err
+	}
+
+	var groupMember group.GroupMember
```
blushi, line 131: could you move up the var declaration to the top (eg below `var groupInfo ...`)?

x/group/server/invariants.go
```diff
@@ -0,0 +1,170 @@
...
+
+	membersWeight := apd.New(0, 0)
+
+	for {
+		_, err := groupIt.LoadNext(&groupInfo)
```
blushi, line 136: since you're not using `ReadAll` anymore (which closed the iterator), you should close `groupIt` too

likhita-809: Sorry for that, thanks @blushi 
Will address it

x/group/server/invariants.go
likhita-809: okay

x/group/server/invariants.go
```diff
@@ -0,0 +1,170 @@
...
+	var broken bool
+
+	var groupInfo group.GroupInfo
+	var groupMember group.GroupMember
+
+	groupIt, err := groupTable.PrefixScan(ctx, nil, nil)
+	if err != nil {
+		return msg, broken, err
+	}
+	defer groupIt.Close()
```
blushi, line 131: could you move the `defer` up before returning the error, right after `groupIt, err := ...`?

x/group/server/invariants.go
```diff
@@ -0,0 +1,170 @@
...
+	for {
+		_, err := groupIt.LoadNext(&groupInfo)
+		if orm.ErrIteratorDone.Is(err) {
+			break
+		}
+		memIt, err := groupMemberByGroupIndex.Get(ctx, groupInfo.GroupId)
+		if err != nil {
+			return msg, broken, err
+		}
+		defer memIt.Close()
```
blushi, line 144: ditto move up

blushi (approved):
pre-approving, pending small change requests

x/group/server/invariants.go
likhita-809: It'll be a lint issue
We should check returned error before deferring `groupIt.Close()` and `memIt.Close()`

blushi: yeah nevermind, the `groupIt` should be nil in case of error anyway so makes sense!

blushi (approved):
Thanks @likhita-809!

PR #277, opened 2021-02-26
bump tendermint and SDK version

robert-zaremba:
+ Tendermint v0.34.8 solves the problem with growing log
+ Evidence in v0.41.0 is broken

So this things should be merged and deployed.


aaronc:
Let's retarget this against the new 0.42

amaurym:
Should we remove the EXPERIMENTAL flag too, since 0.42 does not have authz/feegrant?

clevinson:
Not sure what you mean by remove the EXPERIMENTAL flag? It's justa configuration when building.. Also, we need it in order to not build ecocredit, data module, and group module for the production mainnet.

amaurym:
Oh yeah, forget my comment then, I thought authz & feegrant were also enabled on EXPERIMENTAL=true only.

amaurym (approved):

anilCSE (approved):

aaronc (approved):

PR #281, opened 2021-03-02
add home dir to state export

aleem1314:
##

ref: https://github.com/cosmos/cosmos-sdk/issues/8150#issue-762225676


anilCSE (approved):

amaurym (approved):
utACK, it's the same code as on the SDK

Issue #280, opened 2021-03-02
Drop colon (:) from regen bech32 account prefix


clevinson:
As discussed in discord, we should drop the colon from our bech32 account prefix (`regen:` => `regen`).

PR #282, opened 2021-03-03
Remove colon from regen addresses

atheeshp:
closes: #280

### DIFF
--- a/app/addr_prefixes.go
+++ b/app/addr_prefixes.go
@@ -9,17 +9,17 @@ import (
 
 const (
 	// Bech32PrefixAccAddr defines the Bech32 prefix of an account's address
-	Bech32PrefixAccAddr = "regen:"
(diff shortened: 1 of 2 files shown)


anilCSE (approved):
lgtm

amaurym (approved):
utACK, I didn't check if these changes are exhaustive or not

robert-zaremba:
If we are doing this changes - can we consider making it shorter? Example updates:
* `valoper` -> `vop` or `op`
* `valcons` -> `vcons` or `cons`

aaronc:
> If we are doing this changes - can we consider making it shorter? Example updates:
> 
> * `valoper` -> `vop` or `op`
> 
> * `valcons` -> `vcons` or `cons`

That would be unnecessarily confusing for other cosmos users.

aaronc (approved):

PR #283, opened 2021-03-04
feat: Use ADR 031 / ADR 033 when executing group proposals

blushi:
closes: #225 
ref: https://github.com/regen-network/regen-ledger/pull/269#pullrequestreview-593394700


technicallyty (changes requested):
lgtm overall, just a concern on the panic

types/module/server/msgs.go
```diff
@@ -0,0 +1,82 @@
...
+// This is similar to what is in the cosmos-sdk sdk.Tx
+// and could eventually be merged in.
+func GetMsgs(anys []*types.Any) []sdk.Msg {
+	msgs := make([]sdk.Msg, len(anys))
+	for i, any := range anys {
+		var msg sdk.Msg
+		if isServiceMsg(any.TypeUrl) {
+			req := any.GetCachedValue()
+			if req == nil {
+				panic("Any cached value is nil. Transaction messages must be correctly packed Any values.")
```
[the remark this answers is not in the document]
robert-zaremba: yeah, it's better to return error than repeating the pattern from SDK.

Nodes won't crash, because there is a high level `recover` (try-catch) to handle all that panics. There are tonnes of panics in SDK codebase, because the early developers preferred to panic and catch (recover) rather than chaining errors.

proto/regen/group/v1alpha1/types.proto
```diff
@@ -121,6 +121,11 @@ message GroupAccountInfo {
 
     // decision_policy specifies the group account's decision policy.
     google.protobuf.Any decision_policy = 6 [(cosmos_proto.accepts_interface) = "DecisionPolicy"];
+
+    // path is the "derivation" path of the group account ModuleID,
+    // which is needed to derive the group root module key and execute proposals.
+    bytes path = 7;
```
robert-zaremba, line 127: Let's use `key` or `derivation_key`, rather than path. 
Path is misleading, and we don't do a multi derivation here.

types/module/server/msgs.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+			msg = sdk.ServiceMsg{
+				MethodName: any.TypeUrl,
+				Request:    any.GetCachedValue().(sdk.MsgRequest),
+			}
+		} else {
+			msg = any.GetCachedValue().(sdk.Msg)
```
robert-zaremba, line 55: this can also panic.

proto/regen/group/v1alpha1/types.proto
blushi: yes, generally this PR will need to be updated to be in accordance with https://github.com/regen-network/regen-ledger/pull/313

types/module/server/msgs.go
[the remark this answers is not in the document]
blushi: sure I can update to return an error instead, although these functions are ultimately used in `Get...` functions that will panic in case of error.

types/module/server/msgs.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/cosmos/cosmos-sdk/codec/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// isServiceMsg checks if a type URL corresponds to a service method name,
+// i.e. /cosmos.bank.Msg/Send vs /cosmos.bank.MsgSend.
+// Function copied from cosmos-sdk, once we add ADR 033 and group module to it,
+// we'll want to merge them.
+func isServiceMsg(typeURL string) bool {
```
cyberbono3, line 16: Do u think it is sufficient to count number of slashes here? What If I put `http://`, will it be the the valid typeURL?

cyberbono3 (approved):
Great work!

types/module/server/msgs.go
blushi: This is directly pulled from ADR 031 implementation. `Any` type URLs won't take this form (https://developers.google.com/protocol-buffers/docs/proto3#any) and in the particular context of the sdk for `Msg`s, we do know that they can only look either like  `/cosmos.bank.Msg/Send` or `/cosmos.bank.MsgSend` for instance.

blushi:
@aaronc I'm not sure whether or not we should merge this given https://github.com/cosmos/cosmos-sdk/discussions/9171, wdyt?

robert-zaremba:
What's the status of this PR? Shall we review it?

cyberbono3 (approved):
lgtm

types/module/server/router.go
```diff
@@ -193,3 +210,11 @@ func (rtr *router) testQueryFactory() InvokerFactory {
 		return rtr.invoker(callInfo.Method, nil)
 	}
 }
+
+func TypeURL(req proto.Message) string {
+	return fmt.Sprintf("/%s", proto.MessageName(req))
```
technicallyty, line 215:
```suggestion
	return "/%s" + proto.MessageName(req)
```

iirc there was a small perf PR that made this change in SDK repo, could do it here too

technicallyty (approved):
lgtm, just 2 small nits

types/module/server/manager.go
```diff
@@ -52,8 +52,9 @@ func NewManager(baseApp *baseapp.BaseApp, cdc *codec.ProtoCodec) *Manager {
 		router: &router{
 			handlers:         map[string]handler{},
 			providedServices: map[reflect.Type]bool{},
-			antiReentryMap:   map[string]bool{},
+			legacyRouter:     baseApp.Router(),
```
amaurym, line 55: I asked this question yesterday during Marie's walkthrough: instead of having the ADR-033 router fall back to the legacy router, can it fallback to baseapp's MsgServiceRouter?

types/module/server/router.go
amaurym: Or we can use sdk.MsgTypeURL directly

types/module/server/router.go
```diff
@@ -68,71 +85,71 @@ func (r registrar) RegisterService(sd *grpc.ServiceDesc, ss interface{}) {
...
+			return fmt.Errorf("expected proto.Message, got %T for service method %s", request, methodName)
+		}
+		handler, found := rtr.handlers[TypeURL(req)]
 
-			rtr.antiReentryMap[moduleName] = true
-			defer delete(rtr.antiReentryMap, moduleName)
+		msg, isMsg := request.(sdk.Msg)
+		legacyMsg, isLegacyMsg := request.(legacytx.LegacyMsg)
+		if !found && !isLegacyMsg {
+			return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("cannot find method named %s", methodName))
```
amaurym, line 98: I have a question here: does `invoker` here also handle inter-module queries? I believe in this PR, we only use this for inter-module Msgs.

But if yes, iiuc this would fail. For example if we use ADR-033 to query `cosmos.bank.v1beta1.Balances`, then `found == false` because bank is not registered in this ADR-033 router, and `isLegacyMsg == false`.

x/group/server/msg_server.go
```diff
@@ -265,7 +266,8 @@ func (s serverImpl) CreateGroupAccount(ctx types.Context, req *group.MsgCreateGr
 			return nil, err
 		}
 
-		accountID := s.key.Derive(buf.Bytes())
+		accountDerivationKey = buf.Bytes()
```
amaurym, line 269: `buf` here is defined as the groupAccountSeq, why did we decide that? Can we use for example `accountAddr`? (this would allow us not having `bytes derivation_key = 7;` in the proto def)

amaurym (approved):
lgtm! Thanks for doing the walktrhough yesterday, it helped me go through this PR's review.

types/module/server/router.go
blushi: `sdk.MsgTypeURL` takes a `sdk.Msg` but here that's not limited to msgs

types/module/server/router.go
blushi: yeah it also handles inter-module queries but for it to work, they would have to be registered using the ADR 033 router anyway

x/group/server/msg_server.go
blushi: A given `groupAccountSeq` maps to a given unique group account so we're using it to generate the group account address by deriving the group root module key (`s.key`). We need to store the derivation key because then we need it to call invoke on proposal msg execution. Is this clearer?

types/module/server/manager.go
blushi: I believe that could work too

amaurym:
@blushi Feel free to merge this PR at will!

PR #285, opened 2021-03-05
fix(app): exit with error code and use regen env prefix

clevinson:
When working on #277 I kept having local tests fail despite CI working fine. Turns out that `TestInitCmd()` was looking into the default node home directory (`~/.regen`) for testing out the root command. This is fixed now (we should probably also fix this upstream on the SDK in the simapp tests?).

While investigating, there were a few other updates in usages of `server.Execute()`that I found in the SDK which I wanted to port over here to regen ledger, as it seems this is how we should be interfacing w/ the root command now.

Those PRs were https://github.com/cosmos/cosmos-sdk/pull/8144, and https://github.com/cosmos/cosmos-sdk/pull/7480


app/regen/cmd/cmd_test.go
```diff
@@ -1,23 +1,25 @@
...
 
 func TestInitCmd(t *testing.T) {
+	nodeHome, err := ioutil.TempDir(t.TempDir(), ".regen")
+	require.NoError(t, err)
+
 	rootCmd, _ := cmd.NewRootCmd()
 	rootCmd.SetArgs([]string{
 		"init",          // Test the init cmd
 		"regenapp-test", // Moniker
-		fmt.Sprintf("--%s=%s", cli.FlagOverwrite, "true"), // Overwrite genesis.json, in case it already exists
```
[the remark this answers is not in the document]
aaronc: Hmm I don't think it hurts. There could be edge cases where is needed.

app/regen/cmd/root.go
```diff
@@ -69,26 +66,6 @@ func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
 	return rootCmd, encodingConfig
 }
 
-// Execute executes the root command.
-func Execute(rootCmd *cobra.Command) error {
```
aaronc, line 73: Why is this deleted?

clevinson: In Cosmos SDK simapp we're now using sdk's server.Execute(), i figured this was what we should update to as well. There's no need to have this function defined in regen-ledger if its exposed on the SDK right?

app/regen/cmd/cmd_test.go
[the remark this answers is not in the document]
clevinson: What edge cases are you thinking of? Ideally this should be run as a self contained test with full teardown of any artifacts right? The home directory is created from scratch on each test run, so I don't see how we could expect there to already be a `genesis.json` here.

Previously this line was there because it always looked at your default home directory for running (or created one if it didnt exist). This PR changes that so it creates a temp home dir, which should render the `--overwrite=true` flag unnecessary. Lmk if i'm missing something

aaronc:
@anilCSE could you possibly take a look at this? I don't have very good context.

app/regen/cmd/root.go
clevinson: @anilCSE Hmm... Yeah you're 100% right here. Seems strange then that `server.Execute()` is exposed at all if it doesn't provide a mechanism for chains to add their own prefix... Do you think this should be addressed upstream?

Maybe best to revert this part of my PR for now.

anilCSE: Yes, we can fix this upstream to take an extra argument for this ENV prefix.

amaurym:
Is this PR still relevant? Anyone working on it or should have a look?

app/regen/cmd/root.go
ryanchristo: Opened a pull request in the sdk: https://github.com/cosmos/cosmos-sdk/pull/10950

ryanchristo:
pending https://github.com/cosmos/cosmos-sdk/pull/10950

robert-zaremba:
This PR will need to wait for 0.46 release (see https://github.com/regen-network/regen-ledger/pull/285#discussion_r784394669)

ryanchristo (approved):

Issue #253, opened 2021-02-10
Introduce Clear Module-readiness Criteria


aaronc:
In preparation for mainnet, I would like us to take a more conservative approach to module readiness. I see two pieces to address:

- [x] introduce a `regend` flag for enabling experimental modules which we should do in testnets and devnets (but not the final pre-mainnet testnet)
- [x] tag _all_ new/experimental modules as experimental and _only_ enable them when the `regend` experimental flag is set. This includes: group, authz, feegrant, ecocredit, data and cosmwasm, basically anything that is not already live on the hub
- [ ] define "pre-flight" clear gating criteria for module readiness and only move modules out of experimental once we've checked and re-checked this checklist

I have a number of concerns about module readiness that I think we need to address and I think it's more prudent to not rush things.

I propose to prioritize making modules really production ready in the following order:
1. group
2. ecocredit
3. authz & feegrant
4. data
5. cosmwasm

I would propose the following draft pre-flight checklist and welcome other input:
- [ ] good unit and integration test coverage for all state machine logic
- [ ] internal API audit (at least 1 person)
  - [ ] are `Msg` and `Query` methods and types well-named and organized?
  - [ ] is everything well documented?
- [ ] internal state machine audit (at least 2 people)
  - [ ] read through `MsgServer` code and verify correctness upon visual inspection
  - [ ] ensure all state machine code which could be confusing is properly commented
  - [ ] make sure state machine logic matches `Msg` method documentation
  - [ ] ensure that all state machine edge cases are covered with tests and that test coverage is sufficient
  - [ ] assess potential threats for each method including spam attacks and ensure that threats have been addressed sufficiently. This should be done by writing up threat assessment for each method
  - [ ] assess potential risks of any new third party dependencies and decide whether a dependency audit is needed
- [ ] internal completeness audit (at least 1 person). Verify that the following are properly implemented, ideally with tests:
  - [ ] genesis import and export of all state
  - [ ] query services
  - [ ] CLI methods
- [ ] testnet/devnet testing after above internal audits
  - [ ] all `Msg` methods have been tested especially in light of any potential threats identified
  - [ ] genesis import and export has been tested
- [ ] nice to have (and needed in some cases if threats could be high): official 3rd party audit

clevinson:
We should split this out into an issue for each module (w/ the checklist), and maybe a central place for it.

aaronc:
[...]

PR #287, opened 2021-03-06
Add module readiness checklist issue template

clevinson:
Basic template from @aaronc's suggestions in #253 

Some ideas of things that may be missing from here:
- do we need to audit anything about intermodule wiring in cases where that happens?
- I added a note about documentation (godoc + readme.md). We currently don't have the same docs structure as in the cosmos-sdk (where module directories have .md files with docs that get copied to their vuepress). What is our general documentation expectation for high level docs about things like functionlaity, store usage, API surface, events, etc.?
- Where in this checklist do we check for correct signature verifications? (e.g. that GetSIgners() is setup correctly for all Msg's), anything else abt authentication that we should add in this checklist?
- Do we need to add any base level x/upgrade support (now that we have consensusVersion, etc. from @technicallyty and @amaurym's work?) as part of "Internal completeness audit" ?

### DIFF
--- a/.github/ISSUE_TEMPLATE/module_readiness_checklist.md
+++ b/.github/ISSUE_TEMPLATE/module_readiness_checklist.md
@@ -1,0 +1,25 @@
+---
+name: Module Readiness Checklist
+about: Pre-flight checklist that modules must pass before ready for being enabled in stable builds of regen ledger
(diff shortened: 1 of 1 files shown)


aaronc (approved):
ACK

PR #295, opened 2021-03-15
Likhita/add goreleaser

likhita-809:
Goreleaser adds automation to create a new github release when a new tag is pushed


Makefile
```diff
@@ -14,14 +14,33 @@ DOCKER_BUF := docker run -v $(shell pwd):/workspace --workdir /workspace bufbuil
...
 
+GORELEASER_CONFIG      = .goreleaser.yaml
+
 # process build tags
 
 build_tags = netgo
 
 ifeq ($(EXPERIMENTAL),true)
 	build_tags += experimental
+	GORELEASER_BUILD_TAGS=$(build_tags)
```
anilCSE, line 25: These are not being used it seems. Can we remove them safely?

Makefile
```diff
@@ -14,14 +14,33 @@ DOCKER_BUF := docker run -v $(shell pwd):/workspace --workdir /workspace bufbuil
...
+	GORELEASER_HOMEBREW_CUSTOM=keg_only :unneeded, \"This is an experimental release.\"
+else
+	GORELEASER_BUILD_TAGS=$(BUILD_TAGS)
+	GORELEASER_HOMEBREW_NAME=regen
+	GORELEASER_HOMEBREW_CUSTOM=
 endif
 
+GORELEASER_TAG     ?= $(shell git describe --tags --abbrev=0)
+
+GORELEASER_FLAGS    = -tags="$(GORELEASER_BUILD_TAGS)"
```
anilCSE, line 36: Ditto

.github/workflows/goreleaser.yml
```diff
@@ -0,0 +1,39 @@
+name: goreleaser
+
+on:
+  pull_request:
```
anilCSE, line 4: We can remove on pull_request

clevinson:
Thanks @likhita-809!

Can you add a short description to the PR so folks who maybe won't have as much context can easily understand what you've setup?

.goreleaser.yml
```diff
@@ -0,0 +1,53 @@
...
+      - -tags="netgo ledger"
+      - -trimpath
+      - -mod=readonly
+    ldflags:
+      - -s -w -X github.com/cosmos/cosmos-sdk/version.Name=regen -X github.com/cosmos/cosmos-sdk/version.AppName=regen -X github.com/cosmos/cosmos-sdk/version.Version={{.Version}} -X github.com/cosmos/cosmos-sdk/version.Commit={{.Commit}} -X "github.com/cosmos/cosmos-sdk/version.BuildTags=netgo,ledger"
+archives:
+  - id: regen
+    builds:
+      - regen-darwin-amd64
+      - regen-linux-amd64
```
clevinson, line 39: Can we make separate builds & archives for experimental vs stable?

It should be as simple as creating 2 new `build` sections above with the `experimental` tag (in addiiton `tags="netgo ledger"`), and then referencing htem here in the archives section.

Makefile
```diff
@@ -14,6 +14,8 @@ DOCKER_BUF := docker run -v $(shell pwd):/workspace --workdir /workspace bufbuil
 
 export GO111MODULE = on
 
+GORELEASER_CONFIG      = .goreleaser.yaml
```
clevinson, line 17: Why does Makefile need to know the goreleaser config? Does this get used by anything in here?

.github/workflows/goreleaser.yml
```diff
@@ -0,0 +1,38 @@
...
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      -
+        name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
```
clevinson, line 30: `go.mod` depends on go-version 1.15, Should we have goreleaser use the same version?

.goreleaser.yml
anilCSE: @clevinson this might confuse people if we have two binaries for each release. May be we can fix this by,
- use odd number versioning for expermental tags (v1.1.x, v1.3.x, v3.1.x etc)
- use even number versioning for stable releases (v1.0.x, v1.2.x, etc)

anilCSE: That looks odd. May be we can just follow -experimental for the build name and have two binaries as you suggested

Makefile
likhita-809: No, I thought it would be necessary
But after a few tests, I got to know that it is not necessary
Thanks @clevinson

.goreleaser.yml
likhita-809: We cannot build separate builds & archieves for experimental & stable, since both of them uses the same `regen` binary for `linux_amd64` and `darwin_amd64`

PR #299, opened 2021-03-15
Move cosmwasm into experimental build

### LINKED ISSUE #296: Move cosmwasm into experimental Module

anilCSE:
closes #296

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -165,7 +165,8 @@ type RegenApp struct {
 	IBCKeeper        *ibckeeper.Keeper // IBC Keeper must be a pointer in the app, so we can SetRouter on it correctly
 	EvidenceKeeper   evidencekeeper.Keeper
 	TransferKeeper   ibctransferkeeper.Keeper
-	wasmKeeper       wasm.Keeper
+	//nolint:structcheck
+	wasmKeeper       wasm.Keeper // only used in experimental builds
 
 	// make scoped keepers public for test purposes
 	ScopedIBCKeeper      capabilitykeeper.ScopedKeeper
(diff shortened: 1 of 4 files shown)


amaurym (commented):
lgtm, some lint fixes needed.

You can only add build flags per file, right? Is there a way to add build flags per line too? (I'm thinking on the `wasmKeeper` field on the App struct.

Also `app/wasm_wrapper.go` can be under the experimental flag too, right?

anilCSE:
> lgtm, some lint fixes needed.
> 
> You can only add build flags per file, right? Is there a way to add build flags per line too? (I'm thinking on the `wasmKeeper` field on the App struct.
> 

Yes we can only add build flags for files

> Also `app/wasm_wrapper.go` can be under the experimental flag too, right?

Oh yes. Thanks, will update

amaurym (approved):
lgtm, pending on lint check. Maybe we can also add some comments `// only used in experimental builds` for the places where we add `//nolint`

clevinson (approved):

clevinson:
Is there an easy way to also remove `wasmd` from go.mod when creating stable builds?

The reason I ask is that currently `make localnet-start` is broken on OSX. This is bc of a bug in wasmd where you cannot build for linux (e.g. `GOOS=linux GOARCH=amd64 make build` fails on OSX when go.mod includes wasmd).

If there's a way to have certain dependencies be build-flag dependent that would be awesome.

Regardless we should file an issue with Wasmd, or maybe move to running `make build-linux` in a docker container instaed of using `GOOS=linux GOARCH=amd64` vars.

PR #307, opened 2021-03-25
feat: add safer Dec struct for math operations

aaronc:
In https://github.com/regen-network/mainnet/pull/15, it was discovered that using `apd.Decimal` directly can be unsafe because apd operations mutate the underlying `Decimal`, but when copying the `big.Int` structure can be shared between `Decimal` instances causing corruption. This PR introduces a `Dec` wrapper which does NO mutation of `apd.Decimal`s, instead creating a new `apd.Decimal` for every operation so that usage is safe.


clevinson:
Marking this PR as ready for review, after adding a few additional test and updating to master.

Next follow-ups after this afaiu are:
- add property based test for new `Dec` struct
- refactor all usages of `ParseNonNegativeFixedDecimal`, etc. to be based off of the internal `Dec` struct instead (I can do this, but might be worth waiting until #428 is merged as this changes some of the precision functionality)

types/math/dec.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/cockroachdb/apd/v2"
+	"github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+type Dec struct {
+	dec apd.Decimal
+}
+
+var dec128Context = apd.Context{
+	Precision:   34,
```
robert-zaremba, line 13: Why 34. Can we add a comment to it (eg link to a related discussion)?

types/math/dec.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (x Dec) Sub(y Dec) (Dec, error) {
+	var z Dec
+	_, err := apd.BaseContext.Sub(&z.dec, &x.dec, &y.dec)
+	return z, err
+}
+
+// SafeSub subtracts the value of y from x and stores the result in res with arbitrary precision only
+// if the result will be non-negative. An insufficient funds error is returned if the result would be negative.
```
robert-zaremba, line 46:
```suggestion
// SafeSub subtracts the value of y from x and stores the result in res with arbitrary precision.
// Returns ErrInsufficientFunds error if the result is negative.
```

types/math/dec.go
```diff
@@ -0,0 +1,126 @@
...
+	return Dec{*d}, nil
+}
+
+func NewDecFromInt64(x int64) Dec {
+	var res Dec
+	res.dec.SetInt64(x)
+	return res
+}
+
+func (x Dec) Add(y Dec) (Dec, error) {
```
robert-zaremba, line 33:
```suggestion
// Add returns new Dec with value `x+y` without mutating any argument and error if 
// there is an overflow.
func (x Dec) Add(y Dec) (Dec, error) {
```

robert-zaremba: Shall we add similar comment to other functions?

types/math/dec.go
```diff
@@ -0,0 +1,126 @@
...
+	var z Dec
+	_, err := apd.BaseContext.Add(&z.dec, &x.dec, &y.dec)
```
robert-zaremba, line 35: let's add `SafeAdd` or check here (in `Add`) that both `x` and `y` are not negative.

robert-zaremba (commented):
Looks good. Left few cosmetic comment and one concern about `Add` (maybe we should have `SafeAdd` as well).

types/math/dec_test.go
```diff
@@ -0,0 +1,78 @@
+package math
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestDec(t *testing.T) {
```
ruhatch, line 9: Yup, this looks like a good spot for some property tests :D

Even if we don't add those in this PR, perhaps it would be good to identify some properties we'd expect these types to have? They can just be in comments in some basic mathematical notation for now.

ruhatch (changes requested):
I'd like to see some more documentation about this new type and why we're using it, plus some basic outlines of the properties we expect to hold

types/math/dec.go
clevinson: What's wrong with either `x` or `y` being negative? There's tests in here that include usage of negative `Dec` values, I'm not sure why we would want to enforce at the library level that this function cannot work with negative values.

Can you explain the use case for a `SafeAdd` ? `SafeSub` makes more sense to me, as verifying that balance is non-zero is in fact something we often have to do. I can't really think of a use case for `SafeAdd` though.

types/math/dec.go
```diff
@@ -0,0 +1,126 @@
+package math
+
+import (
+	"github.com/cockroachdb/apd/v2"
+	"github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+type Dec struct {
```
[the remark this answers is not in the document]
clevinson: Why do you say we're trying to use them as unsigned? I think we only are restricting that they're unsigned in `SafeSub`, which will be used sometimes when dealing with token balances (and wanting to ensure a non-negative result when subtracting two numbers).  But more generally speaking, I don't see why we would want to restrict this struct to only work with unsigned decimals.  There could be valid reasons for us storing ecological measurements as negative numbers (e.g. temperature) at some point down the line.

types/math/dec.go
clevinson: This is done!

types/math/dec.go
clevinson: Done!

types/math/dec.go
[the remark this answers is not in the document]
clevinson: Just updated with documentation about the wrapper (quoting from @aaronc in the original PR description).

types/math/dec_test.go
clevinson: @ruhatch  Just added some suggestions as comments in the bottom of `TestDec()`. I'm pretty new to property tests so let me know what you think?

If the decimal nature of this arithmetic means that actual equality isn't really feasible to test, then perhaps we can property test to keep error range as minimal as possible? e.g. `(x/y)*y - x) < 1e-15` or something similar?

types/math/dec.go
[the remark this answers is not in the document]
ruhatch: > Why do you say we're trying to use them as unsigned? I think we only are restricting that they're unsigned in `SafeSub`, which will be used sometimes when dealing with token balances (and wanting to ensure a non-negative result when subtracting two numbers). But more generally speaking, I don't see why we would want to restrict this struct to only work with unsigned decimals. There could be valid reasons for us storing ecological measurements as negative numbers (e.g. temperature) at some point down the line.

Okay, well then I think we're doing some mixing of concepts here. I don't think the decimal type should be holding information about the fact that it will be used as a non-negative token balance. I.e. I don't think an `InsufficientFunds` error should be thrown from within the API of this type. I also then don't think the name `SafeSub` is quite accurate, because it's not unsafe to subtract and get a negative result.

I guess there's a few options here:
- Have a separate type for explicitly unsigned decimals, for cases such as token balances where they'll never be negative
- Rename `SafeSub` to something like `SubNonNegative` (not catchy)
- Don't expose a `SafeSub` function and do the checking and error throwing elsewhere

Does that make sense?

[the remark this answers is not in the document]
ruhatch: Personally I'm in favour of the first option. That way we're explicit in the type system about the kind of values that we're representing.

types/math/dec.go
ruhatch: I think this relates to https://github.com/regen-network/regen-ledger/pull/307/files#r678209627

types/math/dec_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	require.False(t, minusOne.IsZero())
+	require.False(t, minusOne.IsPositive())
+	require.True(t, minusOne.IsNegative())
+
+
+	/*
+
+	TODO: Add Property Tests
+	========================
```
ruhatch, line 90: These looks nice, thanks :)

ruhatch (commented):
Looks good, just need to make a decision about the `SafeSub` stuff

types/math/dec_test.go
ruhatch: Yeah, some error range property would be great. I'm not sure what the standard should be, as I'm not super experienced with float stuff!

types/math/dec.go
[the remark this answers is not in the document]
ruhatch: I think we could also then have a type alias for token balances that returns the `ErrInsufficientFunds` error in it's own `SafeSub` method.

```
type Balance = UDec

func (*Balance) SafeSub ...
```

types/math/dec.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Quo returns a new Dec with value `x/y` (formatted as decimal128, 34 digit precision) without mutating any
+// argument and error if there is an overflow.
+func (x Dec) Quo(y Dec) (Dec, error) {
+	var z Dec
+	_, err := dec128Context.Quo(&z.dec, &x.dec, &y.dec)
+	return z, errors.Wrap(err, "decimal quotient error")
+}
+
+// Quo returns a new integral Dec with value `x/y` (formatted as decimal128, with 34 digit precision)
```
ruhatch, line 84:
```suggestion
// QuoInteger returns a new integral Dec with value `x/y` (formatted as decimal128, with 34 digit precision)
```

types/math/dec.go
```diff
@@ -0,0 +1,149 @@
...
+// without mutating any argument and error if there is an overflow.
+func (x Dec) QuoInteger(y Dec) (Dec, error) {
+	var z Dec
+	_, err := dec128Context.QuoInteger(&z.dec, &x.dec, &y.dec)
+	return z, errors.Wrap(err, "decimal quotient error")
+}
+
+// Quo returns the integral remainder from `x/y` (formatted as decimal128, with 34 digit precision) without
```
ruhatch, line 92:
```suggestion
// Rem returns the integral remainder from `x/y` (formatted as decimal128, with 34 digit precision) without
```

types/math/dec.go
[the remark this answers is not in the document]
clevinson: So I opted for moving this just out of Dec struct, and just migrated to a function in the math package with signature:
```go
func SubBalance(x Dec, y Dec) (Dec, error)
```

Seems like @aaronc also had desire for a SafeAdd type operation, so I've also implemented an `AddBalance()` similar to the `SubBalance()` above.

I do think we should probably think about refactoring these, but probably better to do when we're actually looking at how these functions are used in context (presumably when actually performing arithmetic on ecocredits). So I'd like to merge this in, and when we get to making use of the `Dec` struct for tracking ecocredit balances, we then take another look at where the "safe" arithmetic options should be.

Also holding that from @aaronc's comments today, it seems that we're trying to align the implementation here with a yet-to-be completed refactor of `sdk.Dec` (note: sdk.Dec does not have an implementation of `SafeSub()`, but [sdk.DecCoin does](https://github.com/cosmos/cosmos-sdk/blob/master/types/dec_coin.go#L312-L317)).  I don't think it makes sense to have a `DecCoin` struct in this package (as ecocredits are not necessarily coins...), but maybe it makes sense to have some struct with more explicit coin-like behavior (e.g. treating it more like an unsigned dec) when refactoring the actual balance tracking to use the Dec?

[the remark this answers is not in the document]
ruhatch: Yeah, I think at some point an explicitly unsigned Dec type would be good. Then we can property test it properly and know that we have some nice safety guarantees on it.

types/math/math.go
```diff
@@ -110,3 +110,37 @@ func SafeSub(res, x, y *apd.Decimal) error {
...
+// AddBalance adds the value of x+y and returns the result with arbitrary precision.
+// Returns with ErrInvalidRequest error if either x or y is negative.
+func AddBalance(x Dec, y Dec) (Dec, error) {
+	var z Dec
+	_, err := exactContext.Add(&z.dec, &x.dec, &y.dec)
+	if err != nil {
+		return z, errors.Wrap(err, "decimal subtraction error")
+	}
+
+	if x.IsNegative() || y.IsNegative() {
```
ruhatch, line 139: Let's do this before doing the actual addition to avoid an unnecessary operation!

ruhatch (approved):
Just one small thing in `AddBalance`, but preapproving

types/math/math.go
```diff
@@ -110,3 +110,37 @@ func SafeSub(res, x, y *apd.Decimal) error {
 
 	return nil
 }
+
+// SubBalance subtracts the value of y from x and returns the result with arbitrary precision.
+// Returns with ErrInsufficientFunds error if the result is negative.
+func SubBalance(x Dec, y Dec) (Dec, error) {
```
aaronc, line 116:
```suggestion
func SafeSubBalance(x Dec, y Dec) (Dec, error) {
```

types/math/math.go
```diff
@@ -110,3 +110,37 @@ func SafeSub(res, x, y *apd.Decimal) error {
...
+	if z.IsNegative() {
+		return z, errors.ErrInsufficientFunds
+	}
+
+	return z, nil
+}
+
+// AddBalance adds the value of x+y and returns the result with arbitrary precision.
+// Returns with ErrInvalidRequest error if either x or y is negative.
+func AddBalance(x Dec, y Dec) (Dec, error) {
```
aaronc, line 132:
```suggestion
func SafeAddBalance(x Dec, y Dec) (Dec, error) {
```

aaronc (commented):
Can't approve for some reason on mobile but ACK

types/math/dec.go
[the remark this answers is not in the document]
robert-zaremba: I prefer `Safe*` - this is what we have in other blockchain and it's more familiar to developers.

types/math/dec.go
robert-zaremba: > What's wrong with either x or y being negative?

@clevinson  we want to have "Safe" operations - operations which are assured to not return negative values. `2 + (-3)` is negative.

types/math/math.go
robert-zaremba: Do we need `Balance` suffix? It doesn't need to be related to a balance. I think `SafeSub` or `CheckedSub` is good enough.

robert-zaremba (commented):
There is a proposal to rename `SubBalance` to `SafeSubBalance`. I think it's a bit of over-engineering. The operation doesn't need to be related to balance. It could be a weight for example (it's not possible to have a negative weight). `Safe*` is widely adopted in Ethereum dev community. In Rust they use `check*`

clevinson:
Keeping with `SafeSubBalance` for now as the error is specific to coins (insufficient funds error), and `SafeAddBalance` should use the same language IMO as `SafeSubBalance`.

I'm open to refactoring once we have a more clear context in which these functions are used, but it seems difficult to bike shed over this API naming when we're not actually seeing its usage (e.g. maybe a different package besides math.go is where this belongs?).  I also like the idea from @ruhatch about tying these functions to unsigned math types.

clevinson (approved):
Approving in place of @aaronc's ACK (@aaronc doesn't have approve privileges, as he was original author)

PR #313, opened 2021-03-30
Use Proper adr-28  addresses for modules.

robert-zaremba:
Currently , we are using wrong address function for derived addresses.


types/id.go
```diff
@@ -1,12 +1,16 @@
...
+)
 
 type ModuleID struct {
 	ModuleName string
 	Path       []byte
 }
 
 func (m ModuleID) Address() sdk.AccAddress {
-	return AddressHash(m.ModuleName, m.Path)
+	return address.Module(m.ModuleName, m.Path)
```
robert-zaremba, line 15: According to variable names, I guess that `m.Path` is a composed key. @blushi , could you confirm??
If so, we shouldn't use `address.Module` here. Instead we should implement `Composed` address function which is using length prefixes (`LengthPrefix(...)`).
@aaronc - could you confirm?

x/group/types.go
```diff
@@ -152,6 +152,7 @@ func (g GroupMember) PrimaryKey() []byte {
 }
 
 func (g GroupAccountInfo) PrimaryKey() []byte {
+	// TODO: the following line fails, because sdk.AccAddressFromBech32 requires 20-byte addresses .
 	addr, err := sdk.AccAddressFromBech32(g.Address)
```
robert-zaremba, line 156: here we have a bigger problem. 
`g.Address` is 32 bytes address. `AccAddressFromBech32` doesn't` accept that addresses and return errors.
Proposed solution:
* update `sdk.AccAddressFromBech32`

types/module/server/derived_module_key.go
```diff
@@ -47,5 +47,6 @@ func (d DerivedModuleKey) ModuleID() types.ModuleID {
 }
 
 func (d DerivedModuleKey) Address() sdk.AccAddress {
+	// TODO: here we should use Composed address,  but ModuleID dosesn't know if it's composed or not.
 	return d.ModuleID().Address()
```
robert-zaremba, line 51: For this use case, `ModuleID.Path` should be list of list of bytes (`[][]byte`)

x/group/types.go
robert-zaremba: solved this by refactoring the `rootModuleKey`

robert-zaremba:
The group module `TestServer` is failing, because the current regen-ledger is not based on Cosmos-SDK v0.43. The SDK version used now doesn't have required address validation updates (accept non 20 character long addresses). Once we will update to v0.43, the issue will be solved.

types/id.go
robert-zaremba: Here, path was not a composed parameter. I refactored the code and made it more clear (I hope).

types/module/server/module_key.go
```diff
@@ -1,20 +1,99 @@
...
+
+type moduleKey struct {
+	moduleName string
+	addr       []byte
+	i          InvokerFactory
+}
+
+// NewDerivedModuleKey creates a ModuleKey with a derived moduel address based on parent
+// module address and derivation key.
+func NewDerivedModuleKey(modName string, parentAddr, derivationKey []byte, i InvokerFactory) ModuleKey {
```
blushi, line 36: why is this public?

robert-zaremba: Because the original structure - `DerivedModuleKey` - was public, so I though you want to have it exposed.

robert-zaremba:
Shall we update this branch and merge it? I think we can do it as a part of next Regen release. What do you think?
+ need to clean up `/types/address`
+ fix conflicts

If yes, I can clean it this week.

blushi:
> Shall we update this branch and merge it? I think we can do it as a part of next Regen release. What do you think?
> 
> * need to clean up `/types/address`
> * fix conflicts
> 
> If yes, I can clean it this week.

Yes I think so, that's also kind of needed for #283

types/address/hash.go
```diff
@@ -0,0 +1,47 @@
+package address
+
+// TODO remove this functions and use the latest SDK.
```
blushi, line 3: should this file be removed?

types/module/server/manager.go
```diff
@@ -257,7 +252,7 @@ type configurator struct {
 	sdkmodule.Configurator
 	msgServer                 gogogrpc.Server
 	queryServer               gogogrpc.Server
-	key                       *rootModuleKey
+	key                       RootModuleKey // *rootModuleKey
```
blushi, line 255:
```suggestion
	key                       RootModuleKey
```

types/module/server/module_key.go
```diff
@@ -1,20 +1,99 @@
...
+	sdk.StoreKey
+}
+
+type moduleKey struct {
+	moduleName string
+	addr       []byte
+	i          InvokerFactory
+}
+
+// NewDerivedModuleKey creates a ModuleKey with a derived moduel address based on parent
```
blushi, line 34:
```suggestion
// NewDerivedModuleKey creates a ModuleKey with a derived module address based on parent
```

blushi (commented):
Looks good overall.

types/module/server/module_key.go
```diff
@@ -1,20 +1,99 @@
...
+
+	return invoker(ctx, args, reply)
+}
+
+// NewStream implements ModuleKey interface
+func (d moduleKey) NewStream(context.Context, *grpc.StreamDesc, string, ...grpc.CallOption) (grpc.ClientStream, error) {
+	return nil, fmt.Errorf("unsupported")
+}
+
+// ModuleID implements ModuleKey interface
```
blushi, line 63:
```suggestion
// moduleKey implements ModuleKey interface
```

robert-zaremba: The method name is `ModuleID`, so the comment should start with the method name.

robert-zaremba:
## Update:

When merging the SDK into this branch I encountered few issues related to the `ModuleID` construction, which is:
```
type ModuleID struct {
    Name string
   Path []byte 
} 
```

1. module Name should be unique - so it can be used as an ID 
2. path is used only to derive module key, which in fact is not needed - we could use `name`. Unless we want to have a freedom to customize name in the future.
3. I've changed Path to address in this PR, but this conflicts with the module key construction. Easy solution is to assume that for the ADR-33 module address we will use empty bytes. But firstly we need to decide about (2.)

types/address/hash.go
blushi: up

robert-zaremba: I still need to check with @aaronc  about the ModuleID.

robert-zaremba:
Update: during the call with Aaron:
* we reviewed the routing "signature" check
* discussed to use module subaccount rather than module ID 
* I will check how to update the PR to rework the module ID.

x/group/server/operations.go
```diff
@@ -14,7 +14,8 @@ import (
 func (s serverImpl) WeightedOperations(simState module.SimulationState) []simtypes.WeightedOperation {
 
 	interfaceRegistry := types.NewInterfaceRegistry()
-	queryClient := group.NewQueryClient(s.key)
+	key := s.key.Derive([]byte("client"))
```
robert-zaremba, line 17: I'm deriving a module key, because root module key doesn't provide grpc client connection functionality.

robert-zaremba:
### Summary of changes

+ renamed `ModuleID` to `ModuleAcc`. The structure represents a reference to an account, not to a module ID.
+ reworked root module key and module key:
  + RootModuleKey is used only as a master key, and shall not be used as as an account
  + RootModuleKey is used to derive module accounts
  + ModuleKey is addressable and can be used to further derive keys
  + ModuleKey can be used as a grpc client connection key.
+ proper use of `address.Module` and `address.Derive` ADR-28 functions. The former one is only used to top module accounts - so the ones derived from RootModuleKey.

I'm also considering removing `ModuleAddr.Key` - it's not needed because we store the address. What do you think?

types/id.go
```diff
@@ -1,12 +1,13 @@
...
+// ModuleAcc represents an account managed by a module
+type ModuleAcc struct {
+	// Module name managing this account. Used by router, used when registering a MsgServer
+	// and invoking service methods.
+	Module string
+	// Key for identifying cross-module messages and module address derivation
+	// TODO: probably we don't need it?
+	Key []byte
+	// Address derived from other module account
+	Address []byte
```
aaronc, line 12: Why is this public? Isn't it just a cache of the computed address? In which case it could be private with an accessor method which constructs it if needed?

robert-zaremba: In fact, we don't need `Key` - it's never used. ModuleAcc is used only for the  router by a module key. And the module key copies the `Key` and sets the `Address`. So instead - we can remove `Key` and only use `Address`.

aaronc: It's used for derived module keys. Didn't we talk about that?

blushi:
@robert-zaremba any updates on this? In particular with regards to https://github.com/regen-network/regen-ledger/pull/313#discussion_r660893962

robert-zaremba:
I think all changes related to this PR have been addressed in SDK. We are also thinking about a new module account type which will use ADR-28 functions.

With migration to SDK 0.46 we don't need it any more.

@aaronc could you confirm?

Issue #292, opened 2021-03-11
[x/group] Add documentation


blushi:
At the moment, we only have a README for the group module. We should add more detailed documentation, similar to what we have for existing cosmos-sdk modules.

PR #314, opened 2021-03-31
docs: Add group module documentation

blushi:
closes: #292 

[rendered](https://github.com/regen-network/regen-ledger/blob/marie/292-group-docs/x/group/spec/README.md)


x/group/spec/02_state.md
```diff
@@ -0,0 +1,80 @@
...
+
+The `groupTable` stores `GroupInfo`: `0x0 | []byte(group_id) -> ProtocolBuffer(GroupInfo)`
+
+### groupSeq
+
+The value of `groupSeq` is incremented when creating a new group and corresponds to the new `GroupId`: `0x1 | 0x1 -> BigEndian`
+
+### groupByAdminIndex
+
+`groupByAdminIndex` allows to retrieve groups by admin address using prefix `0x2`.
```
amaurym, line 22: I think it would be interesting for readers to know how the By*Index translates into key/value pairs `0x2 | ??? -> ???` (I would like to know that myself actually)

x/group/spec/02_state.md
```diff
@@ -0,0 +1,80 @@
...
+
+Here's the list of tables and associated sequences and indexes stored as part of the `group` module.
+
+## Group Table
+
+The `groupTable` stores `GroupInfo`: `0x0 | []byte(group_id) -> ProtocolBuffer(GroupInfo)`
+
+### groupSeq
+
+The value of `groupSeq` is incremented when creating a new group and corresponds to the new `GroupId`: `0x1 | 0x1 -> BigEndian`
```
amaurym, line 18: Do we have docs on the ORM module? If yes, might be useful to link those, so that the second `0x1` doesn't come here arbitrarily.

x/group/spec/03_msg_service.md
```diff
@@ -0,0 +1,104 @@
...
+order: 3
+-->
+
+# Msg Service
+
+## Msg/CreateGroup
+
+A new group can be created with the `MsgCreateGroupRequest`, which has an admin address, a list of members and some optional metadata bytes.
+
++++ https://github.com/regen-network/regen-ledger/blob/24290a0f2b219ac831482006952cb7a11d173ecf/proto/regen/group/v1alpha1/tx.proto#L53-L64
```
amaurym, line 11: Should we try the new vuepress syntax with relative paths? https://v1.vuepress.vuejs.org/guide/markdown.html#import-code-snippets

We have netlify preview, if the preview breaks on those snippets, we can fallback to permalinks

x/group/spec/03_msg_service.md
```diff
@@ -0,0 +1,104 @@
+<!--
```
amaurym, line 1: The SDK has a [SPEC-SPEC](https://github.com/cosmos/cosmos-sdk/blob/master/docs/spec/SPEC-SPEC.md), which recommends `03_messages.md` for this file. I propose we follow that (even though those are all service Msgs)

x/group/spec/README.md
```diff
@@ -0,0 +1,50 @@
...
+title: Group
+parent:
+  title: "group"
+-->
+
+## Abstract
+
+This document specifies the group module.
+
+This module allows the creation and management of on chain multisig accounts and enables voting for message execution based on configurable decision policies.
```
amaurym, line 12:
```suggestion
This module allows the creation and management of on-chain multisig accounts and enables voting for message execution based on configurable decision policies.
```

amaurym (commented):
lgtm overall, thanks! Left a couple of comments

x/group/spec/02_state.md
blushi: No there's no doc for the ORM module yet afaik, I have created an issue to track that: #320
In the mean time, I can add a comment to explain that this second `0x1` corresponds to the ORM `sequenceStorageKey`

x/group/spec/03_msg_service.md
blushi: We could try it out
Btw, this made me realize that the `x/group/spec` changes are not reflected in the VuePress app, should we move `x/group/spec/README.md` to `docs/modules/group/` while the group module is still part of regen-ledger so the new groups docs appear on docs.regen.network? cc/ @clevinson @aaronc

x/group/spec/02_state.md
blushi: done

x/group/spec/03_msg_service.md
amaurym: hmm, would simlinking work?

x/group/spec/01_concepts.md
```diff
@@ -1,4 +1,8 @@
-# Group Module
+<!--
```
amaurym, line 1: In the top of this file, there's `It does have an "administrator" which has the weight to add, remove and update members in the group.`

Is weight here correct? Shouldn't it be "power" or "ability" or something like that?

amaurym (approved):
Just read through it again, lgtm, thanks @blushi

aleem1314 (approved):
LGTM

atheeshp (approved):
Lgtm

Issue #326, opened 2021-04-15
Add new group invariant to verify that proposal tally corresponds to the sum of votes 


blushi:
ref: https://github.com/regen-network/regen-ledger/pull/273#issuecomment-819464654

PR #330, opened 2021-04-22
feat: Votes sum invariant

likhita-809:
Closes: #326 
Add an invariant to verify that proposal tally corresponds to the sum of vote weights and proposal VoteState corresponds to the vote choice


x/group/server/invariants.go
```diff
@@ -12,13 +12,15 @@ import (
 )
 
 const (
-	votesInvariant  = "Tally-Votes"
-	weightInvariant = "Group-TotalWeight"
+	votesInvariant              = "Tally-Votes"
+	weightInvariant             = "Group-TotalWeight"
+	proposalTallyVotesInvariant = "Tally-Proposal-Votes-With-Sum-Of-Votes"
```
blushi, line 17: what about this instead?
```suggestion
	VotesSumInvariant = "Tally-Votes-Sum"
```

x/group/server/invariants.go
```diff
@@ -169,3 +181,89 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+
+func proposalTallyInvariant(ctx sdk.Context, proposalTable orm.AutoUInt64Table, voteByProposalIndex orm.UInt64Index) (string, bool, error) {
+
+	var msg string
+	var broken bool
+
+	var proposal group.Proposal
+	var vote group.Vote
+
+	proposalIt, err := proposalTable.PrefixScan(ctx, 1, math.MaxUint64)
```
blushi, line 193: you need to close the `proposalIt`

x/group/server/invariants.go
```diff
@@ -169,3 +181,89 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+		}
+		defer voteIt.Close()
+
+		for {
+			_, err = voteIt.LoadNext(&vote)
+			if orm.ErrIteratorDone.Is(err) {
+				break
+			}
+			voteChoice := vote.GetChoice()
+			if voteChoice != 0 {
```
blushi, line 221:
```suggestion
			if voteChoice != group.Choice_CHOICE_UNSPECIFIED {
```

x/group/server/invariants.go
```diff
@@ -169,3 +181,89 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+	proposalIt, err := proposalTable.PrefixScan(ctx, 1, math.MaxUint64)
+	if err != nil {
+		return msg, broken, err
+	}
+
+	var totalVoteCount int64
+	var totalYesCount int64
+	var totalNoCount int64
+	var totalAbstainCount int64
+	var totalVetoCount int64
```
blushi, line 202: these shouldn't be `int64` but `apd.Decimal`

blushi: also they will need to be reseted for each proposal

blushi (changes requested):

x/group/server/invariants.go
likhita-809: okay

x/group/server/invariants.go
```diff
@@ -12,13 +13,15 @@ import (
...
-	weightInvariant = "Group-TotalWeight"
+	votesInvariant    = "Tally-Votes"
+	weightInvariant   = "Group-TotalWeight"
+	votesSumInvariant = "Tally-Votes-Sum"
 )
 
 func (s serverImpl) RegisterInvariants(ir sdk.InvariantRegistry) {
 	ir.RegisterRoute(group.ModuleName, votesInvariant, s.tallyVotesInvariant())
 	ir.RegisterRoute(group.ModuleName, weightInvariant, s.groupTotalWeightInvariant())
+	ir.RegisterRoute(group.ModuleName, votesSumInvariant, s.proposalTallyInvariant())
```
blushi, line 24: could you change the method name `s.proposalTallyInvariant()` to be more consistent with the route name `votesSumInvariant`? eg `s.tallyVotesSumInvariant()`

x/group/server/invariants.go
```diff
@@ -169,3 +182,91 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+				return msg, broken, err
+			}
+
+			voteChoice := strconv.FormatInt(int64(group.Choice_value[vote.GetChoice().String()]), 10)
+			curVoteWeight, err := regenMath.ParseNonNegativeDecimal(voteChoice)
+			if err != nil {
+				return msg, broken, err
+			}
+
+			if curMemWeight.Cmp(curVoteWeight) != 0 {
```
blushi, line 249: this is wrong, the member's weight is not supposed to be equal to the vote choice, that would mean that a member with weight 2 would always have to vote yes (because `Choice_CHOICE_YES Choice = 2`).

x/group/server/invariants_test.go
```diff
@@ -339,3 +341,263 @@ func TestGroupTotalWeightInvariant(t *testing.T) {
...
+				{
+					ProposalId:          1,
+					Address:             addr1.String(),
+					Proposers:           []string{addr1.String()},
+					SubmittedAt:         *curBlockTime,
+					GroupVersion:        1,
+					GroupAccountVersion: 1,
+					Status:              group.ProposalStatusSubmitted,
+					Result:              group.ProposalResultUnfinalized,
+					VoteState:           group.Tally{YesCount: "2", NoCount: "1", AbstainCount: "0", VetoCount: "0"},
```
blushi, line 422:
```suggestion
					Member: &group.Member{
						Address: addr1.String(),
						Weight:  "3",
					},
				},
				{
					GroupId: 1,
					Member: &group.Member{
						Address: addr2.String(),
						Weight:  "1",
					},
				},
			},
			proposalReq: []*group.Proposal{
				{
					ProposalId:          1,
					Address:             addr1.String(),
					Proposers:           []string{addr1.String()},
					SubmittedAt:         *curBlockTime,
					GroupVersion:        1,
					GroupAccountVersion: 1,
					Status:              group.ProposalStatusSubmitted,
					Result:              group.ProposalResultUnfinalized,
					VoteState:           group.Tally{YesCount: "3", NoCount: "1", AbstainCount: "0", VetoCount: "0"},
```

if you change the test case to this, the invariant shouldn't be broken but with your current implementation, the test fails

x/group/server/invariants_test.go
```diff
@@ -339,3 +341,263 @@ func TestGroupTotalWeightInvariant(t *testing.T) {
...
+					Choice:     group.Choice_CHOICE_YES,
+					SubmittedAt: gogotypes.Timestamp{
+						Seconds: timestamppb.Now().Seconds,
+						Nanos:   timestamppb.Now().Nanos,
+					},
+				},
+			},
+			expErr: true,
+		},
+		"group member's weight must be equal to the correspoding vote": {
```
blushi, line 499:
```suggestion
		"group member's weight must be equal to the corresponding vote": {
```

blushi (changes requested):

x/group/server/invariants.go
likhita-809: okay

x/group/server/invariants.go
likhita-809: okay, I'll make necessary changes

x/group/server/invariants_test.go
likhita-809: okay

x/group/server/invariants.go
```diff
@@ -169,3 +181,125 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+	if err != nil {
+		return msg, broken, err
+	}
+	defer proposalIt.Close()
+
+	totalVotingWeight := apd.New(0, 0)
+	yesVoteWeight := apd.New(0, 0)
+	noVoteWeight := apd.New(0, 0)
+	abstainVoteWeight := apd.New(0, 0)
+	vetoVoteWeight := apd.New(0, 0)
```
blushi, line 204: you need to reset them to 0 after each proposal

blushi (changes requested):

x/group/server/invariants.go
likhita-809: Yeah, I forgot. Sorry for that
Thanks @blushi

blushi: better to move these declarations at the beginning of the loop and then you don't need to reset them at the end of the loop

likhita-809: okay

x/group/server/invariants_test.go
```diff
@@ -339,3 +341,279 @@ func TestGroupTotalWeightInvariant(t *testing.T) {
...
+
+		for i := 0; i < len(groupAcc); i++ {
+			err := groupAcc[i].SetDecisionPolicy(spec.policy)
+			require.NoError(t, err)
+		}
+
+		for i := 0; i < len(groupAcc); i++ {
+			err = groupAccountTable.Create(cacheCurCtx, groupAcc[i])
+			require.NoError(t, err)
+		}
```
atheeshp, line 598: I think this logic can be done in single loop.

x/group/server/invariants_test.go
```diff
@@ -339,3 +341,279 @@ func TestGroupTotalWeightInvariant(t *testing.T) {
...
+
+	curBlockTime, err := gogotypes.TimestampProto(curCtx.BlockTime())
+	require.NoError(t, err)
+
+	specs := map[string]struct {
+		groupAccReq []*group.GroupAccountInfo
+		policy      group.DecisionPolicy
+		membersReq  []*group.GroupMember
+		proposalReq []*group.Proposal
+		voteReq     []*group.Vote
```
atheeshp, line 380: suffix Req is not necessary. wdyt?

x/group/server/invariants_test.go
atheeshp, line 598: I think this logic can be done in single loop.

x/group/server/invariants_test.go
```diff
@@ -339,3 +341,279 @@ func TestGroupTotalWeightInvariant(t *testing.T) {
...
+	_, _, adminAddr := testdata.KeyTestPubAddr()
+	_, _, addr1 := testdata.KeyTestPubAddr()
+	_, _, addr2 := testdata.KeyTestPubAddr()
+
+	curBlockTime, err := gogotypes.TimestampProto(curCtx.BlockTime())
+	require.NoError(t, err)
+
+	specs := map[string]struct {
+		groupAccReq []*group.GroupAccountInfo
+		policy      group.DecisionPolicy
```
atheeshp, line 377: `policy` looks common in every test case, can we remove it from here and add directly below at line number 591?

atheeshp (commented):
looks good, few nits

x/group/server/invariants_test.go
blushi: yeah I also think that just having `groupAccs`, `members`, `proposals` and `votes` would be better

x/group/server/invariants_test.go
likhita-809: Yeah

x/group/server/invariants_test.go
likhita-809: Okay, I'll fix this for all invariants

x/group/server/invariants.go
```diff
@@ -169,3 +181,126 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+		if err != nil {
+			break
+		}
+		groupMemIt, err := groupMemberByGroupIndex.Get(ctx, groupAcc.GroupId)
+		if err != nil {
+			return msg, broken, err
+		}
+		defer groupMemIt.Close()
+
+		for {
```
blushi, line 226: I believe you shouldn't loop through all the members but rather through all the votes for a given proposal.
Imagine that you have a proposal where not all members have voted, when you try to ` voteTable.GetOne` for a member who hasn't voted, it'll return an error

x/group/server/invariants_test.go
```diff
@@ -259,20 +261,20 @@ func TestGroupTotalWeightInvariant(t *testing.T) {
 	_, _, addr2 := testdata.KeyTestPubAddr()
 
 	specs := map[string]struct {
-		groupReq   []*group.GroupInfo
-		membersReq []*group.GroupMember
+		groupsInfo []*group.GroupInfo
+		members    []*group.GroupMember
 		expErr     bool
```
blushi, line 266: please use `expBroken` instead

x/group/server/invariants_test.go
```diff
@@ -339,3 +341,263 @@ func TestGroupTotalWeightInvariant(t *testing.T) {
...
+			_, err = proposalTable.Create(cacheCurCtx, proposals[i])
+			require.NoError(t, err)
+		}
+
+		for i := 0; i < len(votes); i++ {
+			err = voteTable.Create(cacheCurCtx, votes[i])
+			require.NoError(t, err)
+		}
+
+		_, broken, _ := tallyVotesSumInvariant(cacheCurCtx, proposalTable, groupAccountTable, groupMemberByGroupIndex, voteTable)
```
blushi, line 599: check the returned error too

blushi (changes requested):

x/group/server/invariants.go
likhita-809: Yeah, makes sense

x/group/server/invariants.go
```diff
@@ -169,3 +181,129 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+		}
+
+		address, err := sdk.AccAddressFromBech32(proposal.Address)
+		if err != nil {
+			return msg, broken, err
+		}
+
+		err = groupAccountTable.GetOne(ctx, address.Bytes(), &groupAcc)
+		if err != nil {
+			break
```
blushi, line 220:
```suggestion
			return msg, broken, err
```

blushi:
@likhita-809 could you resolve the conflicts?

x/group/server/invariants.go
likhita-809: @blushi I don't think we should be using `return msg, broken, err`, assuming no `GroupAccountInfo` with proposal address, we'll be returning an error which panics on running invariants

blushi: how is it different to the other cases where we do return `msg, broken, err`?

likhita-809: Here, we don't even need to go for other cases since there'll be no such `GroupAccountInfo`

blushi: well if we couldn't find any group account, then that means that we can't verify invariant with the current proposal which belongs to this group account, so for this reason I believe this should error

blushi:
I've been thinking through this invariant a little bit and I believe it might bring some issues in the following case:
Some proposal is created and then there are some votes, the proposal passes and is executed. Later on, the corresponding group member weights are changed, which means that then it will result in a broken invariant because the invariant will get the current weights instead of the ones that were used at the time of the proposal.
So I can see several ways to solve this:
- either we restrict the invariant to active proposals
- or we wait until we have some kind of snapshots implemented

x/group/server/invariants.go
likhita-809: maybe you were right, but `FullAppSimulation` fails because it panics here

likhita-809:
> I've been thinking through this invariant a little bit and I believe it might bring some issues in the following case:
> Some proposal is created and then there are some votes, the proposal passes and is executed. Later on, the corresponding group member weights are changed, which means that then it will result in a broken invariant while it shouldn't, because the invariant will get the current weights instead of the ones that were used at the time of the proposal.
> So I can see several ways to solve this:
> 
> * either we restrict the invariant to active proposals which currently ensures that the member weights haven't changed
> * or we wait until we have some kind of snapshots implemented before going further with this

x/group/server/invariants.go
```diff
@@ -149,23 +168,187 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+		}
+
+		err = groupAccountTable.GetOne(ctx, address.Bytes(), &groupAcc)
+		if err != nil {
+			msg += fmt.Sprintf("group account not found for address: %s\n", proposal.Address)
+			return msg, broken
+		}
+
+		if proposal.GroupAccountVersion != groupAcc.Version {
+			msg += fmt.Sprintf("group account was modified\n")
```
blushi, line 238: could you add info about the group acc address as in the msg just above?

x/group/server/invariants.go
```diff
@@ -149,23 +168,187 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+		}
+
+		err = groupTable.GetOne(ctx, group.ID(groupAcc.GroupId).Bytes(), &groupInfo)
+		if err != nil {
+			msg += fmt.Sprintf("group info not found for group id %d\n", groupAcc.GroupId)
+			return msg, broken
+		}
+
+		if groupInfo.Version != proposal.GroupVersion {
+			msg += fmt.Sprintf("group was modified\n")
```
blushi, line 249: ditto add group id info

x/group/server/invariants.go
```diff
@@ -149,23 +168,187 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+
+			err = groupMemberTable.GetOne(ctx, groupMem.PrimaryKey(), &groupMem)
+			if err != nil {
+				msg += fmt.Sprintf("group member not found with group ID %d and group member %s\n", groupAcc.GroupId, vote.Voter)
+				return msg, broken
+			}
+
+			curMemVotingWeight, err := regenMath.ParseNonNegativeDecimal(groupMem.Member.Weight)
+			if err != nil {
+				msg += fmt.Sprintf("error while parsing non-nengative decimal for group member %s\n", groupMem.Member.Address)
```
blushi, line 276:
```suggestion
				msg += fmt.Sprintf("error while parsing non-negative decimal for group member %s\n", groupMem.Member.Address)
```

x/group/server/invariants.go
```diff
@@ -149,23 +168,187 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+		}
+
+		totalProposalVotes, err := proposal.VoteState.TotalCounts()
+		if err != nil {
+			msg += fmt.Sprintf("error while getting total weighted votes of proposal with ID %d\n", proposal.ProposalId)
+			return msg, broken
+		}
+		proposalYesCount, err := proposal.VoteState.GetYesCount()
+		if err != nil {
+			msg += fmt.Sprintf("error while getting yes votes weight of proposal with ID %d\n", proposal.ProposalId)
```
blushi, line 320:
```suggestion
			msg += fmt.Sprintf("error while getting the weighted sum of yes votes for proposal with ID %d\n", proposal.ProposalId)
```

x/group/server/invariants.go
```diff
@@ -149,23 +168,187 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+			return msg, broken
+		}
+		proposalNoCount, err := proposal.VoteState.GetNoCount()
+		if err != nil {
+			msg += fmt.Sprintf("error while getting no votes weight of proposal with ID %d\n", proposal.ProposalId)
```
blushi, line 325:
```suggestion
			msg += fmt.Sprintf("error while getting the weighted sum of no votes for proposal with ID %d\n", proposal.ProposalId)
```

x/group/server/invariants.go
```diff
@@ -149,23 +168,187 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+			return msg, broken
+		}
+		proposalAbstainCount, err := proposal.VoteState.GetAbstainCount()
+		if err != nil {
+			msg += fmt.Sprintf("error while getting abstain votes weight of proposal with ID %d\n", proposal.ProposalId)
```
blushi, line 330:
```suggestion
			msg += fmt.Sprintf("error while getting the weighted sum of abstain votes for proposal with ID %d\n", proposal.ProposalId)
```

x/group/server/invariants.go
```diff
@@ -149,23 +168,187 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+			return msg, broken
+		}
+		proposalVetoCount, err := proposal.VoteState.GetVetoCount()
+		if err != nil {
+			msg += fmt.Sprintf("error while getting veto votes weight of proposal with ID %d\n", proposal.ProposalId)
```
blushi, line 335:
```suggestion
			msg += fmt.Sprintf("error while getting the weighted sum of veto votes for proposal with ID %d\n", proposal.ProposalId)
```

x/group/server/invariants.go
```diff
@@ -149,23 +168,187 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+			return msg, broken
+		}
+
+		if totalProposalVotes.Cmp(totalVotingWeight) != 0 {
+			broken = true
+			msg += "proposal VoteState must correspond to the sum of votes weights\n"
+			fmt.Printf("Proposal with ID %d has total proposal votes %s, but got sum of votes weights %s\n", proposal.ProposalId, totalProposalVotes.String(), totalVotingWeight.String())
```
blushi, line 342: add this as part of the returned msg

x/group/server/invariants.go
```diff
@@ -149,23 +168,187 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+			break
+		}
+
+		if (yesVoteWeight.Cmp(proposalYesCount) != 0) || (noVoteWeight.Cmp(proposalNoCount) != 0) || (abstainVoteWeight.Cmp(proposalAbstainCount) != 0) || (vetoVoteWeight.Cmp(proposalVetoCount) != 0) {
+			broken = true
+			msg += "proposal VoteState must correspond to the vote choice\n"
+			fmt.Printf("Proposal with ID %d and voter address %s must correspond to the vote choice\n", proposal.ProposalId, vote.Voter)
```
blushi, line 349: ditto

blushi (changes requested):

x/group/server/invariants.go
```diff
@@ -28,95 +31,109 @@ func (s serverImpl) tallyVotesInvariant() sdk.Invariant {
...
+		msg += fmt.Sprintf("error while getting all the proposals at block height %d\n", prevCtx.BlockHeight())
+		return msg, broken
 	}
 
 	for i := 0; i < len(prevProposals); i++ {
 		if prevProposals[i].ProposalId == curProposals[i].ProposalId {
 			prevYesCount, err := prevProposals[i].VoteState.GetYesCount()
 			if err != nil {
-				return msg, broken, err
+				msg += fmt.Sprintf("error while getting yes votes weight of proposal at block height %d\n", prevCtx.BlockHeight())
```
blushi, line 88: should we also specify the error `err` in the returned message? ditto for all the msg below

x/group/server/invariants.go
```diff
@@ -149,23 +168,185 @@ func groupTotalWeightInvariant(ctx sdk.Context, groupTable orm.Table, groupMembe
...
+
+	var groupInfo group.GroupInfo
+	var proposal group.Proposal
+	var groupAcc group.GroupAccountInfo
+	var groupMem group.GroupMember
+	var vote group.Vote
+
+	proposalIt, err := proposalTable.PrefixScan(ctx, 1, math.MaxUint64)
+	if err != nil {
+		msg += fmt.Sprintf("PrefixScan failure on proposal table at block height %d\n%v\n", ctx.BlockHeight(), err)
```
blushi, line 207: I don't think it's necessary to specify the block height in the msg for this invariant

x/group/server/invariants_test.go
```diff
@@ -288,19 +267,173 @@ func TestGroupTotalWeightInvariant(t *testing.T) {
...
+	_, _, adminAddr := testdata.KeyTestPubAddr()
+	_, _, addr1 := testdata.KeyTestPubAddr()
+	_, _, addr2 := testdata.KeyTestPubAddr()
+
+	curBlockTime, err := gogotypes.TimestampProto(curCtx.BlockTime())
+	require.NoError(t, err)
+
+	specs := map[string]struct {
+		groupsInfo   *group.GroupInfo
+		groupAccs    *group.GroupAccountInfo
```
blushi, line 353:
```suggestion
		groupAcc    *group.GroupAccountInfo
```

x/group/server/invariants_test.go
```diff
@@ -288,19 +267,173 @@ func TestGroupTotalWeightInvariant(t *testing.T) {
...
+		groupMembers []*group.GroupMember
+		proposals    *group.Proposal
```
blushi, line 355:
```suggestion
		proposal    *group.Proposal
```

blushi (approved):
Pre-approving, pending small nits

atheeshp (approved):

Issue #288, opened 2021-03-08
[x/group] Try to execute proposal on submission


blushi:
We could add some `tryExec` flag to try to execute a proposal on submission.

ref: https://github.com/regen-network/cosmos-modules/issues/53

blushi:
@clevinson and I were talking about this issue and we were wondering what should be the behavior in the following use case:
- We have a group account with a `ThresholdDecisionPolicy` with 3 as threshold and each group member has a weight of 1
- A proposal is submitted for this group account which should be executed immediately 
- If we have only 2 signers (considered as yes votes), this is not sufficient for the proposal to pass per the decision policy, but should we store those votes in state so that a third voter could vote independently after that for the proposal to get executed?

cc/ @aaronc

robert-zaremba:
My understanding, based on: https://github.com/regen-network/cosmos-modules/issues/53, is that:
* we want to sent a transaction with multiple vote messages: voters will interact and sign transactions off-chain.
* execute immediately in the same transaction.

So, by the spec, the transaction should fail if it can't execute immediately. Hence no won't store the votes. 

However we may consider modifying the spec, and use variant flag:
```
--exec = {try, must}
```
If `--exec=must` - it will follow the interpretation above, if `--exec=try` - it will store the votes and enable future voting for that proposal.

blushi:
> My understanding, based on: [regen-network/cosmos-modules#53](https://github.com/regen-network/cosmos-modules/issues/53), is that:
> 
> * we want to sent a transaction with multiple vote messages: voters will interact and sign transactions off-chain.
> * execute immediately in the same transaction.

My understanding is rather that when creating a new proposal, the signers of the `Msg/CreateProposal` request message would be considered as yes votes (the `proposers`). But yes that means that they will need to sign this transaction off-chain.

> So, by the spec, the transaction should fail if it can't execute immediately. Hence no won't store the votes.
> 
> However we may consider modifying the spec, and use variant flag:
> 
> ```
> --exec = {try, must}
> ```
> 
> If `--exec=must` - it will follow the interpretation above, if `--exec=try` - it will store the votes and enable future voting for that proposal.

Yes, that would make sense. Any thoughts @aaronc @clevinson?

blushi:
I've been thinking through it a little bit, and I'm wondering if this `--exec=try` would really be useful from a user perspective. I guess that if some group uses this, it will likely make sure that the proposal will pass beforehand. 

[...]

PR #333, opened 2021-05-07
feat: Try to execute group proposal on submission or on new vote

blushi:
closes: #288


atheeshp (commented):
looks good

proto/regen/group/v1alpha1/tx.proto
```diff
@@ -239,6 +247,10 @@ message MsgVoteRequest {
 
     // metadata is any arbitrary metadata to attached to the vote.
     bytes metadata = 4;
+
+    // try_exec will try to execute the proposal immediately 
+    // after the vote, if set to true.
+    bool try_exec = 5;
```
robert-zaremba, line 253: shall we use string here `--exec={no, try}` to enable `must` option in the future?

x/group/server/msg_server.go
```diff
@@ -457,6 +457,32 @@ func (s serverImpl) CreateProposal(ctx types.Context, req *group.MsgCreatePropos
 		return nil, err
 	}
 
+	// Try to execute proposal immediately
+	if req.TryExec {
+		// Consider proposers as Yes votes
+		for i := range proposers {
```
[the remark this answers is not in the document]
robert-zaremba: yes, how about `2*loop_step`?

proto/regen/group/v1alpha1/tx.proto
blushi: @robert-zaremba what about using an enum instead?

robert-zaremba: yes, good idea.

x/group/server/msg_server.go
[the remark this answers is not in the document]
blushi: @atheeshp @robert-zaremba what about using something similar to what we have in feegrant for now? https://github.com/cosmos/cosmos-sdk/blob/25ecec6c152aa6406da9da24151f412384a3d94b/x/feegrant/filtered_fee.go#L70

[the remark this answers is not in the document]
robert-zaremba: `gasCostPerIteration` is the `loop_step` in my comment. Usually it should be 10 (as in your code).
Here I suggest to double it, because do more things here.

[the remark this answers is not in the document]
blushi: ok got it

x/group/client/testsuite/tx_test.go
```diff
@@ -1325,6 +1343,23 @@ func (s *IntegrationTestSuite) TestTxVote() {
...
+					"",
+					fmt.Sprintf("--%s=true", client.FlagTryExec),
+				},
+				commonFlags...,
+			),
+			false,
+			"",
+			&sdk.TxResponse{},
+			0,
+		},
```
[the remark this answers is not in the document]
blushi: added, but we have to wait on https://github.com/regen-network/regen-ledger/issues/337 to get fixed before we can actually make sure the new tests pass...

technicallyty:
should there be an update to the `x/group/spec` in this pr?

x/group/server/msg_server.go
```diff
@@ -457,6 +461,33 @@ func (s serverImpl) CreateProposal(ctx types.Context, req *group.MsgCreatePropos
 		return nil, err
 	}
 
+	// Try to execute proposal immediately
+	if req.Exec == group.Exec_EXEC_TRY {
+		// Consider proposers as Yes votes
+		for i := range proposers {
+			ctx.GasMeter().ConsumeGas(gasCostPerIteration, "vote on proposal")
+			_, err = s.Vote(ctx, &group.MsgVoteRequest{
```
technicallyty, line 469: The ref issue suggested we would just assume that all signers of the message would be considered yes votes. is it possible to just set the `YesCount` to the number of signers in the message and then exec? not entirely sure if that would work but at a glance it seems it could be faster than running `Vote` for each proposer

atheeshp: I think the voters' weights can be more than 1 in some cases.

x/group/server/msg_server.go
```diff
@@ -457,6 +461,33 @@ func (s serverImpl) CreateProposal(ctx types.Context, req *group.MsgCreatePropos
...
+				ProposalId: id,
+				Voter:      proposers[i],
+				Choice:     group.Choice_CHOICE_YES,
+			})
+			// Should we return MsgCreateProposalResponse with proposal id?
```
atheeshp, line 474: I am also in same dilemma, may be @amaurym or @robert-zaremba can suggest here.

x/group/client/testsuite/tx.go
```diff
@@ -95,17 +95,21 @@ func (s *IntegrationTestSuite) SetupSuite() {
...
 
-	s.group = &group.GroupInfo{GroupId: 1, Admin: val.Address.String(), Metadata: []byte{1}, TotalWeight: "1", Version: 1}
+	s.group = &group.GroupInfo{GroupId: 1, Admin: val.Address.String(), Metadata: []byte{1}, TotalWeight: "3", Version: 1}
 
-	// create 4 group accounts
-	for i := 0; i < 4; i++ {
+	// create 5 group accounts
+	for i := 0; i < 5; i++ {
+		threshold := i + 1
+		if i+1 > 3 {
```
atheeshp, line 103:
```suggestion
		if threshold > 3 {
```

atheeshp (approved):
looks good

x/group/server/msg_server.go
technicallyty: IMO we should - the proposal has been created so it would be good to return it. Maybe we should also consider wrapping the error with some more description - e.g. `The proposal was created but failed on [vote/exec error]. `

aleem1314 (approved):
lgtm

Issue #336, opened 2021-05-11
Upgrade to 0.43-beta1 and integrate x/authz & x/feegrant into stable


aaronc:
his is ready to start ASAP now that 0.43-beta1 is out.

Important notes:
* this will require a release of ibc-go: https://github.com/cosmos/ibc-go
* we need to setup an upgrade handler that runs migrations
* in the upgrade handler, we need to skip the `x/auth` `ConsensusVersion` upgrade because we patched this into regen-ledger v1.0 (cc @amaurym @technicallyty)

Here's the WIP 0.43 gaia PR: https://github.com/cosmos/gaia/pull/834 for reference

aaronc:
@robert-zaremba, @likhita-809 had already starting working on this integration in #329 and is going to continue working on that PR

likhita-809:
> @robert-zaremba, @likhita-809 had already starting working on this integration in #329 and is going to continue working on that PR

Started working on this

robert-zaremba:
I think testing IBC is _nice to have_ for our team. Instead we should collaborate with the IBC team.

PR #349, opened 2021-05-13
build: Upgrade to 0.43-beta1 and integrate x/authz & x/feegrant into stable

likhita-809:
ref: #336


app/app.go
```diff
@@ -167,13 +169,15 @@ type RegenApp struct {
...
+	FeeGrantKeeper   feegrantkeeper.Keeper
+	AuthzKeeper      authzkeeper.Keeper
+
+	// wasmKeeper wasm.Keeper // only used in experimental builds
 
 	// make scoped keepers public for test purposes
 	ScopedIBCKeeper      capabilitykeeper.ScopedKeeper
 	ScopedTransferKeeper capabilitykeeper.ScopedKeeper
-	ScopedIBCMockKeeper  capabilitykeeper.ScopedKeeper
+	// ScopedIBCMockKeeper  capabilitykeeper.ScopedKeeper
```
anilCSE, line 180: Don't we need this?

app/app.go
```diff
@@ -167,13 +169,15 @@ type RegenApp struct {
 	IBCKeeper        *ibckeeper.Keeper // IBC Keeper must be a pointer in the app, so we can SetRouter on it correctly
 	EvidenceKeeper   evidencekeeper.Keeper
 	TransferKeeper   ibctransferkeeper.Keeper
-	//nolint
-	wasmKeeper wasm.Keeper // only used in experimental builds
+	FeeGrantKeeper   feegrantkeeper.Keeper
+	AuthzKeeper      authzkeeper.Keeper
+
+	// wasmKeeper wasm.Keeper // only used in experimental builds
```
anilCSE, line 175: I believe we are having single RegenApp definition, so we would need wasmkeeper here inorder to make experimental modules work

app/app.go
```diff
@@ -297,12 +304,12 @@ func NewRegenApp(logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest
 
 	// NOTE: the IBC mock keeper and application module is used only for testing core IBC. Do
 	// note replicate if you do not need to test core IBC or light clients.
-	mockModule := ibcmock.NewAppModule(scopedIBCMockKeeper)
+	// mockModule := ibcmock.NewAppModule(scopedIBCMockKeeper)
```
anilCSE, line 307: We might need this for integration tests, don't we? May be this is required only for simapp. Not sure

app/experimental_appconfig.go
```diff
@@ -3,10 +3,8 @@
 package app
 
 import (
-	"path/filepath"
-
-	"github.com/CosmWasm/wasmd/x/wasm"
-	wasmclient "github.com/CosmWasm/wasmd/x/wasm/client"
+	// "github.com/CosmWasm/wasmd/x/wasm"
+	// wasmclient "github.com/CosmWasm/wasmd/x/wasm/client"
```
anilCSE, line 7: Why are we removing wasm from experimental build?

anilCSE (commented):
@likhita-809 looks like the PR is still work-in progress. Why are we removing wasm integration?

likhita-809:
> @likhita-809 looks like the PR is still work-in progress. Why are we removing wasm integration?

wasm is not upgraded with ibc modules

app/app.go
likhita-809: I'm not sure but I think experimental modules works fine even without wasmkeeper

app/app.go
likhita-809: Yes, I think so
I'll add it

app/app.go
likhita-809: Yeah, it is required only for simapp

aaronc:
It's fine to disable wasm for now

app/experimental_appconfig.go
likhita-809: since wasm is not upgraded with ibc module, we're disabling it just for now

anilCSE:
> It's fine to disable wasm for now

Okay, that makes sense. @likhita-809 can you delete the wasm integration code instead of commenting out (atleast from stable config and go.mod)? We can remove this totally. If needed create an issue and mention this commit as reference Integration

x/group/server/testsuite/suite.go
```diff
@@ -120,7 +120,8 @@ func (s *IntegrationTestSuite) SetupSuite() {
 	s.Require().NoError(err)
 	s.groupAccountAddr = addr
 
-	s.Require().NoError(s.bankKeeper.SetBalances(s.sdkCtx, s.groupAccountAddr, sdk.Coins{sdk.NewInt64Coin("test", 10000)}))
+	s.Require().NoError(s.bankKeeper.MintCoins(s.sdkCtx, minttypes.ModuleName, sdk.Coins{sdk.NewInt64Coin("test", 10000)}))
+	s.Require().NoError(s.bankKeeper.SendCoinsFromModuleToAccount(s.sdkCtx, minttypes.ModuleName, s.groupAccountAddr, sdk.Coins{sdk.NewInt64Coin("test", 10000)}))
```
atheeshp, line 124: can't we use `simapp.FundAccount` here? (may be after this https://github.com/cosmos/cosmos-sdk/pull/9315 PR get merged)

app/addr_prefixes.go
```diff
@@ -1,13 +1,15 @@
...
-	"encoding/binary"
 	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/address"
 )
 
 const (
+	// AddrLen defines the length of an account's address
+	AddrLen = 20
```
amaurym, line 12: Why do we still need 20-byte addresses? Afaict it's only used in one test, is that right? If yes, we can probably fix that test too in this PR.

x/group/server/testsuite/suite.go
```diff
@@ -1882,7 +1883,9 @@ func (s *IntegrationTestSuite) TestExecProposal() {
 
 				_, err := s.msgClient.Exec(ctx, &group.MsgExecRequest{Signer: s.addr1.String(), ProposalId: myProposalID})
 				s.Require().NoError(err)
-				s.Require().NoError(s.bankKeeper.SetBalances(ctx.(types.Context).Context, s.groupAccountAddr, sdk.Coins{sdk.NewInt64Coin("test", 10002)}))
+				s.Require().NoError(s.bankKeeper.MintCoins(s.sdkCtx, minttypes.ModuleName, sdk.Coins{sdk.NewInt64Coin("test", 10002)}))
+				s.Require().NoError(s.bankKeeper.SendCoinsFromModuleToAccount(s.sdkCtx, minttypes.ModuleName, s.groupAccountAddr, sdk.Coins{sdk.NewInt64Coin("test", 10002)}))
```
atheeshp, line 1887: ditto

app/addr_prefixes.go
aleem1314: Yeah, this is only used in tests.

amaurym (commented):
lgtm, some small nits

atheeshp (approved):
Looks good

x/group/server/testsuite/suite.go
```diff
@@ -119,14 +116,21 @@ func (s *IntegrationTestSuite) SetupSuite() {
...
-
-	s.Require().NoError(s.bankKeeper.SetBalances(s.sdkCtx, s.groupAccountAddr, sdk.Coins{sdk.NewInt64Coin("test", 10000)}))
+	s.Require().NoError(fundAccount(s.bankKeeper, s.sdkCtx, s.groupAccountAddr, sdk.Coins{sdk.NewInt64Coin("test", 10000)}))
 }
 
 func (s *IntegrationTestSuite) TearDownSuite() {
 	s.fixture.Teardown()
 }
 
+// TODO: https://github.com/cosmos/cosmos-sdk/pull/9315/files/685fb260461fa951c56cfad8efe5f93b6a4f71f3#r630942517
```
amaurym, line 126:
```suggestion
// TODO: https://github.com/cosmos/cosmos-sdk/issues/9346
```

amaurym (approved):
lgtm, also could run the node successfully

amaurym:
@anilCSE I think it's useful for you to take a look too before merging.

app/experimental_appconfig.go
```diff
@@ -102,26 +72,21 @@ func setCustomModules(app *RegenApp, interfaceRegistry types.InterfaceRegistry)
 }
 
 func (app *RegenApp) registerUpgradeHandlers() {
-	app.UpgradeKeeper.SetUpgradeHandler("Mau", func(ctx sdk.Context, plan upgradetypes.Plan) {
-		// no-op handler, does nothing
+	app.UpgradeKeeper.SetUpgradeHandler("Mau", func(ctx sdk.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+		return vm, nil
```
anilCSE, line 76: Should we have the original migration here instead of dummy one? I remember @aleem1314 had worked on this

app/regen/cmd/genaccounts.go
```diff
@@ -19,6 +17,7 @@ import (
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	"github.com/cosmos/cosmos-sdk/x/genutil"
 	genutiltypes "github.com/cosmos/cosmos-sdk/x/genutil/types"
+	"github.com/spf13/cobra"
```
anilCSE, line 20: gofmt needed

app/experimental_appconfig.go
anilCSE: Oh I just found that it is added in stable config. Then we should remove it from here.

Also why are we integrating these new modules in stable and experimental configs?

anilCSE: If we are adding something to both configs, we should directly add it to `app.go`

app/regen/cmd/genaccounts.go
likhita-809: It is up-to-date with gofmt format, I cross-checked too

app/app.go
```diff
@@ -136,14 +144,15 @@ func init() {
 
 	// this changes the power reduction from 10e6 to 10e2 for the regen-test-1001 testnet which will give
 	// every validator 10,000 times more voting power than they currently have
-	sdk.PowerReduction = sdk.NewIntFromBigInt(new(big.Int).Exp(big.NewInt(10), big.NewInt(2), nil))
+	sdk.DefaultPowerReduction = sdk.NewIntFromBigInt(new(big.Int).Exp(big.NewInt(10), big.NewInt(2), nil))
```
anilCSE, line 147: @likhita-809 let's remove this

anilCSE (approved):
lgtm, just one comment

app/app.go
anilCSE: @likhita-809 let's remove this

PR #354, opened 2021-05-20
Fix migration upgrade handler

amaurym:


app/experimental_appconfig.go
```diff
@@ -73,13 +73,45 @@ func setCustomModules(app *RegenApp, interfaceRegistry types.InterfaceRegistry)
...
+		panic(err)
+	}
+
+	if upgradeInfo.Name == "v0.43.0-beta1-upgrade" && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
+		storeUpgrades := storetypes.StoreUpgrades{
+			Added: []string{"authz", "feegrant"},
+		}
+
+		// configure store loader that checks if version == upgradeHeight and applies store upgrades
+		app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
```
amaurym, line 113: Should we add this logic into the SDK directly? In `RunMigrations` more specifically, if `fromVM["modulename"]`==0, then we add the storeloader, and we run InitGenesis

app/experimental_appconfig.go
```diff
@@ -73,13 +73,45 @@ func setCustomModules(app *RegenApp, interfaceRegistry types.InterfaceRegistry)
 }
 
 func (app *RegenApp) registerUpgradeHandlers() {
-	app.UpgradeKeeper.SetUpgradeHandler("v0.43.0-beta1-upgrade", func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
-
-		// skipping x/auth migrations. It is already patched in regen-ledger v1.0
-		fromVM["auth"] = auth.AppModule{}.ConsensusVersion()
+	app.UpgradeKeeper.SetUpgradeHandler("v0.43.0-beta1-upgrade", func(ctx sdk.Context, plan upgradetypes.Plan, _ module.VersionMap) (module.VersionMap, error) {
+		// 1st-time running in-store migrations, using 1 as fromVersion to
```
amaurym, line 77: btw, shouldn't all this be copied to stable_appconfig.go too? As I see,
- stable: v043, authz, feegrant
- experimental: above + group, data, ecocredit

anilCSE: Ah okay. I thought we are still keeping authz in expermental. Yes, let's move them to stable config

anilCSE: we should also create an issue to remove this upgrade handler before tagging rc-1

app/experimental_appconfig.go
anilCSE: That's better UX. I would advise to do that. But how are we gonna pass the storekeys there?

amaurym: Let's discuss here: https://github.com/cosmos/cosmos-sdk/issues/9381

aleem1314 (approved):

anilCSE (approved):
lgtm

PR #322, opened 2021-04-09
refactor: Split regen ledger into multiple go modules

cyberbono3:
Closes #319

Overview:
1. Move testutil to types/testutil and made it standalone go module.
2. Make every x/module standalone go module.
3. Make types and orm  standalone go module.
4. Move the logic that depends on app from types/testutil to app/modules_test.go to break cycle dependency `app -> x/module-> types/testutil -> app`
5. Move codec.proto, codec.pb.go, model.go from types/testdata to orm/testdata
6. Generate codec.pb.go in orm/testdata
7. Move tx.* files to x/group/testdata
8. Generate tx.pb.go in x/group/testdata
9. Move client/msg_service.go to types/module/msg_service.go and move util/utils.go to x/ecocredit/util/utils.go

PR #358, opened 2021-05-26
fix: fix tests

aleem1314:


Makefile
```diff
@@ -251,13 +251,25 @@ test-race: TEST_PACKAGES=$(PACKAGES_NOSIMULATION)
 
 $(TEST_TARGETS): run-tests
 
+SUB_MODULES = $(shell find . -type f -name 'go.mod' -printf '%h\n' | sort)
+CURRENT_DIR = $(shell pwd)
 run-tests:
 ifneq (,$(shell which tparse 2>/dev/null))
 	#go test -mod=readonly -json $(ARGS) $(TEST_PACKAGES) | tparse
-	find . -name go.mod -execdir go test ./... \;
+	# find . -name go.mod -execdir go test ./... \;
```
aleem1314, line 259:
```suggestion
```

Makefile
```diff
@@ -251,13 +251,25 @@ test-race: TEST_PACKAGES=$(PACKAGES_NOSIMULATION)
...
+	@echo "Unit tests"; \
+	for module in $(SUB_MODULES); do \
+		cd ${CURRENT_DIR}/$$module; \
+		go test -mod=readonly -json $(ARGS) $(TEST_PACKAGES) ./... | tparse; \
+	done
 else
 	#go test -mod=readonly $(ARGS) $(TEST_PACKAGES)
-	find . -name go.mod -execdir go test ./... \;
+	# find . -name go.mod -execdir go test ./... \;
```
aleem1314, line 267:
```suggestion
```

x/group/types_test.go
```diff
@@ -395,14 +395,14 @@ func TestGroupAccountInfo(t *testing.T) {
 			timeout:      proto.Duration{Seconds: 1},
 			expErr:       true,
 		},
-		"invalid group account address": {
+		"valid variable length group account address": {
 			group:        1,
 			groupAccount: []byte("any-invalid-group-address"),
```
blushi, line 400: maybe change the string here too? (removing "invalid" which makes it a bit confusing now)

x/group/types_test.go
```diff
@@ -420,14 +420,14 @@ func TestGroupAccountInfo(t *testing.T) {
 			timeout:      proto.Duration{Seconds: 1},
 			expErr:       true,
 		},
-		"invalid admin account address": {
+		"valid variable length admin account address": {
 			group:        1,
 			groupAccount: []byte("valid--group-address"),
 			admin:        []byte("any-invalid-admin-address"),
```
blushi, line 426: ditto

x/group/msgs_test.go
```diff
@@ -146,13 +146,6 @@ func TestMsgCreateGroupAccount(t *testing.T) {
 			timeout:   proto.Duration{Seconds: 1},
 			expErr:    true,
 		},
-		"valid admin required": {
-			admin:     []byte("invalid-address"),
```
amaurym, line 150: why is this test deleted?

amaurym (approved):

app/modules_test.go
```diff
@@ -1,4 +1,5 @@
 // +build experimental
```
blushi, line 1: This is more coming from https://github.com/regen-network/regen-ledger/pull/322 but this means now that the `testutil/network` tests are now only running in experimental mode, which weren't the case and doesn't really make sense actually, should we have a separate test file for this instead?

Makefile
```diff
@@ -251,13 +251,23 @@ test-race: TEST_PACKAGES=$(PACKAGES_NOSIMULATION)
 
 $(TEST_TARGETS): run-tests
 
+SUB_MODULES = $(shell find . -type f -name 'go.mod' -printf '%h\n' | sort)
```
blushi, line 254: this is causing some issues on mac osx `find: -printf: unknown primary or operator`

aleem1314: Updated. can u test this again?

x/group/msgs_test.go
aleem1314: Now we have a variable-length address, this test never fails.

blushi (approved):
LGTM, thanks @aleem1314!

cyberbono3 (approved):
LGTM. Thanks for doing that, @aleem1314 !

Issue #360, opened 2021-05-26
Delete an account's eco-credit balance and supply when it goes to zero


robert-zaremba:
## Is your feature request related to a problem? Please describe.

To assure blockchain resources economics we shall follow same practices as in x/bank and remove _zero records_:
* zero eco-credit balances
* zero eco_credit supply

PR #367, opened 2021-05-28
feat: Delete eco-credit balance and supply when it goes to zero

likhita-809:
ref: #360


x/ecocredit/server/helpers.go
```diff
@@ -56,13 +55,18 @@ func getSubAndSetDecimal(store sdk.KVStore, key []byte, x *apd.Decimal) error {
 		return err
 	}
 
-	err = math.SafeSub(value, value, x)
-	if err != nil {
-		return err
+	if value.IsZero() {
```
blushi, line 58: shouldn't we check for zero value once we have substracted `x` from `value`?

likhita-809: sorry, my bad. I'll fix it

likhita-809: sorry, my bad. I'll fix it

blushi (commented):
should we make this ready for review?

likhita-809:
> should we make this ready for review?

atheeshp:
looks good @likhita-809 , can you also check for zero values [here](https://github.com/regen-network/regen-ledger/blob/987fe7acea659cd491f69dcf88c5fbc675e835fe/x/ecocredit/server/msg_server.go#L119-L120).

atheeshp:
> looks good @likhita-809 , can you also check for zero values [here](https://github.com/regen-network/regen-ledger/blob/987fe7acea659cd491f69dcf88c5fbc675e835fe/x/ecocredit/server/msg_server.go#L119-L120), while creating the batch.

instead we can revert all these changes and add a check for zero value in `setDecimal` ([here](https://github.com/regen-network/regen-ledger/blob/987fe7acea659cd491f69dcf88c5fbc675e835fe/x/ecocredit/server/helpers.go#L29-L35)) func and delete the key if matches zero. (following the same in sdk x/bank).

atheeshp (approved):

blushi (approved):

Issue #328, opened 2021-04-16
Record retirement location for ecocredit module


clevinson:
Retirement location should be stored & recorded. This should be a location of:
- beneficiary
- buyer

Paris Agreements require to track where we require credits (jurisdiction / sub national, eg: NY State can have different accounting than New Jearsy or New York City).
We need short, not ambiguous field for location. 
We propose:
```
<country-code>[-<sub-national>-[<postal-code>]]
```
The former two are codes specified by [SO_3166-2](https://en.wikipedia.org/wiki/ISO_3166-2)

clevinson:
Some refinement here:
- issuance location (geopolygon) will be taken care of via shacl schemas on data module, and are out of scope for this issue, though still a requirement

Adding one field on retirement for location, it should be strings adhering to something similar to: https://en.wikipedia.org/wiki/ISO_3166-2

ruhatch:
A couple of refinement questions here:

**1. Should we record the location in the store?**

We are going to log the location in events, so we can access location information using an index of events, but that might make quick access more difficult. In general, we'd like to store as little as possible in state, so to stick with that design principle we would err on the side of not putting locations in the store. If it's really bad for end-user experience, then we can add them into the store, and we can always do this at a later date, as it doesn't affect the actual chain.

**2. Should we record locations for retired_units in MsgCreateBatch and MsgSend?**

Both these messages include credit retirement during the issuance or transfer of credits from a batch. It seems we therefore need to include retirement location alongside those credit retirements, but I want to check that this is desired behaviour. @clevinson @blushi would you be able to check this with Ron?

PR #384, opened 2021-06-15
feat!: Record retirement location for ecocredit module (#328)

ruhatch:
## Description

Closes: #328 

This PR introduces retirement locations for ecocredits. The retirement location is the location of the beneficiary or buyer of the retired credits. It is a string of the form `<country-code>[-<sub-national-code>[-<postal-code>]]`, with the first two fields conforming to the ISO 3166-2 standard. The location is emitted during retirement events and can therefore be tracked by an indexer.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -66,4 +66,10 @@ message EventRetire {
 
   // units is the decimal number of credits that have been retired.
   string units = 3;
+
+  // location is the location of the beneficiary or buyer of the retiring
+  // credits. It is a string of the form
+  // <country-code>[-<sub-national-code>[-<postal-code>]], with the first two fields
+  // conforming to ISO 3166-2
```
amaurym, line 73:
```suggestion
  // conforming to ISO 3166-2.
```

x/ecocredit/server/msg_server.go
```diff
@@ -2,12 +2,14 @@ package server
...
 import (
 	"fmt"
+	"strings"
 
 	"github.com/regen-network/regen-ledger/types"
 
 	"github.com/cockroachdb/apd/v2"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/TheBookPeople/iso3166"
```
amaurym, line 12: this is GPLv3, I don't think we can use it here unfortunately

x/ecocredit/server/testsuite/suite.go
```diff
@@ -157,25 +175,47 @@ func (s *IntegrationTestSuite) TestScenario() {
...
-			name:      "can't use more than 7 decimal places",
-			toRetire:  "10.00000001",
-			expectErr: true,
+			name:               "can't use more than 7 decimal places",
+			toRetire:           "10.00000001",
+			retirementLocation: "AF",
+			expectErr:          true,
+		},
+		{
+			name:               "can't retire to an invalid country",
```
amaurym, line 198: we could also add one test case with empty retirementLocation

amaurym (commented):
Nice work, lgtm overall. I'm just not 100% certain we can use the github.com/TheBookPeople/iso3166 library

x/ecocredit/server/msg_server.go
ruhatch: Oh...that's annoying! Good spot!

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -66,4 +66,10 @@ message EventRetire {
 
   // units is the decimal number of credits that have been retired.
   string units = 3;
+
+  // location is the location of the beneficiary or buyer of the retiring
```
blushi, line 70:
```suggestion
  // location is the location of the beneficiary or buyer of the retired
```

x/ecocredit/server/msg_server.go
```diff
@@ -345,5 +367,50 @@ func retire(ctx types.Context, store sdk.KVStore, recipient string, batchDenom b
...
+	strings := strings.Split(location, "-")
+	switch numStrings := len(strings); numStrings {
+	case 1:
+		return strings[0], "", "", nil
+	case 2:
+		return strings[0], strings[1], "", nil
+	case 3:
+		return strings[0], strings[1], strings[2], nil
+	default:
+		return "", "", "", fmt.Errorf("Location should have format <country-code>[-<region-code>[-<postal-code>]]")
```
blushi, line 403: maybe we should wrap the error (for instance with `sdkerrors.ErrInvalidRequest`?)

x/ecocredit/server/msg_server.go
```diff
@@ -345,5 +367,50 @@ func retire(ctx types.Context, store sdk.KVStore, recipient string, batchDenom b
...
+	}
+}
+
+// validatePostalCode currently checks that the postal code is not longer than
+// 10 characters, as currently the longest postal codes in the world are 10
+// characters
+func validatePostalCode(postalCode string) error {
+	if len(postalCode) > 10 {
+		return fmt.Errorf("The postal code must not be longer than 10 characters")
```
blushi, line 412: ditto

x/ecocredit/server/msg_server.go
```diff
@@ -345,5 +367,50 @@ func retire(ctx types.Context, store sdk.KVStore, recipient string, batchDenom b
 		Retirer:    recipient,
 		BatchDenom: string(batchDenom),
 		Units:      math.DecimalString(retired),
+		Location:   location,
 	})
 }
+
+// validateLocation checks that the country and region conform to ISO 3166 and
```
blushi, line 374: what about extracting all the location validation related functions into a separate file? and then have a dedicated test file for it, eg location_test.go instead of msg_server_test.go

x/ecocredit/server/msg_server.go
blushi: very unfortunate indeed, there's also https://github.com/pariz/gountries (MIT) but doesn't seem like it's fully complete yet...
thoughts @aaronc? do you believe we should be more simplistic and just use regexp instead (see https://www.iso.org/iso-3166-country-codes.html  "WHAT IS INCLUDED IN ISO 3166?" section) without actually checking if a code really exists?

blushi (commented):
Looks good overall, see small comments and follow-up on the ISO 3166 validation.

x/ecocredit/server/msg_server.go
ruhatch: It looks like `gountries` does have enough to make it worthwhile - it's got all the data and we can just construct our own search on the data structure it hands us.

x/ecocredit/server/location.go
```diff
@@ -0,0 +1,63 @@
...
+		return err
+	}
+
+	iso3166query := gountries.New()
+
+	country, err := iso3166query.FindCountryByAlpha(countryCode)
+	if err != nil {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
+	}
+
```
ruhatch, line 24:
```suggestion
```

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -136,11 +144,19 @@ message MsgSendRequest {
...
-    // retired_units are the units of credits in this issuance that are
-    // effectively retired by the issuer on receipt. Decimal values are
+    // retired_units are the units of credits in this transfer that are
+    // effectively retired by the recipient on receipt. Decimal values are
     // acceptable within the precision returned by Query/Precision.
     string retired_units = 3
         [ (gogoproto.moretags) = "yaml:\"retired_units\"" ];
+
+    // retirement_location is the location of the beneficiary or buyer of the
+    // retired credits. This must be provided if retired_units is positive. It
```
aaronc, line 154: Is it actually required? @blushi or @clevinson can you check with Ron?

x/ecocredit/server/location_test.go
```diff
@@ -0,0 +1,79 @@
...
+			expectErr: false,
+		},
+		{
+			name:      "can't validate invalid region code",
+			location:  "BF-ZZ",
+			expectErr: true,
+		},
+		{
+			name:      "can validate valid postal code",
+			location:  "BF-BAL-1",
```
aaronc, line 52: I would say we want a space instead of a `-` before the postal code. What do you think?

x/ecocredit/server/location.go
```diff
@@ -0,0 +1,62 @@
...
+// the postal code is valid
+func validateLocation(location string) error {
+	countryCode, regionCode, postalCode, err := parseLocation(location)
+	if err != nil {
+		return err
+	}
+
+	iso3166query := gountries.New()
+
+	country, err := iso3166query.FindCountryByAlpha(countryCode)
```
aaronc, line 20: So I'd rather not check the list of countries and sub regions. This can change with little notice and we don't want to require a hard fork to update it. So for now I'd say we should just use a regex to validate 2 alphabetic characters for country and 1-3 alphanumeric characters for subregion. That's more future proof and doesn't introduce a possibly buggy third party dependency

x/ecocredit/server/location.go
```diff
@@ -0,0 +1,62 @@
...
+	default:
+		return "", "", "", sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "Location should have format <country-code>[-<region-code>[-<postal-code>]]")
+	}
+}
+
+// validatePostalCode currently checks that the postal code is not longer than
+// 10 characters, as currently the longest postal codes in the world are 10
+// characters
+func validatePostalCode(postalCode string) error {
+	if len(postalCode) > 10 {
```
aaronc, line 57: This seems a little too stringent. Maybe we just validate that it's alphanumeric plus maybe dashes and not longer than 16 or 20 characters maybe?

x/ecocredit/server/msg_server.go
aaronc: Just regex please as I commented elsewhere

aaronc: In the future we can have a dataset managed with x/params maybe but we do not want to force a hard fork

x/ecocredit/server/location_test.go
ruhatch: Not sure! Why would you prefer one to the other?

x/ecocredit/server/location.go
ruhatch: Yeah, makes sense

x/ecocredit/server/location.go
ruhatch: Sure. I've just done an audit of global postcode formats and we'll want alphanumeric, plus dash, parentheses, and spaces. Some postal codes include the name of a city, so actually we'd probably want more like 30 characters. Here's a long example from Gabon: `01 LIBREVILLE–GROS-BOUQUET 03`.

x/ecocredit/server/location_test.go
aaronc: Hmm just seems more visually clear

x/ecocredit/server/location.go
aaronc: Why not be generous and say 32 or 64?

x/ecocredit/server/location_test.go
ruhatch: Sure, I can just work the whole thing into a single regex then

ruhatch:
Also need to add `client/tx` for locations

aaronc (commented):
Let's also add the location validation logic to the `Msg.ValidateBasic` methods

proto/regen/ecocredit/v1alpha1/tx.proto
blushi: why wouldn't it be required in this case? As long as credits get retired, even if it's on issuance, the location should be recorded, even though this use case might not happen that often.

x/ecocredit/location_test.go
```diff
@@ -0,0 +1,79 @@
...
+			expectErr: true,
+		},
+		{
+			name:      "can validate valid region code",
+			location:  "AF-BDS",
+			expectErr: false,
+		},
+		{
+			name:      "can validate valid region code 2",
+			location:  "BF-BAL",
```
aaronc, line 42: Let's check numeric region codes too

x/ecocredit/location_test.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func TestValidateLocation(t *testing.T) {
+	specs := []struct {
+		name      string
+		location  string
+		expectErr bool
+	}{
+		{
+			name:      "can validate valid country code",
```
aaronc, line 16:
```suggestion
			name:      "valid country code",
```
we can drop the "can/can't validate" for better readability

x/ecocredit/location_test.go
```diff
@@ -0,0 +1,79 @@
...
+			location:  "AF",
+			expectErr: false,
+		},
+		{
+			name:      "can validate valid country code 2",
+			location:  "BF",
+			expectErr: false,
+		},
+		{
+			name:      "can't validate invalid country code",
```
aaronc, line 26:
```suggestion
			name:      "invalid country code",
```

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -136,11 +144,19 @@ message MsgSendRequest {
...
+    // is a string of the form
+    // <country-code>[-<sub-national-code>[-<postal-code>]], with the first two
```
aaronc, line 156: This needs to be updated to describe the space character and postal code format (64 char alphanumeric)

x/ecocredit/msgs.go
```diff
@@ -60,10 +67,17 @@ func (m *MsgSendRequest) ValidateBasic() error {
...
 		}
 
-		_, err = math.ParseNonNegativeDecimal(iss.RetiredUnits)
+		retiredUnits, err := math.ParseNonNegativeDecimal(iss.RetiredUnits)
 		if err != nil {
 			return err
 		}
+
+		if !retiredUnits.IsZero() {
+			err = validateLocation(iss.RetirementLocation)
```
aaronc, line 76: @blushi @clevinson do we always require a valid location? Could someone retire credits without providing a location theoretically?

aaronc (approved):
A few small comments, but pre-approving. Thanks @ruhatch !

x/ecocredit/msgs.go
blushi: I don't think so, this should be required as long as credits get retired

ruhatch: @blushi is it worth a final check with Ron?

amaurym (approved):
nice! lgtm

PR #389, opened 2021-06-17
feat!: ecocredit genesis import and export

aleem1314:
## Description

Closes: #187 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1alpha1/genesis.proto
```diff
@@ -0,0 +1,72 @@
...
+
+option go_package = "github.com/regen-network/regen-ledger/x/ecocredit";
+
+// GenesisState defines ecocredit module's genesis state.
+message GenesisState {
+  // Params contains the updateable global parameters for use with the x/params
+  // module
+  Params params = 1 [ (gogoproto.nullable) = false ];
+
+  // class_infos is the list of credit class info.
```
blushi, line 15:
```suggestion
  // class_info is the list of credit class info.
```
information has no plural form, ditto for the fields/comments below

proto/regen/ecocredit/v1alpha1/genesis.proto
```diff
@@ -0,0 +1,72 @@
...
+  // used to represent some quantity of credit units. It is an experimental
+  // feature to concretely explore an idea proposed in
+  // https://github.com/cosmos/cosmos-sdk/issues/7113.
+  uint32 max_decimal_places = 3;
+}
+
+// Balance defines credit batch tradable/retired balance.
+message Balance {
+  // reciepient is the address of the account receiving credits.
+  string reciepient = 1;
```
blushi, line 56: why not just using `address` instead (as in x/bank genesis)?
```suggestion
  // address is the account address of the account holding credits.
  string address = 1;
```

blushi (commented):
I haven't looked at the implementation yet, just some comments regarding the proto definitions

x/ecocredit/server/testsuite/suite.go
```diff
@@ -58,11 +66,11 @@ func (s *IntegrationTestSuite) SetupSuite() {
 	s.paramsQueryClient = params.NewQueryClient(s.fixture.QueryConn())
 }
 
-func fundAccount(bankKeeper bankkeeper.Keeper, ctx sdk.Context, addr sdk.AccAddress, amounts sdk.Coins) error {
-	if err := bankKeeper.MintCoins(ctx, minttypes.ModuleName, amounts); err != nil {
+func (s *IntegrationTestSuite) fundAccount(ctx sdk.Context, addr sdk.AccAddress, amounts sdk.Coins) error {
```
ruhatch, line 69: I guess we can also remove the `ctx` parameter, as that is using `s.sdkCtx`?

ruhatch (approved):
This looks good to me. We might want some more eyes on it though, as there are a lot of bits and I'm not very experienced with Go yet!

x/ecocredit/server/genesis.go
```diff
@@ -1,32 +1,177 @@
...
+
+	for _, precision := range genesisState.Precisions {
+		key := MaxDecimalPlacesKey(batchDenomT(precision.BatchDenom))
+		setUInt32(store, key, precision.MaxDecimalPlaces)
+	}
+
 	return []abci.ValidatorUpdate{}, nil
 }
 
+// ExportGenesis will dump the contents of the keeper into a serializable GenesisState.
```
blushi, line 90: strictly speaking, there's no keeper in x/ecocredit so maybe we should upgrade the language here

blushi (commented):
Looks good overall, just some refactoring requests

robert-zaremba (commented):
Proper go doc comments are missing

blushi:
> Proper go doc comments are missing

@aleem1314 could you address that? I believe it's mainly in data_prefixes.go

x/ecocredit/go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	github.com/pkg/errors v0.9.1
+	github.com/pariz/gountries v0.0.0-20200430155801-1c6a393df9c7
```
aaronc, line 14: Why is this needed?

aleem1314: Not using anywhere, removing it.

ruhatch: This was from the retirement location stuff, but should have been removed - perhaps this snuck in a merge/rebase at some point, then disappeared after squash/merge.

x/ecocredit/server/genesis.go
```diff
@@ -1,31 +1,189 @@
...
+		return TradableSupplyKey(bd)
+	}); err != nil {
+		return nil, err
+	}
+
+	if err := setSupply(store, genesisState.RetiredSupplies, func(bd batchDenomT) []byte {
+		return RetiredSupplyKey(bd)
+	}); err != nil {
+		return nil, err
+	}
```
aaronc, line 64: Supplies should only be used as an invariant check. Setting balances should update the supply and then we should check that the supply is what is expected. We should never just set supply based on a value in genesis.

aaronc (changes requested):

x/ecocredit/server/invariants.go
```diff
@@ -0,0 +1,146 @@
...
+func retiredSupplyInvariant(store types.KVStore) (string, bool) {
+	var (
+		msg    string
+		broken bool
+	)
+	calRetiredSupplies := make(map[string]*apd.Decimal)
+	if err := iterateBalances(store, TradableBalancePrefix, func(denom string, balance *apd.Decimal) bool {
+		if supply, ok := calRetiredSupplies[denom]; ok {
+			if err := math.Add(supply, balance, supply); err != nil {
+				msg += fmt.Sprintf("error adding credit batch retired supply %v", err)
```
anilCSE, line 82: Don't we  need to set `broken` here?

anilCSE (commented):
lgtm, just one question

blushi (commented):
Could you add some tests for the invariants? (as done in x/group)

blushi (approved):
LGTM!

ruhatch:
@aleem1314 looks like you need to `make proto-gen` and then this is good to go :)

atheeshp (approved):
lgtm

blushi:
@aleem1314 is this good to merge?

aleem1314:
@aaronc could you have a final look?

proto/regen/ecocredit/v1alpha1/genesis.proto
```diff
@@ -0,0 +1,72 @@
...
+  repeated Balance tradable_balances = 5;
+
+  // retired_balances is the list of credit batch retired units.
+  repeated Balance retired_balances = 6;
+
+  // tradable_supplies is the list of credit batch tradable supply.
+  repeated Supply tradable_supplies = 7;
+
+  // retired_supplies is the list of credit batch retired supply.
+  repeated Supply retired_supplies = 8;
```
aaronc, line 34:
```suggestion
  repeated Balance balances = 5;

  // tradable_supplies is the list of credit batch tradable supply.
  repeated Supply supplies = 6;

```

aaronc (approved):
Logic looks correct.

Can we simplify this API? Won't block on it because I know this has been in progress for a while, so pre-approving. But I think we can merge tradable and retired balances and supplies.

proto/regen/ecocredit/v1alpha1/genesis.proto
ruhatch: Would this mean adding an enum to the `Balance` type to keep track of if it's retired or tradable?

aleem1314: updated

clevinson: I just spoke with @aaronc about this and actually the intention of his comment was to still have separate fields for `retired_balance` and `tradable_balance`, but not have them be entirely separate `Balance` / `Supply` messages (for tradable / retired).

I think enums over complicate things, and we don't want/need to be tracking any balances outside of those that are retired or tradable (e.g. cancelled credits should never be tracked in genesis import/export).

So this would look something more like:
```proto
message Balance {
   // address is the account address of the account holding credits.
   string address = 1;

   // batch_denom is the unique ID of the credit batch.
   string batch_denom = 2;

   // tradable_balance is the tradable balance of the credit batch.
   string tradable_balance = 3;

   // retired_balance is the retired balance of the credit batch.
   string retired_balance = 3;
 }
 ```

File: proto/regen/ecocredit/v1alpha1/genesis.proto [hunk 9]
```diff
@@ -0,0 +1,80 @@
...
+  // Type is the credit balance type.
+  enum Type {
+    // TYPE_UNSPECIFIED is the unknown credit balance type.
+    TYPE_UNSPECIFIED = 0;
+    // TYPE_TRADABLE is the tradable balance type.
+    TYPE_TRADABLE = 1;
+    // TYPE_RETIRED is the retired balance type.
+    TYPE_RETIRED = 2;
+  }
+}
```
clevinson, line 71: I think adding an enum actually overcomplicates things.  @aleem1314 can we change this to something like:
```
message Balance {
   // address is the account address of the account holding credits.
   string address = 1;

   // batch_denom is the unique ID of the credit batch.
   string batch_denom = 2;

   // tradable_balance is the tradable balance of the credit batch.
   string tradable_balance = 3;

   // retired_balance is the retired balance of the credit batch.
   string retired_balance = 3;
 }
 ```
 
 I spoke with @aaronc today and confirmed that this was his original idea as well when he suggested combining into a singular `Balance` message.

File: proto/regen/ecocredit/v1alpha1/genesis.proto [hunk 10]
```diff
@@ -0,0 +1,80 @@
...
+
+// Supply represents a tradable or retired supply of a credit batch.
+message Supply {
+  // batch_denom is the unique ID of the credit batch.
+  string batch_denom = 1;
+
+  // supply is the tradable or retired supply of the credit batch.
+  string supply = 2;
+}
```
clevinson, line 80: Can you update this to the same format as above (e.g. with separate `tradable_balance` and `retired_balance` fields in the `Supply` message?)

clevinson (changes requested):
Few comments on the formatting of the Supply / Balance messages. Can you take a look at this tomorrow @aleem1314 and then I think we're good to merge!

proto/regen/ecocredit/v1alpha1/genesis.proto
robert-zaremba: retired balance ID  should be `4`,  right?

proto/regen/ecocredit/v1alpha1/genesis.proto
```diff
@@ -0,0 +1,57 @@
+syntax = "proto3";
+package regen.ecocredit.v1alpha1;
+
+import "regen/ecocredit/v1alpha1/types.proto";
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/regen-network/regen-ledger/x/ecocredit";
```
robert-zaremba, line 7: I was posting similar comment to NFT: how about putting genesis into a separate package, eg `x/ecocredit/genesis` - it has different domain than RPC messages. Or even we can have two packages:
* x/ecocredit/  -- for all things related to RPC
* x/ecocredit/DAL -- for all things related to storage (data access layer)

clevinson: I agree with this separation in principle, but don't want to block this PR on it. adding a new versioning scheme outside of the one for just modules (which is the current setup) is something that I'd like to make sure is the direction we want to go in across the SDK before doing it just on our own modules.

We will already need to be doing a version upgrade for ecocredit module  proto files in regen ledger v3 as we expect the API to change possibly quite a bit. So seems fine IMO to not block the v2.0 upgrade on this question.

clevinson (approved):
Few merge conflicts now, but pending those getting fixed, I think this is good to go.

There's some stuff where we may want to add some additional validation to `InitGenesis`, namely that all the `CreditType` values in any given genesis credit classes actually correspond to a CreditType in the ecocredit's genesis params.

Happy to move that to a separate issue and just continue a discussion there to see what makes the most sense after this PR gets merged.

Thanks @aleem1314!

Issue #385, opened 2021-06-15
Add support for credit cancelling


clevinson:
We should add basic functionality in the ecocredit module whereby:
- the holder of credits can cancel their own credits (any number of credits that are unretired)

Down the road we may want to add functionality for issuers to cancel credits as well, but more requirements need to be assessed before this gets implemented.

aaronc:
This would probably look like adding `MsgCancel`:

```proto
message MsgCancel {
  string holder = 1;
  repeated Unit units = 2;

  message Units {
    string batch_denom = 1;
    string amount = 2;
  }
}
```

robert-zaremba:
Only token owner can cancel his tokens.

robert-zaremba:
When cancelling we:
* reduce supply 
* subtract /remove balance

Issue #150, opened 2020-11-05
Add integration tests


robert-zaremba:
- [ ] Add CLI integration tests

ruhatch:
@clevinson @aaronc does this belong in the v1.1 - Baikal Upgrade milestone?

ruhatch:
@blushi @amaurym Do we need to add integration tests other than CLI? gRPC perhaps?

PR #391, opened 2021-06-17
feat!: Add support for credit cancelling (#385)

ruhatch:
## Description

This PR adds support for cancelling ecocredits. An account holder can cancel credits that they own from any credit batch. Cancelled credits are removed from the holders tradeable balance and from the tradeable supply. Cancelling credits has no effect on the retirement balances and supplies. After cancellation it is as if the credit was never issued, except that it is still accounted for in the `total_amount` in the `BatchInfo` for the batch.

Closes: #385

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -67,3 +67,19 @@ message EventRetire {
...
+
+  // canceller is the account which has cancelled the credits, which should be
+  // the holder of the credits.
+  string canceller = 1;
+
+  // batch_denom is the unique ID of credit batch.
+  string batch_denom = 2 [ (gogoproto.moretags) = "yaml:\"batch_denom\"" ];
+
+  // units is the decimal number of credits that have been retired.
+  string units = 3;
```
aaronc, line 84: Should we say `units` or `amount`?

aaronc: I think I introduced the language `units` but it seems weird to me now

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -172,6 +177,31 @@ message MsgRetireRequest {
...
+  // CancelUnits are the units of the batch being cancelled.
+  message CancelUnits {
+
+    // batch_denom is the unique ID of the credit batch.
+    string batch_denom = 1 [ (gogoproto.moretags) = "yaml:\"batch_denom\"" ];
+
+    // units are the units of credits being cancelled.
+    // Decimal values are acceptable within the precision returned by
+    // Query/Precision.
+    string units = 2;
```
aaronc, line 198: `amount`? we would also need to change other fields named `units` if we do this renaming

x/ecocredit/client/tx.go
```diff
@@ -172,6 +173,34 @@ Parameters:
...
 }
 
+func txCancel() *cobra.Command {
+	return &cobra.Command{
+		Use:   "cancel [credits]",
+		Short: "Cancels a specified amounts of credits from the account of the transaction author (--from)",
+		Long: `Cancels a specified amounts of credits from the account of the transaction author (--from)
+
+Parameters:
+  credits:  YAML encoded credit list. Note: numerical values must be written in strings.
```
aaronc, line 183: Let's not do YAML in the CLI. Can we think of some other way to do this?

aaronc (commented):
LGTM other than YAML in the CLI and my small thought about naming. What alternatives do we have to YAML? Could we do something like coins does (ex `100atom,1000regen`)?

proto/regen/ecocredit/v1alpha1/events.proto
ruhatch: Yeah, I agree! I was aiming for consistency here.

proto/regen/ecocredit/v1alpha1/tx.proto
ruhatch: Cool, I guess that's a good change to make now, as we've not launch the module yet

x/ecocredit/client/tx.go
ruhatch: Yeah, I'll have a think about something else. The YAML was awkward during manual testing.

ruhatch: It's slightly harder than with coins as batch denoms look like `18AV53K/1b1vrHF`, possibly having digits in front. We could do something like `10:18AV53K/1b1vrHF,100:18AV53K/1b1vrHF`?

ruhatch: There's also YAML in the `create-batch` command. Would you prefer to remove that as well?

ruhatch:
@aaronc @blushi I found this as well:
```
// BatchInfo represents the high-level on-chain information for a credit batch.
message BatchInfo {

  // class_id is the unique ID of credit class.
  string class_id = 1 [ (gogoproto.moretags) = "yaml:\"class_id\"" ];

  // batch_denom is the unique ID of credit batch.
  string batch_denom = 2 [ (gogoproto.moretags) = "yaml:\"batch_denom\"" ];

  // issuer is the issuer of the credit batch.
  string issuer = 3;

  // total_amount is the total number of credits in the credit batch and is
  // immutable.
  string total_amount = 4 [ (gogoproto.moretags) = "yaml:\"total_amount\"" ];

  // metadata is any arbitrary metadata to attached to the credit batch.
  bytes metadata = 5;
}
```

It states in the comment of `total_amount` (which I just renamed), that this field is immutable. Does this change with the new cancellation logic and do we need to update the batch info during cancellation? Can you ask Ron if you're unsure?

ruhatch:
> LGTM other than YAML in the CLI and my small thought about naming. What alternatives do we have to YAML? Could we do something like coins does (ex `100atom,1000regen`)?

@aaronc I've implemented with the form `100:ABC/123,256.974:123/XYZ`, so please review and let me know what you think

aaronc:
> > LGTM other than YAML in the CLI and my small thought about naming. What alternatives do we have to YAML? Could we do something like coins does (ex `100atom,1000regen`)?
> 
> 
> 
> @aaronc I've implemented with the form `100:ABC/123,256.974:123/XYZ`, so please review and let me know what you think. It should be easy to adjust that format if you think we can do something better there.

That sounds good. Separately we should find a way for credit batch names to always start with a letter so a colon isn't needed. I propose `eco/` as a prefix for all credit classes and batches as that's what we'll be doing when we integrate with bank. What do you think?

ruhatch:
> That sounds good. Separately we should find a way for credit batch names to always start with a letter so a colon isn't needed. I propose `eco/` as a prefix for all credit classes and batches as that's what we'll be doing when we integrate with bank. What do you think?

Yeah, I think that seems reasonable and makes it interoperable with `Coin` types. I'll open a separate issue for that. Is that something we'll want to get into v1.1?

aaronc:
> > That sounds good. Separately we should find a way for credit batch names to always start with a letter so a colon isn't needed. I propose `eco/` as a prefix for all credit classes and batches as that's what we'll be doing when we integrate with bank. What do you think?
> 
> 
> 
> Yeah, I think that seems reasonable and makes it interoperable with `Coin` types. I'll open a separate issue for that. Is that something we'll want to get into v1.1?
> 
> 

Yes 👍

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -64,6 +64,22 @@ message EventRetire {
...
+  // canceller is the account which has cancelled the credits, which should be
+  // the holder of the credits.
+  string canceller = 1;
+
+  // batch_denom is the unique ID of credit batch.
+  string batch_denom = 2 [ (gogoproto.moretags) = "yaml:\"batch_denom\"" ];
+
+  // amount is the decimal number of credits that have been cancelled.
+  string amount = 3;
+}
```
likhita-809, line 85:
```suggestion
}

```

likhita-809 (commented):
looks neat, just a small nit

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -47,13 +47,13 @@ message EventReceive {
   // batch_denom is the unique ID of credit batch.
   string batch_denom = 3 [ (gogoproto.moretags) = "yaml:\"batch_denom\"" ];
 
-  // units is the decimal number of both tradable and retired credits received.
-  string units = 4;
+  // amonut is the decimal number of both tradable and retired credits received.
```
blushi, line 50:
```suggestion
  // amount is the decimal number of both tradable and retired credits received.
```

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -28,8 +28,8 @@ message EventCreateBatch {
   // issuer is the account address of the issuer of the credit batch.
   string issuer = 3;
 
-  // total_units is the total number of units in the credit batch.
```
blushi, line 31: why renaming to amount? to be more consistent with x/bank?

x/ecocredit/client/tx.go
```diff
@@ -172,6 +173,34 @@ Parameters:
 	}
 }
 
+func txCancel() *cobra.Command {
```
blushi, line 176: it doesn't have to be part of this PR but it would be good to add some integration tests for the ecocredit CLI for next release
I believe https://github.com/regen-network/regen-ledger/issues/150 should be added to v1.1 - Baikal Upgrade milestone, wdyt @clevinson @aaronc?

x/ecocredit/client/tx.go
```diff
@@ -172,6 +173,34 @@ Parameters:
...
+	return &cobra.Command{
+		Use:   "cancel [credits]",
+		Short: "Cancels a specified amounts of credits from the account of the transaction author (--from)",
+		Long: `Cancels a specified amounts of credits from the account of the transaction author (--from)
```
blushi, line 180:
```suggestion
		Short: "Cancels a specified amount of credits from the account of the transaction author (--from)",
		Long: `Cancels a specified amount of credits from the account of the transaction author (--from)
```

blushi (commented):
LGTM, just small comments

blushi:
> @aaronc @blushi I found this as well:
> 
> ```
> // BatchInfo represents the high-level on-chain information for a credit batch.
> message BatchInfo {
> 
>   // class_id is the unique ID of credit class.
>   string class_id = 1 [ (gogoproto.moretags) = "yaml:\"class_id\"" ];
> 
>   // batch_denom is the unique ID of credit batch.
>   string batch_denom = 2 [ (gogoproto.moretags) = "yaml:\"batch_denom\"" ];
> 
>   // issuer is the issuer of the credit batch.
>   string issuer = 3;
> 
>   // total_amount is the total number of credits in the credit batch and is
>   // immutable.
>   string total_amount = 4 [ (gogoproto.moretags) = "yaml:\"total_amount\"" ];
> 
>   // metadata is any arbitrary metadata to attached to the credit batch.
>   bytes metadata = 5;
> }
> ```
> 
> It states in the comment of `total_amount` (which I just renamed), that this field is immutable. Does this change with the new cancellation logic and do we need to update the batch info during cancellation? Can you ask Ron if you're unsure?

Not sure, will ask Ron

aaronc:
I think we should update the docs to mention cancellation.

proto/regen/ecocredit/v1alpha1/events.proto
ruhatch: Yes, and @aaronc suggested this would be a more suitable name

x/ecocredit/client/tx.go
ruhatch: Yeah, sounds good - how do we test the CLI currently?

ruhatch:
> I think we should update the docs to mention cancellation.

@aaronc And do we need to change the logic so that when we cancel credits, the `total_amount` changes in the batch info?

blushi:
> > I think we should update the docs to mention cancellation.
> 
> @aaronc And do we need to change the logic so that when we cancel credits, the `total_amount` changes in the batch info?

I asked Ron about that he responded:

> It might make sense to track the  “active”, “retired” and “canceled” and “initial issuance” separately. Where active = initial - canceled - retired.

We do already keep track of active and retired credits as part of the tradable and retired balances / supplies. Canceled credits can be indexed based on cancel events. So I believe it would make sense to keep the `total_amount` unchanged.

ruhatch:
> We do already keep track of active and retired credits as part of the tradable and retired balances / supplies. Canceled credits can be indexed based on cancel events. So I believe it would make sense to keep the `total_amount` unchanged.

Okay, sounds good! I'll adjust some docs that suggest cancel is as if they're never issued, because that's not the case here. Thanks for checking that!

aaronc:
I think language which distinguishes issued from active/valid at the batch level makes sense. I think leaving total amount as is could be confusing

x/ecocredit/client/tx.go
blushi: see for instance in `x/group/client/testsuite`

ruhatch:
> I think language which distinguishes issued from active/valid at the batch level makes sense. I think leaving total amount as is could be confusing

@aaronc hmm, okay, so perhaps renaming to `total_issued` or even `total_amount_issued`? Would that be enough?

aaronc:
> > I think language which distinguishes issued from active/valid at the batch level makes sense. I think leaving total amount as is could be confusing
> 
> @aaronc hmm, okay, so perhaps renaming to `total_issued` or even `total_amount_issued`? Would that be enough?

Sure, but then I would also add a `total_valid` field or something. Honestly, I think it would be cleaner to just change the `total_amount` to reflect the cancellation. My concern is that people will use these numbers in analytics and misinterpret the number of valid carbon credits.

ruhatch:
> Sure, but then I would also add a `total_valid` field or something. Honestly, I think it would be cleaner to just change the `total_amount` to reflect the cancellation. My concern is that people will use these numbers in analytics and misinterpret the number of valid carbon credits.

@aaronc Hmm, yeah, I had similar concerns. What about adjusting `total_amount` and adding something like `amount_cancelled`, so that someone can reconstruct the `total_issued`, but is unlikely to use the wrong values?

amaurym:
@aaronc Could you have a look at Ru's proposal above to get this PR move forward?

aaronc:
> > Sure, but then I would also add a `total_valid` field or something. Honestly, I think it would be cleaner to just change the `total_amount` to reflect the cancellation. My concern is that people will use these numbers in analytics and misinterpret the number of valid carbon credits.
> 
> @aaronc Hmm, yeah, I had similar concerns.
> 
> Ron's requirement above only needs us to track 3 of those 4 values, so we can decide which one to leave for reconstruction, or just accept the redundancy and store them all?
> 
> What about adjusting `total_amount` and adding something like `amount_cancelled`, so that someone can reconstruct the `total_issued`, but is unlikely to use the wrong values?

I think that makes sense and is likely least error prone.

ruhatch:
@aaronc could you give this another review with the `BatchInfo` updates? Hoping that it should be basically there now.

likhita-809 (approved):
lgtm

aaronc (approved):

PR #420, opened 2021-07-15
feat!: Add project location field to credit batches (#394)

ruhatch:
## Description

Add `project_location` field to `BatchInfo` using the same format as `retirement_location`s (ISO 3166-2 + postal code).

This PR also refactors the `create-batch` command line tool to accept a JSON file and add a `gen-batch-json` command to generate JSON.

Closes: #394

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -101,6 +101,13 @@ message MsgCreateBatchRequest {
   google.protobuf.Timestamp end_date = 6
       [(gogoproto.stdtime) = true, (gogoproto.moretags) = "yaml:\"end_date\""];
 
+  // project_location is the location of the project backing the credits in this
+  // batch. It is a string of the form
+  // <country-code>[-<sub-national-code>[ <postal-code>]], with the first two
```
aaronc, line 106: Should we add some text to clarify that sub national and postal codes are optional and can allow greater or lesser granularity?

x/ecocredit/client/tx.go
```diff
@@ -81,23 +81,27 @@ Parameters:
 
 func txCreateBatch() *cobra.Command {
 	return &cobra.Command{
-		Use:   "create-batch [issuer] [class_id] [start_date] [end_date] [metadata] [issuance]",
+		Use:   "create-batch [issuer] [class_id] [start_date] [end_date] [project_location] [metadata] [issuance]",
```
aaronc, line 84: Hmm... this has too many positional args now. We should probably make them all flags

aaronc (commented):
LGTM but maybe let's just improve the docs and cli UX a bit as I commented.

proto/regen/ecocredit/v1alpha1/tx.proto
ruhatch: Yup, good point

x/ecocredit/client/tx.go
```diff
@@ -79,40 +79,85 @@ Parameters:
...
+  issuance:         YAML encode issuance list.
+                    eg: '[{recipient: "xrn:sdgkjhs2345u79ghisodg", tradable_amount: "10", retired_amount: "2", retirement_location: "YY-ZZ 12345"}]'
+                    Note: numerical values must be written in strings.
+                    Note: "tradable_amount" and "retired_amount" default to 0.
+                    Note: "retirement_location" is only required when
+		          "retired_amount" is positive.`,
+		Args: cobra.ExactArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			startDate, err := parseDate("start_date", args[2])
+			issuer, err := cmd.Flags().GetString(FlagIssuer)
```
aaronc, line 116: I think issuer can just reuse the --from flag now...

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * add support for credit cancelling (#385)
 * record retirement locations of ecocredit (#328)
 * add dates as top level fields in credit batches (#393)
+* add project location as field in credit batches (#394)
```
robert-zaremba, line 13: Shall we have a sub-section (###) saying that all this added things are for x/ecocredit?

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -38,6 +38,14 @@ message EventCreateBatch {
   // end_date is the end of the period during which this credit batch was
   // quantified and verified.
   string end_date = 6 [ (gogoproto.moretags) = "yaml:\"end_date\"" ];
+
+  // project_location is the location of the project backing the credits in this
+  // batch. It is a string of the form
+  // <country-code>[-<sub-national-code>[ <postal-code>]], with the first two
+  // fields conforming to ISO 3166-2, and postal-code being up to 64
+  // alphanumeric characters. country-code is required, while sub-national-code
+  // and postal-code can be added for increasing precision.
```
robert-zaremba, line 47: Please document the `sub-national-code` format. Is there a limit of characters?

proto/regen/ecocredit/v1alpha1/events.proto
```diff
@@ -38,6 +38,14 @@ message EventCreateBatch {
   // end_date is the end of the period during which this credit batch was
   // quantified and verified.
   string end_date = 6 [ (gogoproto.moretags) = "yaml:\"end_date\"" ];
+
+  // project_location is the location of the project backing the credits in this
+  // batch. It is a string of the form
+  // <country-code>[-<sub-national-code>[ <postal-code>]], with the first two
```
robert-zaremba, line 44:
```suggestion
  // <country-code>[-<sub-national-code>][-<postal-code>], with the first two
```

robert-zaremba: 1. I  think we can have postal code without sub-national code.
2. shouldn't we have `-` before postal code as well?

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -101,6 +101,14 @@ message MsgCreateBatchRequest {
...
+  // fields conforming to ISO 3166-2, and postal-code being up to 64
+  // alphanumeric characters. country-code is required, while sub-national-code
+  // and postal-code can be added for increasing precision.
```
robert-zaremba, line 109: same comments as above. Maybe we should document it here, and in other places note that the documentation of the field is in `MsgCreateBatchRequest.project_location` to avoid copy pasting and desyncing?

x/ecocredit/client/tx.go
```diff
@@ -79,40 +79,85 @@ Parameters:
...
+	FlagProjectLocation string = "project-location"
+	FlagMetadata        string = "metadata"
+	FlagIssuance        string = "issuance"
+)
+
 func txCreateBatch() *cobra.Command {
-	return &cobra.Command{
-		Use:   "create-batch [issuer] [class_id] [start_date] [end_date] [metadata] [issuance]",
+	cmd := &cobra.Command{
+		Use:   "create-batch [--issuer issuer] [--class-id class_id] [--start-date start_date] [--end-date end_date] [--project-location project_location] [--metadata metadata] [--issuance issuance]",
```
robert-zaremba, line 94: I  think having a JSON parameter is better than all this command parameters, eg:
```
create-batch msg-create-batch
Parameters:
   msg-create-batch   a JSON object representing MsgCreateBatchRequest: %s  ...", json.Marshal(MsgCreateBatchRequest{})
```

With this, parsing we will simplify parameters parsing.

x/ecocredit/client/tx.go
robert-zaremba: Is it possible that issuer is not the msg signer?

robert-zaremba (commented):
Main suggestion (maybe to discuss in a broader group) is to use JSON object rather than multiple `--` parameters.

proto/regen/ecocredit/v1alpha1/events.proto
aaronc: It's documented in ISO 3166-2

proto/regen/ecocredit/v1alpha1/events.proto
aaronc: I think it's okay. We've already done many design cycles on this and I don't see a reason to change now

x/ecocredit/client/tx.go
aaronc: No

x/ecocredit/client/tx.go
aaronc: I had the same thought but then it's just basically the whole MsgIssueBatch json which is maybe okay

proto/regen/ecocredit/v1alpha1/events.proto
robert-zaremba: So, to confirm: based on the change in this PR: sub national code is required if one wants to put postal code. Correct?

x/ecocredit/client/tx.go
ruhatch: So, I agree this would probably be better for UX. I think it would be helpful to add some sort of util that give people a template JSON file or even lets them pass some of the values as command line params. Then they can manually inspect the JSON and then run:
```
regen tx ecocredit create-batch < batch.json
```

ruhatch: The util could take:
- Class ID string
- Number of issuances
- Metadata as a string
- Start date in yyyy-mm-dd format
- End date in yyyy-mm-dd format
- Project location string

This then creates a JSON file which has:
- Template issuances the user to fill in
- Base64 encoded the metadata
- Put the start date and end date into the full timestamp format for parsing

This basically takes care of all the tedious/tricky bits of getting the JSON file right.

Thoughts?

ruhatch:
@aaronc @robert-zaremba I've implemented JSON on the CLI for this one. Please can you take a look and see if you think it works and also check out the comment above about a CLI util to generate JSON for the user.

x/ecocredit/client/tx.go
```diff
@@ -79,59 +82,82 @@ Parameters:
...
+)
+
 func txCreateBatch() *cobra.Command {
-	return &cobra.Command{
-		Use:   "create-batch [issuer] [class_id] [start_date] [end_date] [metadata] [issuance]",
+	var (
+		startDate = time.Unix(10000, 10000).UTC()
+		endDate   = time.Unix(10000, 10050).UTC()
+	)
+	createBatchJSON, err := json.MarshalIndent(
```
aaronc, line 100:
```suggestion
	createBatchExampleJSON, err := json.MarshalIndent(
```

x/ecocredit/client/tx.go
```diff
@@ -79,59 +82,82 @@ Parameters:
 	}
 }
 
+const (
+	FlagIssuer          string = "issuer"
+	FlagClassId         string = "class-id"
+	FlagStartDate       string = "start-date"
+	FlagEndDate         string = "end-date"
+	FlagProjectLocation string = "project-location"
+	FlagMetadata        string = "metadata"
```
aaronc, line 91: Do we still need all these now?

aaronc (approved):
A few minor things. Otherwise LGTM

x/ecocredit/client/tx.go
aaronc: I think that makes sense. I don't think we need to spend a ton of time on it yet, but if it's simple sure. I'm sort of of the opinion the metadata should just be a binary hash and nothing more at this point. But maybe a string is fine. Also we would keep this under the `tx` command set

x/ecocredit/server/testsuite/suite.go
```diff
@@ -112,40 +112,66 @@ func (s *IntegrationTestSuite) TestScenario() {
...
+	s.Require().Error(err)
+
+	// Batch creation should fail if the ProjectLocation is invalid
+	err = (&ecocredit.MsgCreateBatchRequest{
+		Issuer:          issuer1,
+		ClassId:         clsID,
+		Issuance:        []*ecocredit.MsgCreateBatchRequest_BatchIssuance{},
+		StartDate:       &time1,
+		EndDate:         &time2,
+		ProjectLocation: "ABCD",
```
blushi, line 164: just a side note, all those `ValidateBasic` tests should be moved to some unit tests for the `ecocredit` package as part of https://github.com/regen-network/regen-ledger/issues/423

x/ecocredit/client/tx.go
```diff
@@ -79,59 +82,82 @@ Parameters:
...
+			ProjectLocation: "AB-CDE FG1 345",
+		},
+		"                    ",
+		"    ",
+	)
+	if err != nil {
+		panic("Couldn't marshal MsgCreateBatch to JSON")
+	}
+	cmd := &cobra.Command{
+		Use:   "create-batch msg-create-batch",
```
blushi, line 125:
```suggestion
		Use:   "create-batch [msg-create-batch-json-file]",
```

x/ecocredit/client/tx.go
```diff
@@ -79,59 +82,82 @@ Parameters:
...
-              quantified and verified. Format: yyyy-mm-dd.
-  metadata:   base64 encoded issuance metadata
-  issuance:   YAML encode issuance list. Note: numerical values must be written in strings.
-              eg: '[{recipient: "xrn:sdgkjhs2345u79ghisodg", tradable_amount: "10", retired_amount: "2", retirement_location: "YY-ZZ 12345"}]'
-              Note: "tradable_amount" and "retired_amount" default to 0.
-              Note: "retirement_location" is only required when "retired_amount" is positive.`,
-		Args: cobra.ExactArgs(6),
+  msg-create-batch: A JSON object representing MsgCreateBatch. This could be
+                    passed from a JSON file like:
+		    regen tx create-batch "$(< batch.json)"
```
blushi, line 132: why not just using `io/ioutil` instead (see for instance in x/group `parseMembers` function) and you would just need the following, which is more straightforward from a user standpoint and aligned with how it's done in the sdk tx commands: 
```suggestion
		    regen tx create-batch batch.json"
```

blushi (commented):
tACK, not approving yet since I think the CLI UX can be improved a bit

x/ecocredit/client/tx.go
ruhatch: @aaronc It was fairly quick to implement, so I've done it. Would you give it a quick look over?

blushi (approved):

Issue #424, opened 2021-07-20
Add parameter to track credit types and associated precision


ruhatch:
As part of integration with `x/bank`, we decided to have a single consistent precision for each type of credit, rather than being able to set precision per credit class. This issue should:

- [ ] Add a governance parameter to `x/ecocredit` containing a map of credit types and precisions
- [ ] Remove the ability to set precision on a per-class basis
- [ ] Add credit type as a field on credit class
- [ ] Enforce precisions of vaules based on the governance parameter
- [ ] Ensure that precision cannot be updated via governance (at least until bank has support for decimals)

ruhatch:
@aaronc how does this interact with #398?

aaronc:
Replaces #398

technicallyty:
was the intention to use the map in the proto files? might be an issue: the safety of using protobuf maps isn't quite well defined yet see:  https://github.com/cosmos/cosmos-sdk/pull/9073#discussion_r628474547

perhaps we do something like this using `repeated`: 

```proto
// Params defines the updatable global parameters of the ecocredit module for
// use with the x/params module.
message Params {
  // -- snip -- //

  // credit_types is a list of definitions for credit types
  repeated CreditType credit_types = 2;
}

message CreditType {
    string typeOf = 1;
    string unit = 2;
    uint32 precision = 3;
}
```

aaronc:
Let's use repeated 👍

technicallyty:
should we just rip out all `set precision` from the proto files and stuff? or is there a smoother way to depreciate it?

ruhatch:
> should we just rip out all `set precision` from the proto files and stuff? or is there a smoother way to depreciate it?

I think just rip it out! This isn't currently deployed, so we're changing the design to just have precision at the credit type level.

aaronc:
I would even goes as far as to say we might allow 6 as the only valid value for precision and then consider whether we loosen that in the future. What do you think @ruhatch?

technicallyty:
> I would even goes as far as to say we might allow 6 as the only valid value for precision and then consider whether we loosen that in the future. What do you think @ruhatch?

related q: how do we enforce credit types as a gov param while simultaneously not allowing the precision to change? the validation function is stateless so i can't grab the existing params and check for changes.. so is just enforcing a hard 6 the solution for now?

aaronc:
[...]

PR #426, opened 2021-07-21
feat: gov params for permissioned eco credits

technicallyty:
## Description

Closes: #425 

### summary
adds new gov params for permissioned credit creation

side effects of this pr: 
- fixed error in `app/regen/cmd/genaccounts.go` - couldn't compile due to error of cdc casting to itself
- `make proto-all` reformatted a lot of the files

### main changes

[add new params](https://github.com/regen-network/regen-ledger/compare/ty/425-permissioned_credits?expand=1#diff-3a1cfcf0e22efb6ebdffe109913260003fec3c681ebbb3f0c726b6f66f53ff37R75-R86)

[added new param keys](https://github.com/regen-network/regen-ledger/compare/ty/425-permissioned_credits?expand=1#diff-3a1cfcf0e22efb6ebdffe109913260003fec3c681ebbb3f0c726b6f66f53ff37R75-R86)

[add check in CreateCredit for the new allowlist](https://github.com/regen-network/regen-ledger/compare/ty/425-permissioned_credits?expand=1#diff-3a1cfcf0e22efb6ebdffe109913260003fec3c681ebbb3f0c726b6f66f53ff37R75-R86)

[added functions for checking these new params](https://github.com/regen-network/regen-ledger/compare/ty/425-permissioned_credits?expand=1#diff-3a1cfcf0e22efb6ebdffe109913260003fec3c681ebbb3f0c726b6f66f53ff37R75-R86)

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [x] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [x] added `!` to the type prefix if API or client breaking change
- [x] targeted the correct branch (see [PR Targeting](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))
- [x] provided a link to the relevant issue or specification
- [x] followed the guidelines for [building modules](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules)
- [x] included the necessary unit and integration [tests](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#testing)
- [x] added a changelog entry to `CHANGELOG.md`
- [x] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [x] updated the relevant documentation or specification
- [x] reviewed "Files changed" and left comments if necessary
[...]

PR #428, opened 2021-07-22
feat: credit types

technicallyty:
## Description

+Adds credit types which includes type name, unit type (kg, tonne, etc), and precision (currently locked @ 6).
+Adds query to get all credit types on CLI and grpc/http
+Updates CreateClass tx command to include credit class arg
+Adds tests for validating the credit type params
-removes per-class precision setting functionality

Closes: #424 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/params.go
```diff
@@ -37,12 +45,45 @@ func validateCreditClassFee(i interface{}) error {
 	return nil
 }
 
-func NewParams(creditClassFee sdk.Coins) Params {
+func validateCreditTypes(i interface{}) error {
```
technicallyty, line 48: should we enforce the units here? was wondering if we should have a hard coded list or something of valid unit types like kg, tonne, etc

go.sum
```diff
@@ -1234,8 +1234,8 @@ google.golang.org/genproto v0.0.0-20210310155132-4ce2db91004e/go.mod h1:FWY/as6D
 google.golang.org/genproto v0.0.0-20210319143718-93e7006c17a6/go.mod h1:FWY/as6DDZQgahTzZj3fqbO1CbirC29ZNUFHwi0/+no=
 google.golang.org/genproto v0.0.0-20210402141018-6c239bbf2bb1/go.mod h1:9lPAdzaEmUacj36I+k7YKbEc5CXzPIeORRgDAUOu28A=
 google.golang.org/genproto v0.0.0-20210602131652-f16073e35f0c/go.mod h1:UODoCrxHCcBojKKwX1terBiRUaqAsFqJiF615XL43r0=
-google.golang.org/genproto v0.0.0-20210721163202-f1cecdd8b78a h1:17YmRUuEF4d+t2ygJZaDPhqNL2Hf17832xWKcMU7r2I=
-google.golang.org/genproto v0.0.0-20210721163202-f1cecdd8b78a/go.mod h1:ob2IJxKrgPT52GcgX759i1sleT07tiKowYBGbczaW48=
+google.golang.org/genproto v0.0.0-20210722135532-667f2b7c528f h1:YORWxaStkWBnWgELOHTmDrqNlFXuVGEbhwbB5iK94bQ=
+google.golang.org/genproto v0.0.0-20210722135532-667f2b7c528f/go.mod h1:ob2IJxKrgPT52GcgX759i1sleT07tiKowYBGbczaW48=
```
technicallyty, line 1238: my go.sum/go.mod keep diverging from master when i gen proto, anyone know why this keeps happening?

proto/regen/ecocredit/v1alpha1/types.proto
```diff
@@ -73,10 +76,20 @@ message Params {
...
 // GenesisState defines the state of the ecocredit module that is needed at genesis
 message GenesisState {
   // Params contains the updateable global parameters for use with the x/params module
   Params params = 1 [(gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"params\""];
+}
+
+// CreditType defines the measurement unit/precision of a certain credit type (e.g. carbon, biodiversity...)
+message CreditType {
+  string type = 1;
+  string unit = 2;
```
aaronc: Let's add more docs, i.e. `type` should be something like `carbon` and `units` should be some SI unit like `tons`.
```suggestion
  string type = 1;
  string units = 2;
```

x/ecocredit/params.go
aaronc: not a bad idea, but then we'd need a units library... something like UCUM might work

aaronc: I found one golang UCUM library: https://github.com/BertVerhees/ucum. It's old and not maintained. We'd probably just want to download the XML file from https://ucum.org/trac and parse that.

@blushi have you thought much about units from the registry SHACL perspective?

A good alternative would be http://qudt.org which is more up to date, but requires us parsing RDF files...

aaronc: in short, probably a good idea but maybe we address later..?

proto/regen/ecocredit/v1alpha1/types.proto
technicallyty: i ninja'd some docs in right before this review comment 😏 

also changed unit -> units 👍🏻

x/ecocredit/params.go
technicallyty: ok ill open an issue for this

go.sum
ruhatch: Yeah, the `proto-gen` script runs `go get` on the `genproto` stuff. It's all good :)

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -34,11 +34,10 @@ service Query {
         "/regen/ecocredit/v1alpha1/batch_info/{batch_denom}/supply";
   }
 
-  // Precision queries the number of decimal places that can be used to
-  // represent credits in a batch. See Tx/SetPrecision for more details.
-  rpc Precision(QueryPrecisionRequest) returns (QueryPrecisionResponse) {
+  // CreditTypes queries the allow credit type information
```
ruhatch, line 37:
```suggestion
  // CreditTypes returns the list of allowed types that credit classes can have. 
  // See Types/CreditType for more details.
```

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -34,11 +34,10 @@ service Query {
...
+  rpc CreditTypes(QueryCreditTypesRequest) returns (QueryCreditTypesResponse) {
     option (google.api.http).get =
-        "/regen/ecocredit/v1alpha1/batches/{batch_denom}/precision";
+        "/regen/ecocredit/v1alpha1/credit_types";
```
ruhatch, line 40:
```suggestion
        "/regen/ecocredit/v1alpha1/credit-types";
```

Not sure if there's a convention here?

x/ecocredit/client/tx.go
```diff
@@ -49,15 +47,16 @@ func txflags(cmd *cobra.Command) *cobra.Command {
 
 func txCreateClass() *cobra.Command {
 	return &cobra.Command{
-		Use:   "create-class [designer] [issuer[,issuer]*] [metadata]",
+		Use:   "create-class [designer] [issuer[,issuer]*] [metadata] [credit type]",
```
ruhatch, line 50:
```suggestion
		Use:   "create-class [designer] [issuer[,issuer]*] [credit type] [metadata]",
```

Not sure why exactly, but I like to have `metadata` at the end, once the regular data is done

x/ecocredit/client/tx.go
```diff
@@ -49,15 +47,16 @@ func txflags(cmd *cobra.Command) *cobra.Command {
...
 
 Parameters:
-  designer:  address of the account which designed the credit class
-  issuer:    comma separated (no spaces) list of issuer account addresses. Example: "addr1,addr2"
-  metadata:  base64 encoded metadata - arbitrary data attached to the credit class info`,
-		Args: cobra.ExactArgs(3),
+  designer:  	address of the account which designed the credit class
+  issuer:    	comma separated (no spaces) list of issuer account addresses. Example: "addr1,addr2"
+  metadata:  	base64 encoded metadata - arbitrary data attached to the credit class info
+  credit type: 	the credit class type (e.g. carbon, biodiversity, etc)`,
```
ruhatch, line 58:
```suggestion
  designer:  	    address of the account which designed the credit class
  issuer:    	    comma separated (no spaces) list of issuer account addresses. Example: "addr1,addr2"
  metadata:  	    base64 encoded metadata - arbitrary data attached to the credit class info
  credit type:    the credit class type (e.g. carbon, biodiversity, etc)`,
```

x/ecocredit/client/tx.go
```diff
@@ -70,13 +69,17 @@ Parameters:
 			if err != nil {
 				return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "metadata is malformed, proper base64 string is required")
 			}
+			if args[3] == "" {
+				return errors.New("credit type is required")
```
ruhatch, line 73:
```suggestion
				return sdkerrors.ErrInvalidRequest.Wrap("credit type is required")
```

x/ecocredit/params.go
blushi: On the registry, I've been using http://qudt.org indeed in SHACL schemas

x/ecocredit/msgs.go
```diff
@@ -8,15 +8,19 @@ import (
...
+		&MsgRetire{}, &MsgCancel{}
 )
 
 func (m *MsgCreateClass) ValidateBasic() error {
 	if len(m.Issuers) == 0 {
 		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "issuers cannot be empty")
 	}
 
+	if len(m.CreditType) == 0 {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "credit class must have a credit type")
```
ruhatch, line 21:
```suggestion
		return sdkerrors.ErrInvalidRequest.Wrap("credit class must have a credit type")
```

Saves repeating `sdkerrors`

x/ecocredit/params.go
```diff
@@ -37,12 +45,45 @@ func validateCreditClassFee(i interface{}) error {
...
+
+	// ensure no duplicate credit types and that all precisions conform to hardcoded PRECISION above
+	seenTypes := make(map[string]bool)
+	for _, ct := range v {
+		t := strings.ToLower(ct.Type)
+		t = util.FastRemoveWhitespace(t)
+
+		// TODO: remove after we open governance changes for precision
+		if ct.Precision != PRECISION {
+			return fmt.Errorf("invalid precision %d: precision is currently lockd to %d", ct.Precision, PRECISION)
```
ruhatch, line 62: I'm not sure what context this is called in, but is there an `sdkerrors` error type that would be appropriate to `Wrap` here?

x/ecocredit/params.go
```diff
@@ -37,12 +45,45 @@ func validateCreditClassFee(i interface{}) error {
 	return nil
 }
 
-func NewParams(creditClassFee sdk.Coins) Params {
+func validateCreditTypes(i interface{}) error {
+	v, ok := i.([]*CreditType)
```
ruhatch, line 49: Could you use more descriptive names in this function?
```
v -> creditTypes
ct -> creditType
t -> type?
```

x/ecocredit/params_test.go
```diff
@@ -0,0 +1,44 @@
+package ecocredit
+
+import "testing"
+
+func Test_validateCreditTypes(t *testing.T) {
+	tests := []struct {
```
ruhatch, line 6: Maybe add one with the wrong type?

x/ecocredit/server/credit_type.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/regen-network/regen-ledger/x/ecocredit"
+	"strings"
+)
+
+func (s serverImpl) getCreditType(ctx sdk.Context, creditTypeName string) (*ecocredit.CreditType, error) {
+	creditTypes := s.getAllCreditTypes(ctx)
+	creditTypeName = strings.ToLower(creditTypeName)
+	creditTypeName = strings.TrimSpace(creditTypeName)
+	for _, creditType := range creditTypes {
+		if creditType.Type == creditTypeName {
```
ruhatch, line 15: Can we guarantee that `creditType.Type` is normalized?

x/ecocredit/server/msg_server.go
```diff
@@ -373,35 +378,6 @@ func (s serverImpl) Cancel(goCtx context.Context, req *ecocredit.MsgCancel) (*ec
...
-	var batchInfo ecocredit.BatchInfo
-	err := s.batchInfoTable.GetOne(ctx, orm.RowID(req.BatchDenom), &batchInfo)
-	if err != nil {
-		return nil, err
-	}
-	if req.Issuer != batchInfo.Issuer {
-		return nil, sdkerrors.ErrUnauthorized
-	}
-	store := ctx.KVStore(s.storeKey)
-	key := MaxDecimalPlacesKey(batchDenomT(req.BatchDenom))
```
ruhatch, line 387: We also need to look in `CreateBatch`, `Retire`, and `Cancel` at the places where `MaxDecimalPlacesKey` is being used and replace that with `CreditType` based precision checks.

x/ecocredit/server/testsuite/suite.go
```diff
@@ -618,45 +621,61 @@ func (s *IntegrationTestSuite) TestScenario() {
...
-		}, {
-			"can increase", ecocredit.MsgSetPrecision{
-				Issuer: issuer1, BatchDenom: batchDenom, MaxDecimalPlaces: 8},
-			true,
-		}, {
-			"can NOT change precision of not existing denom", ecocredit.MsgSetPrecision{
-				Issuer: issuer1, BatchDenom: "not/existing", MaxDecimalPlaces: 1},
-			false,
+			name:        "valid eco credit creation",
+			creditTypes: []*ecocredit.CreditType{{Type: "carbon", Units: "kg", Precision: 3}},
```
ruhatch, line 633: Shouldn't this fail, because the precision is wrong? If not, why not? Is our validation code for params actually called?

x/ecocredit/server/testsuite/suite.go
```diff
@@ -618,45 +621,61 @@ func (s *IntegrationTestSuite) TestScenario() {
...
+			tx: ecocredit.MsgCreateClass{Designer: s.signers[0].String(),
+				Issuers: []string{s.signers[1].String(), s.signers[2].String()}, Metadata: nil, CreditType: "carbon"},
```
ruhatch, line 635:
```suggestion
			tx: ecocredit.MsgCreateClass{
			        Designer: s.signers[0].String(),
				Issuers: []string{
				        s.signers[1].String(),
				        s.signers[2].String(),
				}, 
				Metadata: nil, 
				CreditType: "carbon",
		        },
```

Can you wrap these to 80cols and multiline format struct literals once they need wrapping?

x/ecocredit/server/testsuite/suite.go
```diff
@@ -618,45 +621,61 @@ func (s *IntegrationTestSuite) TestScenario() {
...
-	/****   TEST SET PRECISION   ****/
-	precisionCases := []struct {
-		name string
-		msg  ecocredit.MsgSetPrecision
-		ok   bool
+	/****   TEST CREDIT TYPES   ****/
+	creditTypeCases := []struct {
+		name        string
+		creditTypes []*ecocredit.CreditType
+		tx          ecocredit.MsgCreateClass
```
ruhatch, line 628:
```suggestion
		msg         ecocredit.MsgCreateClass
```

This isn't really a transaction!

ruhatch (changes requested):
Design looks good, but a few bits and bobs to sort out

x/ecocredit/server/msg_server.go
ruhatch: And then we can remove all references to `MaxDecimalPlacesKey`

x/ecocredit/server/testsuite/suite.go
technicallyty: Validation code only gets called when the param change route gets hit in the sdk code: [x/params/proposal_handler.go](https://github.com/cosmos/cosmos-sdk/blob/master/x/params/proposal_handler.go#L14-L24)

so here i am directly overriding the param space with the creditTypes field in the test case struct

x/ecocredit/server/credit_type.go
technicallyty: it should - the validation function for the param also calls ToLower and removes whitespace. i just realized though i didnt update this part to the new function in utils so i'll swap that

x/ecocredit/params_test.go
technicallyty: this test just tests against the validation function which only really catches a few things: 

1. duplicates 
2. empty credit type names
3. empty credit units
4. duplicates

do we have a canonical list of what could be valid types that im missing here?

x/ecocredit/params.go
technicallyty: probably invalidrequest - will only really be called during a tx lifecycle for params

x/ecocredit/server/credit_type.go
ruhatch: As far as I can see the validation function is read only, so it does normalise while doing the checks, but doesn't affect the param that actually gets stored. Do you agree?

x/ecocredit/server/testsuite/suite.go
ruhatch: Cool, makes sense

x/ecocredit/server/credit_type.go
technicallyty: ahh right.. would it be better to normalize it here? my initial thought on this is that it would be more ideal to transform the string to a normalized form *and then* store it but not sure if theres a way to do that 🤔

x/ecocredit/server/msg_server.go
technicallyty: how strictly should we enforce the precision - should it be a hard conditional like:

`tradable decimal places must == credit type precision`

or

`tradable decimal places must be <= credit type precision`?

x/ecocredit/server/credit_type.go
ruhatch: Well, there's three pathways to storing it:
- The default value, which should be normalised.
- Some genesis file + import genesis. You can either validate and throw an error, or just normalise quietly.
- Governance update. This one I guess you can only validate? I don't know which message is executed to update a param during governance proposal, so maybe check that and see if there's any chance of modifying the value that gets stored?

x/ecocredit/server/msg_server.go
ruhatch: Definitely `<=`! We don't want to force people to write `1.000000` when they want to send 1 credit!

proto/regen/ecocredit/v1alpha1/types.proto
robert-zaremba: `type` may cause problems in some langauges, where `type` is a keyword. Eg: typescript, haskell...

x/ecocredit/server/testsuite/suite.go
technicallyty: ah, i use the two interchangeably - will switch to msg if that makes more sense here

x/ecocredit/server/credit_type.go
technicallyty: i think there is still the concern if governance updates the params, and gives us something like CaRbOn - we don't really have access to normalizing that before storage? i believe that power lies inside the sdk. 

maybe it would be better to just check normalized string against the given string and if they dont equal, throw error? that way we have a hard enforcement on normalized strings?

technicallyty: ok i strictly enforce normalizaiton at the validation level here: https://github.com/regen-network/regen-ledger/pull/428/commits/231edf12dba2bc5377d5140b1d9f5a24c07f2501

thoughts on this? @aaronc @ruhatch

x/ecocredit/server/testsuite/suite.go
technicallyty: hows this look: https://github.com/regen-network/regen-ledger/pull/428/commits/dd9653f8a9ed26e82a6df79f5192d0096cf94f34 ?

x/ecocredit/server/credit_type.go
ruhatch: Yeah, I think that seems reasonable to me. At what stage in the proposal process is that validation function called? Does it check it before the proposal is submitted, because it would be an annoying UX if they vote a proposal through and then it fails with that kind of normalisation error!!

x/ecocredit/server/testsuite/suite.go
ruhatch: Bellissima!

ruhatch:
@technicallyty looking good, let me know when it's ready for another full review

x/ecocredit/server/credit_type.go
technicallyty: it looks like it will be validated upon proposal request see: <https://github.com/cosmos/cosmos-sdk/blob/master/x/params/proposal_handler.go#L26>

which calls update which then calls validate, which then calls our validate funtion :^) so looks like the UX will be ok

x/ecocredit/client/tx.go
```diff
@@ -49,34 +47,39 @@ func txflags(cmd *cobra.Command) *cobra.Command {
...
 
 Parameters:
-  designer:  address of the account which designed the credit class
-  issuer:    comma separated (no spaces) list of issuer account addresses. Example: "addr1,addr2"
-  metadata:  base64 encoded metadata - arbitrary data attached to the credit class info`,
-		Args: cobra.ExactArgs(3),
+  designer:  	    address of the account which designed the credit class
+  issuer:    	    comma separated (no spaces) list of issuer account addresses. Example: "addr1,addr2"
+  metadata:  	    base64 encoded metadata - arbitrary data attached to the credit class info
+  credit type:    the credit class type (e.g. carbon, biodiversity, etc)`,
```
ruhatch, line 58:
```suggestion
  credit type:    the credit class type (e.g. carbon, biodiversity, etc)`,
  metadata:  	    base64 encoded metadata - arbitrary data attached to the credit class info
```

x/ecocredit/params_test.go
ruhatch: I meant something that's not a `*[]CreditType` like we have in other validation tests, because we might pass in e.g. a `bool`!

ruhatch: And now you might want to add a test that says an non-normalized name fails

x/ecocredit/server/msg_server.go
```diff
@@ -85,7 +94,8 @@ func (s serverImpl) CreateBatch(goCtx context.Context, req *ecocredit.MsgCreateB
 
 			decPlaces := math.NumDecimalPlaces(tradable)
 			if decPlaces > maxDecimalPlaces {
-				maxDecimalPlaces = decPlaces
+				return nil, sdkerrors.ErrInvalidRequest.Wrapf("tradable amount does not conform to credit type "+
+					"precision: is %v, should be %v", decPlaces, maxDecimalPlaces)
```
ruhatch, line 98:
```suggestion
				return nil, sdkerrors.ErrInvalidRequest.Wrapf("tradable amount exceeds precision for credit type: "+
					"is %v, should be < %v", decPlaces, maxDecimalPlaces)
```

ruhatch: The same for other errors too :)

x/ecocredit/server/msg_server.go
```diff
@@ -199,12 +205,19 @@ func (s serverImpl) Send(goCtx context.Context, req *ecocredit.MsgSend) (*ecocre
...
+			return nil, err
+		}
 
-		maxDecimalPlaces, err := getUint32(store, MaxDecimalPlacesKey(denom))
+		classInfo, err := s.getClassInfo(ctx, batchInfo.ClassId)
 		if err != nil {
 			return nil, err
 		}
 
+		maxDecimalPlaces := classInfo.CreditType.Precision
```
ruhatch, line 219: Perhaps we could refactor this into some `s.getBatchPrecision(denom)`?

x/ecocredit/server/msg_server.go
```diff
@@ -342,14 +370,6 @@ func (s serverImpl) Cancel(goCtx context.Context, req *ecocredit.MsgCancel) (*ec
 			return nil, err
 		}
 
-		// Remove the credits from the total_amount in the batch and add
-		// them to amount_cancelled
```
ruhatch, line 346: Could you revive this comment?

ruhatch (changes requested):
A few more changes and also are we going to add the abbreviation stuff?

proto/regen/ecocredit/v1alpha1/types.proto
ruhatch: @technicallyty did you address Robert's comment above?

ruhatch:
Oh and we added some CLI integration tests in #430. Could you add one for the new query and also make sure existing tests are passing?

ruhatch:
The abbreviation for credit types was mentioned here: https://github.com/regen-network/regen-ledger/pull/426#issuecomment-885645248

I think it belongs on this PR instead!

x/ecocredit/params.go
```diff
@@ -87,5 +87,5 @@ func NewParams(creditClassFee sdk.Coins, creditTypes []*CreditType) Params {
 }
 
 func DefaultParams() Params {
-	return NewParams(sdk.NewCoins(sdk.NewCoin(sdk.DefaultBondDenom, DefaultCreditClassFeeTokens)), []*CreditType{})
+	return NewParams(sdk.NewCoins(sdk.NewCoin(sdk.DefaultBondDenom, DefaultCreditClassFeeTokens)), []*CreditType{{Name: "carbon", Units: "kg", Precision: PRECISION}})
```
technicallyty, line 90: Not sure if this is ideal. This is the easiest path forward to appease the new integration tests, as there wasn't really a simple way to set the params with those tests. 

Would this line have to be reverting before build? not really sure how params get shipped with upgrades..

technicallyty: thoughts ? - @aaronc @blushi @clevinson

aaronc: Maybe this is okay. But shouldn’t it be tons not kg?

technicallyty: ok sure - just had it as a placeholder , switched to tons. I opened an issue here https://github.com/regen-network/regen-ledger/issues/457 to discuss what to do about default params

technicallyty:
Should we open an issue for credit type abbreviations or do we want that with this PR? @ruhatch @aaronc

x/ecocredit/params.go
ruhatch: Looks like this should be fine :)

ruhatch:
> Should we open an issue for credit type abbreviations or do we want that with this PR? @ruhatch @aaronc

Either's fine, but my preference would be to do it in this PR as it's not a big change and then the feature comes all at once.

ruhatch (approved):
Looks all good to me now, but I've made a couple of commits, so please could @technicallyty and @blushi take a fairly thorough look?

blushi (commented):
Implementation looks good overall, just left a few minor comments.

blushi (approved):
LGTM

Issue #456, opened 2021-08-04
ORM Tables can't handle key which is prefix of another


ruhatch:
## Summary of Bug

`PrimaryKeyTable`s in the `orm` package cannot have keys where one is a prefix of another. For example, if there is already a key `AA` in the table, then the table will claim to already have the key `A` in the table. Calling `Create` on an object with primary key `A` would then fail, because it would violate the unique constraint.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

clevinson:
@blushi Was the latest from your conversations with Ru that this needs to be addressed as part of v2.0?

blushi:
> @blushi Was the latest from your conversations with Ru that this needs to be addressed as part of v2.0?

yeah just opened up a PR for that #475

PR #459, opened 2021-08-06
fix: Fix primary keys format and add property tests to the ORM package

ruhatch:
## Description

Closes: #456

- Fix primary keys format:
  * `bytes` are length prefixed
  * `string` is null-terminated
  * integer types (i.e. `uint32`, `uint64`, etc.) are stored as 4 or 8 big-endian
- Add property tests to the ORM package
- Make `tableBuilder` and `table` private to the `orm` package since there's no way to enforce that the keys used with it follow what is described above (which would result in the same bug as described in https://github.com/regen-network/regen-ledger/issues/456#issue-960703199)

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


orm/primary_key.go
```diff
@@ -30,10 +32,54 @@ type PrimaryKeyed interface {
...
+func buildPrimaryKey(fields ...interface{}) []byte {
+	bytesSlice := make([][]byte, len(fields))
+	totalLen := 0
+	for i, field := range fields {
+		bytesSlice[i] = primaryKeyFieldBytes(field)
+		totalLen += len(bytesSlice[i])
+	}
+	primaryKey := make([]byte, 0, totalLen)
+	for _, bs := range bytesSlice {
+		primaryKey = append(primaryKey, bs...)
```
robert-zaremba, line 56: You can use `copy` instead of append to squeeze some performance cycles.

orm/primary_key.go
```diff
@@ -30,10 +32,54 @@ type PrimaryKeyed interface {
...
+	return primaryKey
+}
+
+func primaryKeyFieldBytes(field interface{}) []byte {
+	switch v := field.(type) {
+	case []byte:
+		return AddLengthPrefix(v)
+	case string:
+		// TODO(@ruhatch): Do something with strings
+		return []byte(v)
```
robert-zaremba, line 67: We have `UnsafeBytesToString` which we should use here to remove unnecessary memory allocations

orm/primary_key.go
```diff
@@ -30,10 +32,54 @@ type PrimaryKeyed interface {
...
+	default:
+		panic(fmt.Sprintf("Type %T not allowed as primary key field", v))
+	}
+}
+
+// TODO(@ruhatch): This isn't super efficient for small cases, should we do something different?
+func AddLengthPrefix(bytes []byte) []byte {
+	prefixedBytes := make([]byte, 0, 8+len(bytes))
+	prefixedBytes = append(prefixedBytes, EncodeSequence(uint64(len(bytes)))...)
+	prefixedBytes = append(prefixedBytes, bytes...)
```
robert-zaremba, line 79: Use `copy`

orm/primary_key.go
robert-zaremba: BTW: why we don't Len prefix strings?

ruhatch: In the discussion [here](https://github.com/cosmos/cosmos-sdk/discussions/9156), @aaronc suggested using null termination on strings. I haven't implemented it yet, but was planning on doing so if the general approach looked okay.

cyberbono3:
can anyone from west team take care of it since Ru and Marie are on vacation? cc@clevinson

orm/sequence_property_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Init sets up the real Sequence, including choosing a random initial value,
+// and intialises the model state
+func (m *sequenceMachine) Init(t *rapid.T) {
+	// Create context
+	m.ctx = orm.NewMockContext()
+
+	// Create primary key table
+	storeKey := sdk.NewKVStoreKey("test")
+	const testSeqPrefix = iota
```
technicallyty, line 33:
```suggestion
```

unused variable

orm/table.go
```diff
@@ -188,8 +203,13 @@ func (a Table) Delete(ctx HasKVStore, rowID RowID) error {
 	return nil
 }
 
-// Has checks if a key exists. Panics on nil key.
-func (a Table) Has(ctx HasKVStore, rowID RowID) bool {
+// Has checks if a key exists. Panics on nil or empty key.
```
technicallyty, line 206: not sure that this comment is true anymore - looks like it will just return false on the empty key

technicallyty (approved):
looks good, pre-approving

orm/primary_key.go
```diff
@@ -10,43 +12,117 @@ var _ Indexable = &PrimaryKeyTableBuilder{}
...
+// with a length prefix, strings are null-terminated, and integers are encoded
+// using 4 or 8 byte big endian.
+func PrimaryKey(obj PrimaryKeyed) []byte {
+	fields := obj.PrimaryKeyFields()
+	return buildPrimaryKey(fields...)
+}
+
+// buildPrimaryKey encodes and concatenates the PrimaryKeyFields. See PrimaryKey
+// for full documentation of the encoding.
+func buildPrimaryKey(fields ...interface{}) []byte {
```
robert-zaremba: I would prefer that this function and all related functions return error rather than panic.

orm/primary_key.go
```diff
@@ -10,43 +12,117 @@ var _ Indexable = &PrimaryKeyTableBuilder{}
...
+
+// Prefix the byte array with its length as 8 bytes
+func AddLengthPrefix(bytes []byte) []byte {
+	byteLen := len(bytes)
+	if byteLen > 256 {
+		panic("Cannot create primary key with an []byte of length greater than 256 bytes. Try again with a smaller []byte.")
+	}
+
+	newLen := 1 + len(bytes)
+	prefixedBytes := make([]byte, newLen, newLen)
```
robert-zaremba: `gosimple` says that we don't need to put cap

orm/table.go
```diff
@@ -188,8 +203,13 @@ func (a Table) Delete(ctx HasKVStore, rowID RowID) error {
...
+func (a table) Has(ctx HasKVStore, rowID RowID) bool {
```
robert-zaremba: I will rename `rowID` to `key` to match the code description / comment

orm/table.go
robert-zaremba: True - we accept empty key, but panic on `nil`.

robert-zaremba (approved):
tACK. Added tests and pushed cosmetic changes and comment updates. We should note when a function panics.

orm/primary_key.go
robert-zaremba: I've updated the comment, and left function signature, but will be worth to consider not panicing.

orm/sequence_property_test.go
clevinson: @robert-zaremba you resolved- but did you actually commit these suggestions? Or is there some reason this `testSeqPrefix` is still necessary?

technicallyty: it looks like its used now

orm/table.go
technicallyty: the nil value here would just be the empty []byte slice though, so there shouldn't ever be a panic -->> https://play.golang.org/p/eLzItdNL8He

orm/sequence_property_test.go
robert-zaremba: I included it in other changes: https://github.com/regen-network/regen-ledger/pull/459/commits/1075cee0b52bfcbca428c831983944c0d592f370

orm/table.go
robert-zaremba: right, will do a PR to update the comment.

Issue #437, opened 2021-07-27
Update v2.0 UpgradeHandler with regen specific stores


clevinson:
- [x] migrate ecocredit module to base config (in [app.go](https://github.com/regen-network/regen-ledger/blob/master/app/app.go#L113-L133)), and remove from experimental_app_config
- [ ] stable config UpgradeHandler
  - [ ] ecocredit initgenesis (ensure default param for `allowlist_enabled` is overriden, and set to true)
  - [ ] ensure auth/vesting bug fix migration (v0.43) doesn't run

aaronc:
We also need to make sure that the auth/vesting bugfix migration doesn't run because we used a fork with the bugfix. I think we just need to setup the initial version map correctly. /cc @amaurym

blushi:
>  Add migrations for all existing credit vintages (cc @blushi)

@clevinson it seems like this issue is a bit out of date, since we decided that I'm going to re-do the transactions manually on the new devnet, could you update the description above please?

clevinson:
Yep! Updated now, thanks @blushi

clevinson:
Maybe we could actually delete the experimental_config upgrade handler entirely?

amaurym:
> Maybe we could actually delete the experimental_config upgrade handler entirely?

x/group's integration testsuite actually only runs in experimental, so we should keep it for now. https://github.com/regen-network/regen-ledger/blob/7287f51c14838f28fafb2ce9514b68aaaa6d97dd/app/testsuite/modules_test.go

amaurym:
> Maybe we could actually delete the experimental_config upgrade handler entirely?

Oh I misunderstood! Yeah, we can remove the upgrade handler in experimental_config!

PR #479, opened 2021-08-16
fix : Set Params.AllowlistEnabled to true

cyberbono3:
## Description

Closes: #437

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


clevinson:
Why are we overriding this here? Setting this value to `true` should be done in the upgrade handler, not in `InitGenesis`, `InitGenesis` should always just parse whatever is in the genesis data itself.

x/ecocredit/params.go
```diff
@@ -149,7 +149,7 @@ func DefaultParams() Params {
 	return NewParams(
 		sdk.NewCoins(sdk.NewCoin(sdk.DefaultBondDenom, DefaultCreditClassFeeTokens)),
 		[]string{},
-		false,
+		true,
```
cyberbono3, line 152: @clevinson Do you think  this is a good way to go?

app/stable_appconfig.go
```diff
@@ -67,8 +67,15 @@ func (app *RegenApp) registerUpgradeHandlers() {
 			"ibc":          1,
 			"genutil":      1,
 			"transfer":     1,
+			"ecocredit":    1, // we dont run InitiGenesis for ecocredit, right?
 		}
 
+		var params ecocredittypes.Params
```
cyberbono3, line 73: I am not sure if it makes sense.

app/stable_appconfig.go
```diff
@@ -67,8 +67,15 @@ func (app *RegenApp) registerUpgradeHandlers() {
 			"ibc":          1,
 			"genutil":      1,
 			"transfer":     1,
+			"ecocredit":    1, // we dont run InitiGenesis for ecocredit, right?
```
cyberbono3, line 70: so we dont run InitGenesis for ecocredit module, right?

amaurym:
```suggestion
			"ecocredit":    0, // Run InitGenesis for ecocredit
```

We should run InitGenesis for ecocredit

app/stable_appconfig.go
```diff
@@ -67,8 +67,15 @@ func (app *RegenApp) registerUpgradeHandlers() {
...
 		}
 
+		var params ecocredittypes.Params
+		subSpace := app.GetSubspace(ecocredittypes.DefaultParamspace)
+		subSpace.GetParamSet(ctx, &params)
+		params.AllowlistEnabled = true
+	    subSpace.SetParamSet(ctx, &params)
```
amaurym, line 77: We don't need all this

app/stable_appconfig.go
amaurym: Setting params doesn't make much sense, we should just simply run InitGenesis on ecocredit. It makes sense to skip InitGenesis as you did on L70, and instead run it manually.

So we can replace these 4 lines with:
```go
		// force genesis to run so that ecocredit's upgrade handler will execute
		gen := ecocredittypes.DefaultGenesisState()
		gen.Params.AllowlistEnabled = true
		app.mm.Modules[ecocredittypes.ModuleName].InitGenesis(ctx, app.appCodec, app.cdc.MustMarshalJSON(gen))
```

x/ecocredit/params.go
amaurym: I believe we want to keep in `false` by default (if anyone out of regen wants to use the ecocredit module), but in our regen-ledger, we set it to `true` (see above comment). Is that right @clevinson ?

app/stable_appconfig.go
amaurym: Correct, because we run it manually, see https://github.com/regen-network/regen-ledger/pull/479#discussion_r691105190

x/ecocredit/params.go
technicallyty: yes, i believe this also makes it easier for testing internally

app/stable_appconfig.go
amaurym:
```suggestion
			"ecocredit":    1, // we don't run InitGenesis for ecocredit in `RunMigrations`, but manually instead.
```

amaurym (approved):

technicallyty (approved):

cyberbono3:
Do we require to write tests here or merge as it is?

app/stable_appconfig.go
```diff
@@ -45,7 +45,7 @@ func setCustomKVStoreKeys() []string {
 
 func (app *RegenApp) registerUpgradeHandlers() {
 	// This is the upgrade plan name we used in the gov proposal.
-	upgradeName := "v0.43.0-rc0-upgrade"
+	upgradeName := "regen-ledger v2.0-upgrade"
```
clevinson, line 48:
```suggestion
	upgradeName := "v2.0-upgrade"
```

clevinson:
I think we should merge as is, and focus on manual testing of the upgrade handler. Writing integration tests would be a nice-to-have, and maybe something we can get to after the `beta` tag?

cyberbono3:
> I think we should merge as is, and focus on manual testing of the upgrade handler. Writing integration tests would be a nice-to-have, and maybe something we can get to after the `beta` tag?

cc @amaurym @aaronc

Issue #350, opened 2021-05-13
Ecocredit Module Readiness Checklist


clevinson:
## x/ecocredit Module Readiness Checklist

This checklist is to be used for tracking the final internal audit of new Cosmos SDK modules prior to inclusion in a published release.

### Release Candidate Checklist

The following checklist should be gone through once the module has been fully implemented. This audit should be performed directly on `master`, or preferably on a `alpha` or `beta` release tag that includes the module.

The module **should not** be included in any Release Candidate tag until it has passed this checklist.

- [x] API audit (at least 1 person) (@ryanchristo)
  - [x] Are Msg and Query methods and types well-named and organized?
    - [x] #500
    - [x] #516
  - [x] Is everything well documented (inline godoc as well as [`/spec/` folder](https://github.com/cosmos/cosmos-sdk/blob/master/docs/spec/SPEC-SPEC.md) in module directory)
    - [x] spec docs #478
    - [x] proto comments #522
    - [x] inline godoc #552
- [x] State machine audit (at least 2 people) (@aleem1314- https://github.com/regen-network/regen-ledger/pull/492 , @technicallyty - https://github.com/regen-network/regen-ledger/pull/491)
  - [x] @technicallyty 
    - [x] Read through MsgServer code and verify correctness upon visual inspection
    - [x] Ensure all state machine code which could be confusing is properly commented
    - [x] Make sure state machine logic matches Msg method documentation
  - [x] @aleem1314 
    - [x] Ensure that all state machine edge cases are covered with tests and that test coverage is sufficient (at least 90% coverage on module code)
    - [x] Assess potential threats for each method including spam attacks and ensure that threats have been addressed sufficiently. This should be done by writing up threat assessment for each method
    - [x] Assess potential risks of any new third party dependencies and decide whether a dependency audit is needed (ORM AUDIT NEEDED)
  - [x] Followups to be addressed:
    - [x] get consensus around the % fee we should burn or are we just leaving it at 100%? #517
    - [x] CreateBatch, Send, Retire, and Cancel all require looping to some degree. Should we be charging some set gas per loop step? #519
    - [x] there is no check for metadata in ValidateBasic. Should this be validated at all? shouldn't we at least enforce a length? what if someone sent a tx with a HUGE amount of metadata? #520
- [x] #509
  - [x] Genesis import and export of all state
  - [x] Query services
  - [x] CLI methods
  - [x] All necessary migration scripts are present (if this is an upgrade of existing module)

[...]

PR #491, opened 2021-08-19
perf/fix: ecocredit cleanup (audit)

technicallyty:
## Description

ref: #350 

During my audit of Eco Credit, I found a few areas to squeeze out perf cycles. Also found a few areas of eco that needed updating and or fixing.

+Added `NewPositiveDecFromString` as we were still using `ParsePositiveDecimal` in `msgs.go`. 
+Tighten up code in `ValidateBasic` to use the `if [assignment]; [condition] {}` syntax.
-Deleted redundant test cases against CLI - testing for what was thought to be missing _specific_ arguments, however, the execution of the cmd is immediately rejected as long as `exactArgs != len(argsPassed)`.  Which arg passed and not passed makes no difference
-Reduce usage of pointers with Credit Type. This value is read from a few times, so `getCreditType` now returns the value instead of a pointer.
-Remove unnecessary error checking in GetSigners, as this will be done in `ValidateBasic` beforehand.
-Change function name from `getAddAndSetDecimal` to `addAndSetDecimal` as it felt confusing. `Get` usually implies something is returned, but this is not the case with this function.
-Remove `get/setUint32` functions as they are no longer used
-The allowlistEnabled and isDesignerAllowlisted function both fetched the params, which triggers a loop. changed to get params only once beforehand.
-assertClassIssuer and nextBatchinClass both required fetching classInfo. This is now fetched beforehand and passed into these functions.
-In retire and cancel functions, the holder in the request was being converted to address every step in the loop. this was brought outside the loop to only execute once.
-added missing test case for precision check in send
-deleted duplicate test case for precision check in cancel

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/client/testsuite/query.go
```diff
@@ -89,7 +89,7 @@ func (s *IntegrationTestSuite) TestQueryClassInfo() {
 		expectedClassInfo *ecocredit.ClassInfo
 	}{
 		{
-			name:           "missing credit class",
```
technicallyty, line 92: technically not missing specifically the credit class - when cmd's have `exactArgs` it will fail immediately if len(argsPassed) isn't equal to amount of exactArgs.

x/ecocredit/client/testsuite/tx.go
```diff
@@ -199,29 +199,11 @@ func (s *IntegrationTestSuite) TestTxCreateClass() {
...
-			args:           []string{validCreditType},
-			expectErr:      true,
-			expectedErrMsg: "accepts 4 arg(s), received 1",
-		},
-		{
-			name:           "missing metadata",
-			args:           []string{val0.Address.String(), val0.Address.String()},
-			expectErr:      true,
-			expectedErrMsg: "accepts 4 arg(s), received 2",
-		},
```
technicallyty, line 224: this is essentially all the same

x/ecocredit/server/helpers.go
```diff
@@ -37,7 +35,7 @@ func setDecimal(store sdk.KVStore, key []byte, value math.Dec) {
 	}
 }
 
-func getAddAndSetDecimal(store sdk.KVStore, key []byte, x math.Dec) error {
```
technicallyty, line 40: i felt this name was kinda strange, as the `get` usually implies it would return the decimal

x/ecocredit/server/msg_server.go
```diff
@@ -331,11 +340,6 @@ func (s serverImpl) Retire(goCtx context.Context, req *ecocredit.MsgRetire) (*ec
 			return nil, err
 		}
 
-		holderAddr, err := sdk.AccAddressFromBech32(holder)
```
technicallyty, line 334: no point to do this every loop iteration, `holder` never changes

x/ecocredit/server/msg_server.go
```diff
@@ -525,24 +524,12 @@ func (s serverImpl) getBatchPrecision(ctx types.Context, denom batchDenomT) (uin
...
+		allowListedAddr, _ := sdk.AccAddressFromBech32(addr)
+		if designer.Equals(allowListedAddr) {
+			return true
 		}
 	}
-	return false, nil
-}
-
-// Checks if the allowlist of credit class designers is enabled
-func (s serverImpl) allowlistEnabled(ctx sdk.Context) bool {
```
technicallyty, line 544: params is now just fetched directly in the `CreateClass` method instead of fetching it twice in both `allowlistEnabled `and `isDesignerAllowlisted`.

x/ecocredit/server/testsuite/suite.go
```diff
@@ -270,12 +270,6 @@ func (s *IntegrationTestSuite) TestScenario() {
 			toCancel:  "0",
 			expectErr: true,
 		},
-		{
-			name:      "can't cancel beyond precision of batch",
```
technicallyty, line 274: duplicate test case

aleem1314 (approved):
lgtm

x/ecocredit/server/msg_server.go
```diff
@@ -444,12 +450,7 @@ func (s serverImpl) Cancel(goCtx context.Context, req *ecocredit.MsgCancel) (*ec
 
 // assertClassIssuer makes sure that the issuer is part of issuers of given classID.
 // Returns ErrUnauthorized otherwise.
-func (s serverImpl) assertClassIssuer(goCtx context.Context, classID, issuer string) error {
-	ctx := types.UnwrapSDKContext(goCtx)
-	classInfo, err := s.getClassInfo(ctx, classID)
-	if err != nil {
-		return err
-	}
+func (s serverImpl) assertClassIssuer(classInfo *ecocredit.ClassInfo, issuer string) error {
```
ruhatch, line 453: You could move this to a method on `ClassInfo` now that you've removed the fetching logic

ruhatch (changes requested):
Couple of small changes, but looks good

x/ecocredit/server/msg_server.go
technicallyty: https://github.com/regen-network/regen-ledger/pull/491/commits/5e380b62603390a94c147da06eb9553161c264cb

x/ecocredit/client/testsuite/tx.go
```diff
@@ -575,7 +557,7 @@ func (s *IntegrationTestSuite) TestTxCreateBatch() {
 				s.commonTxFlags()...,
 			),
 			expectErr:      true,
-			expectedErrMsg: "expected a non-negative decimal, got abcde",
```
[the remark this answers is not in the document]
technicallyty: actually the test string is still changed, but the error message is more precise, lmk if that commit looks ok

ruhatch (approved):
LGTM!

Issue #535, opened 2021-09-13
Add support for updating a credit class


clevinson:
Currently we don't actually have implemented a msg for updating a credit class (by CreditClassAdmin).

We discussed this when changing the creator's name from `CreditClassDesigner` to `CreditClassAdmin`, and also noticed that its actually [in the RFC](https://github.com/regen-network/regen-ledger/blob/master/rfcs/001-ecosystem-service-credit-module.md#update-credit-class).

Unless there was a specific reason we decided to exclude it from Regen Ledger v2.0, I think it would be valuable to add in and include in this release. In particular, I think it would be very useful for credit class admins to be able to update issuer lists as those may change over time.

## Proposal

```proto

// MsgUpdateClassAdmin is the Msg/UpdateClassAdmin request type.
message MsgUpdateClassAdmin {

  // admin is the address of the account that is the admin of the credit class.
  string admin = 1;

  // class_id is the unique ID of the credit class.
  string class_id = 2;

  // new_admin is the address of the new admin of the credit class.
  string new_admin = 3;
}

// MsgUpdateClassIssuers is the Msg/UpdateClassIssuers request type.
message MsgUpdateClassIssuers {

  // admin is the address of the account that is the admin of the credit class.
  string admin = 1;

  // class_id is the unique ID of the credit class.
  string class_id = 2;

  // issuers are the updated account addresses of the approved issuers.
  repeated string issuers = 3;
}

// MsgUpdateClassMetadata is the Msg/UpdateClassMetadata request type.
message MsgUpdateClassMetadata {

  // admin is the address of the account that is the admin of the credit class.
  string admin = 1;

  // class_id is the unique ID of the credit class.
  string class_id = 2;

  // metadata is the updated arbitrary metadata to be attached to the credit class.
  bytes metadata = 3;
}
```

aaronc:
Yeah, I just didn't make time for adding these in my initial MVP

PR #539, opened 2021-09-15
feat: add methods for updating a credit class

technicallyty:
## Description

Adds support for updating metadata, admin, and issuers for credit classes

Closes: #535 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/server/msg_server.go
```diff
@@ -447,6 +447,57 @@ func (s serverImpl) Cancel(goCtx context.Context, req *ecocredit.MsgCancel) (*ec
...
+	cInfo, err := s.getClassInfo(ctx, req.ClassId)
+	if err != nil {
+		return nil, err
+	}
+
+	if cInfo.Admin != req.Admin {
+		return nil, sdkerrors.ErrUnauthorized.Wrapf("you are not the administrator of this class")
+	}
+
+	cInfo.Metadata = req.Metadata
```
technicallyty, line 495: this should probably cost some gas as metadata can be variable length - doesnt look like we have a set hardcapped length for metadata length. we should also charge some gas based on the length itself

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -217,4 +226,52 @@ message MsgCancel {
 }
 
 // MsgCancelResponse is the Msg/Cancel response type.
-message MsgCancelResponse {}
\ No newline at end of file
+message MsgCancelResponse {}
+
+// MsgUpdateClassAdmin is the Msg/UpdateClassAdmin request type.
+message MsgUpdateClassAdmin {
```
technicallyty, line 232: we could also perhaps just squeeze all these new messages into 1 MsgUpdateClass -- and just have each field (admin, metadata, issuers) as optional? not sure which approach is more ideal here

aleem1314: +1 for using a single message type.

technicallyty: i looked into this and wasn't sure what the best way to implement it is - specifically with the issuers type ( []string ). I'm unsure if there's a good way in proto to tell if a field has been explicitly set. otherwise, it appears a bit troublesome to figure out intent via the default values

technicallyty: so looking at this again, I could just make each update field its own message i.e.:

```proto
message MsgUpdateCreditClass {
   string admin = 1;
   string class_id = 2;
   UpdateMetadata metadata = 3;
  // ...
}

message UpdateMetadata {
   bytes metadata = 1;
}
```

that way each update field will be a pointer and we can check intent via nil. Does this seem like a sane path forward?

ryanchristo: I think the way you have it now with separate update messages would be best. In most cases, I would imagine the admin would only be changing one field at a time. It's also done this way in the group module.

x/ecocredit/msgs.go
```diff
@@ -254,3 +254,96 @@ func (m *MsgCancel) GetSigners() []sdk.AccAddress {
...
+func (m *MsgUpdateClassIssuers) ValidateBasic() error {
+	if _, err := sdk.AccAddressFromBech32(m.Admin); err != nil {
+		return sdkerrors.ErrInvalidAddress
+	}
+
+	if err := ValidateClassID(m.ClassId); err != nil {
+		return err
+	}
+
+	if len(m.Issuers) == 0 {
```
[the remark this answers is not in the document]
ryanchristo: Good question. We require issuers when creating a credit class, so I would lean towards requiring issuers when updating a credit class. I could however see wiping issuers as a way of disabling the credit class but the admin could also assign themselves as the only issuer to accomplish.

x/ecocredit/server/msg_server.go
ryanchristo: We have a max length of 256 following https://github.com/regen-network/regen-ledger/pull/540. We should make sure the metadata does not exceed this max length by adding a check in validate basic.

x/ecocredit/msgs.go
```diff
@@ -254,3 +254,96 @@ func (m *MsgCancel) GetSigners() []sdk.AccAddress {
...
+
+func (m MsgUpdateClassMetadata) Route() string { return sdk.MsgTypeURL(&m) }
+
+func (m MsgUpdateClassMetadata) Type() string { return sdk.MsgTypeURL(&m) }
+
+func (m MsgUpdateClassMetadata) GetSignBytes() []byte {
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&m))
+}
+
+func (m *MsgUpdateClassMetadata) ValidateBasic() error {
```
ryanchristo, line 334: We need to make sure the metadata does not exceed the max length.

```
	if len(m.Metadata) > MaxMetadataLength {
		return ErrMaxLimit.Wrap("credit class metadata")
	}
```

x/ecocredit/server/testsuite/suite.go
```diff
@@ -83,6 +83,168 @@ func (s *IntegrationTestSuite) fundAccount(addr sdk.AccAddress, amounts sdk.Coin
...
+	createClsRes, err := s.msgClient.CreateClass(s.ctx, &ecocredit.MsgCreateClass{Admin: admin.String(), Issuers: []string{issuer1}, Metadata: nil, CreditTypeName: "carbon"})
+	s.Require().NoError(err)
+	s.Require().NotNil(createClsRes)
+	classID := createClsRes.ClassId
+
+	testCases := []struct {
+		name   string
+		msg    ecocredit.MsgUpdateClassMetadata
+		expErr bool
+	}{
```
ryanchristo, line 209: We should add a test case for metadata exceeding max length.

ryanchristo (commented):
Looking good. I think we should keep the update messages separate, which is the same as the group module. Also, we should check for metadata length following https://github.com/regen-network/regen-ledger/pull/540.

x/ecocredit/server/testsuite/suite.go
technicallyty: added a test for this in `msg_test.go`

x/ecocredit/msgs.go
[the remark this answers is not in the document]
technicallyty: okay, leaving as is

x/ecocredit/client/tx.go
```diff
@@ -361,3 +364,125 @@ Parameters:
 		},
 	})
 }
+
+func TxUpdateClassMetadataCmd() *cobra.Command {
+	return txflags(&cobra.Command{
+		Use:   "update-metadata [class-id] [metadata]",
```
blushi, line 370:
```suggestion
		Use:   "update-class-metadata [class-id] [metadata]",
```

x/ecocredit/client/tx.go
```diff
@@ -361,3 +364,125 @@ Parameters:
...
+			}
+
+			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), &msg)
+		},
+	})
+}
+
+func TxUpdateClassAdminCmd() *cobra.Command {
+	return txflags(&cobra.Command{
+		Use:   "update-metadata [class-id] [admin]",
```
blushi, line 412:
```suggestion
		Use:   "update-class-admin [class-id] [admin]",
```

x/ecocredit/client/tx.go
```diff
@@ -361,3 +364,125 @@ Parameters:
...
+			}
+
+			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), &msg)
+		},
+	})
+}
+
+func TxUpdateClassIssuersCmd() *cobra.Command {
+	return txflags(&cobra.Command{
+		Use:   "update-issuers [class-id] [issuers]",
```
blushi, line 452:
```suggestion
		Use:   "update-class-issuers [class-id] [issuers]",
```

x/ecocredit/msgs_test.go
```diff
@@ -778,3 +778,138 @@ func TestMsgCancel(t *testing.T) {
...
+		"invalid: bad class ID": {
+			src:    MsgUpdateClassMetadata{Admin: a1.String(), ClassId: "6012949", Metadata: []byte("hello world")},
+			expErr: true,
+		},
+		"invalid: no class ID": {
+			src:    MsgUpdateClassMetadata{Admin: a1.String()},
+			expErr: true,
+		},
+		"invalid: metadata too large": {
+			src:    MsgUpdateClassMetadata{Admin: a1.String(), ClassId: "C01", Metadata: []byte(randomString(256))},
```
blushi, line 890: this could just work too (see other metadata length related tests in this file):
```suggestion
			src:    MsgUpdateClassMetadata{Admin: a1.String(), ClassId: "C01", Metadata: []byte(simtypes.RandStringOfLength(r, 288)),
```

blushi (commented):
Looks good overall, just a few comments

blushi (approved):
utACK

aleem1314 (approved):
LGTM

Issue #505, opened 2021-08-27
Ability to List/Buy/Sell Ecocredits


dtkong:
## Summary
With the ecocredit module allowing the creation and transfer of ecocredits, we'd now like the ability to list, sell, and purchase. This will be minimum viable functionality towards an ecocredit exchange.  

## User stories
* As an issuer, I'd like the ability to list and sell an ecocredit at a particular price, so that I can monetize my ecocredits. 
* As a user, I'd like the ability to purchase a listed ecocredit at a listed price.

## Problem Definition
Issuers need the ability to not only create and transfer ecocredits but also market them. 

## Proposal
* Issuer can list an ecocredit at a fixed price denominated in $REGEN
* Buyers can purchase a listed ecocredit at the issuer's fixed price. 
* ~TBD whether an ecocredit is automatically-retired upon purchase or whether there's a choice of the seller to auto-retire vs let the buyer retire.~ Update: an ecocredit is auto-retired upon purchase.

technicallyty:
It might be cool to implement a credit marketplace using CosmWasm, we could deploy some marketplace contracts to other chains and have our credits be supported everywhere cosmwasm is supported.

clevinson:
I think there's a few different ways we could choose to build something like this out, one of which is the CosmWasm approach Tyler's illustrating. I think using CosmWasm may require that we implement #407 first though. Cosmwasm has some basic x/bank integrations, but not sure how easy it is for cosmwasm to integrate / interact with other module's substores.

Here's the options as I see it. All of these could implement the desired functionality, but have different tradeoffs. Happy to hear others' thoughts w/ more pros&cons too!

### 1. Cosmwasm smart contract

We integrate CosmWasm (which has long been discussed, though currently still TBD exactly when) and write a basic CosmWasm smart contract that can hold assets of any denom in x/bank (which includes ecocredit assets) and offer them for sale at a fixed price.  Seller can withdraw unsold credits from the contract at any point in time.

**Pros:**
- reusable code (any chain that supports cosmwasm could have this contract deployed and provide the exact same functionality). In fact -- if we do a bit of digging, this type of cosmwasm smart contract may already exist off the shelf from another project?
- bringing CosmWasm into Regen (something we've talked about for a long time, but are still waiting for the initial use case to prioritize it)
- Can be deployed to a live chain without a coordinated software proposal & upgrade (we would still need one initial upgrade to integrate cosmwasm into Regen Ledger)
[...]

PR #560, opened 2021-09-27
feat: ecocredit direct buy/sell proto definitions

aaronc:
ref #505 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/proto/regen/ecocredit/v1alpha1/tx.proto
+++ b/proto/regen/ecocredit/v1alpha1/tx.proto
@@ -301,7 +301,7 @@ message MsgSellResponse {
 
 message MsgCancelSellOrder {
   string owner = 1;
-  repeated uint64 sell_order_ids = 1;
+  repeated uint64 sell_order_ids = 2;
 }
 
 message MsgCancelSellOrderResponse {}
@@ -309,6 +309,7 @@ message MsgCancelSellOrderResponse {}
 message MsgBuyDirect {
   string buyer = 1;
   repeated Order orders = 2;
+
   message Order {
     uint64 sell_order_id = 1;
     string quantity = 2;
--- a/proto/regen/ecocredit/v1alpha1/tx.proto
+++ b/proto/regen/ecocredit/v1alpha1/tx.proto
@@ -39,6 +39,12 @@ service Msg {
 
   // UpdateClassMetadata updates the credit class metadata
   rpc UpdateClassMetadata(MsgUpdateClassMetadata) returns (MsgUpdateClassMetadataResponse);
+
+  rpc Sell(MsgSell) returns (MsgSellResponse);
+
+  rpc CancelSellOrder(MsgCancelSellOrder) returns (MsgCancelSellOrderResponse);
+
+  rpc BuyDirect (MsgBuyDirect) returns (MsgBuyDirectResponse);
 }
 
 // MsgCreateClass is the Msg/CreateClass request type.
@@ -275,3 +281,39 @@ message MsgUpdateClassMetadata {
 
 // MsgUpdateClassMetadataResponse is the MsgUpdateClassMetadata response type.
 message MsgUpdateClassMetadataResponse {}
+
+message MsgSell {
+  string owner = 1;
+  repeated Order orders = 2;
+
+  message Order {
+    string batch_denom = 1;
+    string quantity = 2;
+    string ask_denom = 3;
+    string ask_price = 4;
+    bool dont_auto_retire = 5;
+  }
+}
+
+message MsgSellResponse {
+  repeated uint64 sell_order_ids = 1;
+}
+
+message MsgCancelSellOrder {
+  string owner = 1;
+  repeated uint64 sell_order_ids = 1;
+}
+
+message MsgCancelSellOrderResponse {}
+
+message MsgBuyDirect {
+  string buyer = 1;
+  repeated Order orders = 2;
+  message Order {
+    uint64 sell_order_id = 1;
+    string quantity = 2;
+    bool dont_auto_retire = 3;
+  }
+}
+
+message MsgBuyDirectResponse {}


proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -274,4 +280,41 @@ message MsgUpdateClassMetadata {
...
+
+message MsgSell {
+  string owner = 1;
+  repeated Order orders = 2;
+
+  message Order {
+    string batch_denom = 1;
+    string quantity = 2;
+    string ask_denom = 3;
+    string ask_price = 4;
```
clevinson, line 293: Could this just be one `Coin` field `ask_coin` ?

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -274,4 +280,41 @@ message MsgUpdateClassMetadata {
...
+    bool dont_auto_retire = 5;
```
clevinson, line 294: I imagine that long-term we would want auto retire disabled by default. If that's the case, should this bool be flipped? e.g. set this field to `bool auto_retire_enabled = 5;`?

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -274,4 +280,41 @@ message MsgUpdateClassMetadata {
...
+message MsgCancelSellOrderResponse {}
+
+message MsgBuyDirect {
+  string buyer = 1;
+  repeated Order orders = 2;
+
+  message Order {
+    uint64 sell_order_id = 1;
+    string quantity = 2;
+    bool dont_auto_retire = 3;
```
clevinson, line 316: What does it mean for a buyer to provide an `auto_retire` flag? Isn't auto-retire something that would be set / determined by the seller only?

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: Sure 👍

aaronc: I'm assuming that the `ask_coin` is the price per one unit right?

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: Hmmm... my thinking is we want it on by default - thus the awkward "dont_auto_retire"

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: I'm thinking we would have both default to auto-retire. Even if the seller checks `dont_auto_retire`, they'll still retire unless the buyer also checks `dont_auto_retire`. It biases the whole design towards the intended use (i.e. retirement) which I've gotten the sense is desirable

proto/regen/ecocredit/v1alpha1/tx.proto
clevinson: yes that's what my assumption would be.

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -274,4 +280,41 @@ message MsgUpdateClassMetadata {
...
+message MsgUpdateClassMetadataResponse {}
+
+message MsgSell {
+  string owner = 1;
+  repeated Order orders = 2;
+
+  message Order {
+    string batch_denom = 1;
+    string quantity = 2;
+    string ask_denom = 3;
```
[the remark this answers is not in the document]
robert-zaremba: `ask_price` is a minimum sell price, `ask_denom` is the currency.
It will be worth add comments to the attributes.

proto/regen/ecocredit/v1alpha1/tx.proto
robert-zaremba: let's add a doc string to specify the details.

proto/regen/ecocredit/v1alpha1/tx.proto
robert-zaremba: Can we use `no_` prefix instead of `dont`?

proto/regen/ecocredit/v1alpha1/tx.proto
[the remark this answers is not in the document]
aaronc: exactly. sorry about lack of comments! will get to that soon

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -274,4 +280,41 @@ message MsgUpdateClassMetadata {
...
+
+message MsgCancelSellOrderResponse {}
+
+message MsgBuyDirect {
+  string buyer = 1;
+  repeated Order orders = 2;
+
+  message Order {
+    uint64 sell_order_id = 1;
+    string quantity = 2;
```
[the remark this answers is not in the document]
aaronc: I just added a `MsgUpdateSellOrder` now which allows changing all the sell order attributes as well as canceling

aaronc:
Made some updates to this, adding the ability to update buy orders and allow ask denoms through governance. I also put comments everywhere.

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: don't have a strong preference. `disable_auto_retire` would be most clear i think

blushi: I prefer `disable_auto_retire` too

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -39,6 +40,18 @@ service Msg {
...
+  rpc Sell(MsgSell) returns (MsgSellResponse);
+
+  // UpdateSellOrders updates existing sell orders.
+  rpc UpdateSellOrders(MsgUpdateSellOrders) returns (MsgUpdateSellOrdersResponse);
+
+  // BuyDirect creates buy orders directly against sell orders.
+  rpc BuyDirect (MsgBuyDirect) returns (MsgBuyDirectResponse);
+
+  // AllowAskDenom is a governance operation which authorizes a new ask denom to be used in sell orders
+  rpc AllowAskDenom (MsgAllowAskDenom) returns (MsgAllowAskDenomResponse);
```
blushi, line 54: so this is relying on https://github.com/cosmos/cosmos-sdk/issues/9438?

blushi (commented):
API looks good overall, just one concern regarding sell orders update

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -274,4 +287,133 @@ message MsgUpdateClassMetadata {
...
+
+    // ask_price is the price the seller is asking for each unit of the
+    // batch_denom. Each credit unit of the batch will be sold for at least the
+    // ask_price or more.
+    cosmos.base.v1beta1.Coin ask_price = 3;
+
+    // dont_auto_retire disables auto-retirement of credits which allows a
+    // buyer to disable auto-retirement in their buy order enabling them to
+    // resell the credits to another buyer.
+    bool dont_auto_retire = 4;
```
anilCSE, line 320: Can we just call it `auto_retire` and use it otherwise. `dont` prefix is bit odd

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -274,4 +287,133 @@ message MsgUpdateClassMetadata {
...
+
+    // new_quantity is the updated quantity of credits available to sell, if it
+    // is set to zero then the order is cancelled.
+    string new_quantity = 2;
+
+    // new_ask_price is the new ask price for this sell order
+    cosmos.base.v1beta1.Coin new_ask_price = 3;
+
+    // dont_auto_retire updates the dont_auto_retire field in the sell order.
+    bool dont_auto_retire = 4;
```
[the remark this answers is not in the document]
aaronc: This API assumes everything needs to be explicitly set in the update message. We can have a more granular update API of course if that's desired

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -274,4 +287,133 @@ message MsgUpdateClassMetadata {
...
+    // bid price is the bid price for this buy order. A credit unit will be
+    // settled at a purchase price that is no more than the bid price.
+    cosmos.base.v1beta1.Coin bid_price = 3;
+
+    // dont_auto_retire allows auto-retirement to be disabled. If it is set to
+    // the credits will not auto-retire and can be resold assuming that the
+    // corresponding sell order has auto-retirement disabled. If the sell order
+    // hasn't disabled auto-retirement and the buy order tries to disable it,
+    // that buy order will fail.
+    bool dont_auto_retire = 4;
```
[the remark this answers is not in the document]
aaronc: Yes, exactly

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: Yes

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: My understanding is that we generally want to default to auto-retirement. Maybe `disable_auto_retirement` is clearer?

anilCSE: Yes, that sounds better

proto/regen/ecocredit/v1alpha1/tx.proto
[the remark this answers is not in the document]
blushi: Ok could you add this assumption to the docs then?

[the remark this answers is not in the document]
blushi: Another thing is that we've started using a granular update API within the ecocredits module already for updating credit classes so I'm afraid that could result in inconsistencies across the ecocredits msg service and confusions for users.

clevinson:
Another thing that came up in a conversation w/ Ryan later today...

It seems like we're going with a non-escrow model approach in this current architecture (e.g. SellOrders are tracked separately from a person's balance, and when a person opens a sell order they could still theoretically transfer those ecocredits to another user without invalidating the existence of the sell order).

A nice principle of this is that a user can list multiple sell orders against the same batch of ecocredits (each sell order with a different asking price denom).

However, it means that the existence of a sell order on its own is not enough to guarantee that it can be executed on.

Let's say I'm a holder of 100 ecocredits and I make a sell order for 50 ecocredits at 5 REGEN/credit price.  I could transfer 80 of these ecocredits to Alice, and then if a user tries to buy 50 ecocredits against my sell order, it would error out since I don't have sufficient ecocredits in my balance.

Similarly, if we look at the application developer trying to build a marketplace of sell orders. It's not enough to query all sell orders of a given credit class or batch and expose them to potential buyers. If you want to only show sell orders that can be executed at their full quantity, you have to be checking against each sell order to verify that the seller has those credits in their balance - right?

Or is your intention already that querying would show you not the sell order quanitty, but the minimum of sell order quantity and tradable credits owned by the seller?

clevinson:
Stepping back, I'm actually not 100% sure if we want to enable listing for multiple prices until we have more dynamic pricing available (via price feeds or oracles). 

Explicitly setting prices in multiple denoms opens up an arbitrage opportunity (at least when auto-retire is disabled), that I'm not sure we want to encourage.  Then again, this is only a 1-sided marketplace for now so its not like the arbitrage opportunity is that straightforward to exploit... but this could be the case when we have hte ability for buyers to place asks directly without corresponding to a specific sell order.

aaronc:
Good questions @clevinson. Another question that comes to mind is whether we want to allow buy orders to be partially filled in the case that the full quantity isn't available. Maybe filling partially is a flag that can be set in the buy order?

In general I'm skeptical of enabling more than one denom at all for the credit marketplace because it reduces effective liquidity and also like you say could create weird arbitrage opportunities.

aaronc:
But that still doesn't help because I imagine we'll do batch processing once there are other types of buy orders. Even without batches there's no way of guaranteeing someone else doesn't buy part of the remaining supply first in an earlier tx.

PR #608, opened 2021-10-29
feat(x/ecocredit)!: add projects to the credit data model

aaronc:
## Description

Adds projects to the credit data model. In this model:
* credit types contain classes
* classes contain projects
* projects contain batches

It is assumed that the issuer of a project will issue all batches within the project.

This is proto breaking and hence I bumped the proto package version to `v1alpha2`.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


S4mmyb:
I'm not sure if the issuer of the project will be the one issuing all the batches within that project. Projects can run for up to 25 years and if credits are being issued ex-ante that's a pretty long time to commit to being an issuer. I think any issuer of a credit class should be able issue credit batches to a project living in that credit class

aaronc:
> I'm not sure if the issuer of the project will be the one issuing all the batches within that project. Projects can run for up to 25 years and if credits are being issued ex-ante that's a pretty long time to commit to being an issuer. I think any issuer of a credit class should be able issue credit batches to a project living in that credit class

Is it true that an issuer of credits would be the party to create the project? Or is that some other party?

aaronc:
Could theoretically a project developer (basically anyone) create a project?

proto/regen/ecocredit/v1alpha2/tx.proto
```diff
@@ -72,19 +74,47 @@ message MsgCreateClass {
...
 // MsgCreateBatch is the Msg/CreateBatch request type.
 message MsgCreateBatch {
 
   // issuer is the address of the batch issuer.
   string issuer = 1;
 
-  // class_id is the unique ID of the class.
-  string class_id = 2;
+  // project_id is the unique ID of the project this batch belongs to.
+  string project_id = 2;
```
clevinson, line 117: Similar to what I said in the `types.proto` file- it may be nice to keep `class_id` for easier querying.

Additionally - if not all `project_id`s are actually corresponding to on-chain entities (e.g. for Verra credit classes), then I think we still should require the `class_id` field here.

clevinson (commented):
Thanks @aaronc !

I'm in general in favor of having projects as on-chain entities.  One thing I'm still not totally sure of is whether that's required for all credit classes though.  e.g. for "Verra" credits -- how do we want to handle project_id?

Few Options I see:

1. Make `project_id` optional (or have the "optionality" set at the credit class level) so that for Verra credits we just leave it blank
2. Register on-chain entities for projects for 3rd party credit classes. This would have to be done by the issuer, even if we change our data model to have project developers generally managing on-chain project registration, because I assume we cannot assume that all Verra project developers for our brokered credits will be willing to register on-chain entities for their projects in the immediate future.
3. Change `project_id` to be namespaced in a way that could resolve either to on-chain entities, or known off-chain entities (I elaborate on one way to do this in a later comment here).


Separately, there's still the question (which @aaronc and @S4mmyb were already discussing) around who can create projects.  IMO I think it's better to have this be permissionless, so that on-chain project entities can be managed independently of a specific issuer (as they may need to be managed over decade-long time spans as @S4mmyb indicated).

One middle ground that we could have is an apply-to-register model, where anyone (typically a project developer) can apply to register a project under a credit class, but an issuer of the credit class must approve the application in order for the project to be successfully registered on-chain and start receiving credits. We could have an expiry on applications so not to bloat state unnecessarily.

aaronc:
I thought we already established that verra credits always have project IDs anyway

clevinson:
Sure- that makes sense to me, and does seem in line with what I've heard from the science team. 

does this PR already handle the use case of pre-existing verra project ids?

Maybe I'm just not understanding how you intended that to work.

aaronc:
You would reference the verra project ID in the project metadata

proto/regen/ecocredit/v1alpha2/types.proto
```diff
@@ -30,18 +30,32 @@ message ClassInfo {
   uint64 num_batches = 6;
 }
 
+// ProjectInfo represents the high-level on-chain information for a project.
+message ProjectInfo {
+
+  // project_id is the unique ID of the project.
+  string project_id = 1;
```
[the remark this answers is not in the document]
S4mmyb: This is a good point @clevinson. I agree it seems important we include information about the 'origin registry'. I think this will help reduce complexity and makes it a lot easier to find the original issuer of the credits. I think both solutions would work well. If we choose your second solution I think it would make more sense to put the `project_registry` field in  `ProjectInfo` rather than the credit class because it's still a bit unclear how the credit class will/can be used. For instance, someone might define a credit class to allow methodologies from multiple registries. Storing it in `ProjectInfo` makes more sense because the project itself should only be registered with a single registry.

proto/regen/ecocredit/v1alpha2/tx.proto
S4mmyb: I agree it would make it easier for querying.

S4mmyb:
> You would reference the verra project ID in the project metadata

I think the issue here would be that someone could use the project metadata field to store something other than the projectID, such as the project page or link to the project plan, which would then make it more difficult to locate the projectID. I think having an explicit field for projectID when you create a new project makes more sense.

aaronc:
> > You would reference the verra project ID in the project metadata
> 
> 
> 
> I think the issue here would be that someone could use the project metadata field to store something other than the projectID, such as the project page or link to the project plan, which would then make it more difficult to locate the projectID. I think having an explicit field for projectID when you create a new project makes more sense.

So the project ID is auto-generated. I think it's a special case where there is an external project ID so I think it might be confusing to have that field. Can't we just make sure we use metadata in a specific way for verra projects to reference the verra project ID?

proto/regen/ecocredit/v1alpha2/types.proto
```diff
@@ -30,18 +30,32 @@ message ClassInfo {
...
+  string project_location = 4;
+}
+
 // BatchInfo represents the high-level on-chain information for a credit batch.
 message BatchInfo {
 
-  // class_id is the unique ID of credit class.
-  string class_id = 1;
+  // project_id is the unique ID of the project this batch belongs to.
+  string project_id = 1;
```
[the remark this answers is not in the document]
technicallyty: doesn't look like we currently have an RPC for that type of query, since the model is changed a bit. Should that type of query be added in this PR?

aaronc:
How about we just choose a regex (something like `[A-Za-z0-9]{3,10}`) for `project_id` and then allow issuers to choose an unique ID for every project within a class?

For verra projects we would create a "verra bridge" credit class and the project ID's would match what's in Verra.

sarahbaxendell:
> I'm not sure if the issuer of the project will be the one issuing all the batches within that project. Projects can run for up to 25 years and if credits are being issued ex-ante that's a pretty long time to commit to being an issuer. I think any issuer of a credit class should be able issue credit batches to a project living in that credit class

Agree. the person requesting a credit to be issued on the registry could be a land owner, land steward, project developer, project proponent or broker, within a program. For longevity, having anyone with credit issuance ability (ie. anyone of 3 people with the permissions to do so, etc.) issue those credits, makes sense. We want to ensure that credit issuers are active parties and that the process is monitored, but doesn't have a 3 month delay like within the Verra systems.

sarahbaxendell:
> Thanks @aaronc !
> 
> I'm in general in favor of having projects as on-chain entities. One thing I'm still not totally sure of is whether that's required for all credit classes though. e.g. for "Verra" credits -- how do we want to handle project_id?
> 
> Few Options I see:
> 
> 1. Make `project_id` optional (or have the "optionality" set at the credit class level) so that for Verra credits we just leave it blank
> 2. Register on-chain entities for projects for 3rd party credit classes. This would have to be done by the issuer, even if we change our data model to have project developers generally managing on-chain project registration, because I assume we cannot assume that all Verra project developers for our brokered credits will be willing to register on-chain entities for their projects in the immediate future.
> 3. Change `project_id` to be namespaced in a way that could resolve either to on-chain entities, or known off-chain entities (I elaborate on one way to do this in a later comment here).
> 
> Separately, there's still the question (which @aaronc and @S4mmyb were already discussing) around who can create projects. IMO I think it's better to have this be permissionless, so that on-chain project entities can be managed independently of a specific issuer (as they may need to be managed over decade-long time spans as @S4mmyb indicated).
> 
> One middle ground that we could have is an apply-to-register model, where anyone (typically a project developer) can apply to register a project under a credit class, but an issuer of the credit class must approve the application in order for the project to be successfully registered on-chain and start receiving credits. We could have an expiry on applications so not to bloat state unnecessarily.

It is unlikely that project developers for a verra credit would engage with this system if we are working with brokers. there are different PD-broker relationships - sometimes the project developer has custody of credits, sometimes broker, sometimes both. 

There are project IDs on Verra - ex: project 981; I would recommend for external credits (non-native to regen registry) we adopt a very simple project ID structure - V-981 (transalting to Verra project 981) or GS-24 (gold standard project 24). 

The apply-to-register model is more akin the the centralized approaches on existing registries. As long as that credit issuer is active, this structure makes sense - a project developer is applying to create a project using a credit class when they submit a compelted project plan.

proto/regen/ecocredit/v1alpha2/tx.proto
sarahbaxendell: > I agree it would make it easier for querying.

i recommend queries at the credit class and methodology level. what are use-cases scenarios of the query function?

sarahbaxendell:
> I thought we already established that verra credits always have project IDs anyway

they all have a project ID number (ex: 981) and a serial number that contains a broader set of data (method, vintage, etc.)

sarahbaxendell:
> > You would reference the verra project ID in the project metadata
> 
> I think the issue here would be that someone could use the project metadata field to store something other than the projectID, such as the project page or link to the project plan, which would then make it more difficult to locate the projectID. I think having an explicit field for projectID when you create a new project makes more sense.

I agree that projectID should be a field autonomous from project documents.

S4mmyb:
Following up on our conversation last week @clevinson @aaronc @sarahbaxendell. We didn't come to any conclusions on how we want to represent projects as on-chain entities, but our conversation was helpful in scoping out some requirements for projects and project IDs. The two main takeaways were that we wanted to: 

- Create the ability to use Project IDs for querying basketing, and indexing credit batches for use in the order-book & credit marketplace, possibly at the credit class level. 
- Create the ability to list projects from existing registries in a way which makes it easy to look up a project from their on-chain references (on a website or similar), in an unambiguous way

Possible solutions we discussed were using namespace identifiers on project IDs to attribute them to a specific registry which would allow us to set filters on credit classes / batches based on the namespaced identifier (such as `regen:project:<regen_project_id>` or `verra:project:<verra_project_id>`).

proto/regen/ecocredit/v1alpha2/query.proto
```diff
@@ -1,62 +1,76 @@
...
   rpc ClassInfo(QueryClassInfoRequest) returns (QueryClassInfoResponse) {
     option (google.api.http).get =
-        "/regen/ecocredit/v1alpha1/classes/{class_id}";
+        "/regen/ecocredit/v1alpha2/classes/{class_id}";
+  }
+
+  // Projects queries for all projects within a class with pagination.
+  rpc Projects(QueryProjectsRequest) returns (QueryProjectsResponse) {
+    option (google.api.http).get =
+        "/regen/ecocredit/v1alpha2/classes";
```
ryanchristo, line 29:
```suggestion
        "/regen/ecocredit/v1alpha2/classes/{class_id}/projects";
```

proto/regen/ecocredit/v1alpha2/tx.proto
robert-zaremba: This is a type to create a new batch. It seams redundant to provide a class here - could be error prone. Each project is assigned to only one class. So having a project id we have its associated class id.

We can add a query method by class id (to return batches by class id) separately.

proto/regen/ecocredit/v1alpha2/types.proto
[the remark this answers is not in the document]
robert-zaremba: Blockchain is not a relational DB, eventually we will use indexers. Hopefully soon. So on the protocol level let's keep the minimum. 
We can always create a module level index without denormalizing proto. Later, once we will have a proper indexer we will remove the module indexes.

robert-zaremba (approved):
lgtm. Let's clean up the stuff with `Info` infix (maybe we can drop it).

proto/regen/ecocredit/v1alpha2/tx.proto
sarahbaxendell: Hi Robert. A carbon credit project can be attached to multiple credit classes. This is true for Verra credits purchased OTC that they link to multiple methodologies. In the case of Regen produced credits, they are usually formatted as primary indicator (credit class) plus cobenefits. However, there is not always a 1:1 project:credit class relationship in the industry, so the credit class reference needs to be coded to accept multiple credit class types per project.

clevinson: While this is true with other registries & how they manage projects, since we don't have the concept of a "registry" as an onchain entity at the moment, its not really possible to have projects defined on a 'per registry basis', which is why in this initial PR, we're defining projects on a 'per credit class' basis.

If we go with @aaronc's proposal of allowing project_id's to be manually set by the creator of a project, then this should work out fine. Different credit classes could have their own issuer list and management process for creating projects under that credit class, and when those projects are created on-chain they could still point to the same "Verra" project, by either having the same metadata, or the same `V-456` project_id, or both.

aaronc: In this model a project can only be associated with one class. If we want to change that then this will need a redesign

proto/regen/ecocredit/v1alpha2/query.proto
```diff
@@ -1,62 +1,76 @@
 syntax = "proto3";
 
-package regen.ecocredit.v1alpha1;
+package regen.ecocredit.v1alpha2;
+
 import "google/api/annotations.proto";
-import "regen/ecocredit/v1alpha1/types.proto";
```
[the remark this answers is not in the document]
aaronc: Hmm... I guess we can call this v1. What do others think? I think that's for a separate PR and we would make the switch right before RC phase

clevinson (approved):
After speaking a bit more with @S4mmyb and @aaronc today here's where I think we currently stand with alignment on the high level API:

There's rough consensus on creating projects being a permissioned action that only issuers of a credit class can take.

There also seems to be alignment on the fact that we want on-chain entities for all projects (whether the credit classes represent native credits, or credits on some 3rd party chain).

Project level metadata can and should be used to reference a link to a project on a 3rd party registry where applicable. Project level metadata was partially omitted in the original PR, but is fixed now so that looks good to me.

In order to allow for the "project_id" field to be set in a more custom way (e.g. to correspond directly to a Verra project), I think @aaronc's suggestion to allow this to be manually set as an y unique alphanumeric value makes sense.

I'm partially leaning towards credit classes defining whether they accept manually set project_id's, as I think in the long run, with most projects being native on-chain, it may make more sense to not have manual `project_id`s as they may be abused as a name registry of projects (e.g. project_id: "coolproject-5`).  But I'm happy to approve this current PR as is and we can add that feature later down the line if it feels valuable to others.

aaronc:
I updated this PR to allow optional user-specified project IDs. These IDs must be alphanumeric and between 2-16 characters long - somewhat arbitrary but seems reasonable.

proto/regen/ecocredit/v1alpha2/query.proto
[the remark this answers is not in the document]
clevinson: I'm in favor of us trying to target v1 proto definitions for regen-ledger 3.0. And agreed that we should probably make that call when we're closer to release and have a bit more clarity on whether or not we expect to still have any significant design upgrades from new requirements from the science team.

PR #610, opened 2021-10-29
feat(x/ecocredit): add basket API

aaronc:
## Description

Adds basket functionality to x/ecocredit.

Currently editing basket criteria is not supported because it needs to be thought through more.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1alpha1/types.proto
```diff
@@ -117,4 +117,62 @@ message CreditTypeSeq {
 
   // The sequence number of classes of the credit type
   uint64 seq_number = 2;
+}
+
+message Filter {
+  oneof sum {
+    And and = 1;
+    Or or = 2;
```
technicallyty, line 125: do we actually need `And` here? i cant really see how that would work in this context

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -427,3 +452,107 @@ message MsgAllowAskDenom {
...
+
+message AdmissionCriteria {
+  Filter filter = 1;
+
+  // multiplier is an integer number which is applied to credit units when
+  // converting to basket units. For example if the multiplier is 2000, then
+  // 1.1 credits will result in 2200 basket tokens. If there are any fractional
+  // amounts left over in this calculation when adding credits to a basket,
+  // those fractional amounts will not get added to the basket.
+  string multiplier = 2;
```
amaurym, line 492: What is this multiplier field used? Is it more appropriate to put it in `MsgCreateBasket` ?

proto/regen/ecocredit/v1alpha1/types.proto
```diff
@@ -117,4 +117,62 @@ message CreditTypeSeq {
...
+    // their project location is more specific, ex. "US-NY 12345". If
+    // a country, sub-national-code and postal code are all provided then
+    // only projects in that postal code will match.
+    string project_location = 10;
+
+    DateRange date_range = 11;
+
+    // tag specifies a curation tag to match against.
+    string tag = 12;
+  }
```
amaurym, line 157: I couldn't `make proto-gen`, I think this `}` is extra

proto/regen/ecocredit/v1alpha1/tx.proto
technicallyty:
```suggestion
  // multiplier is an integer number which is applied to credit units when
  // converting to basket units. For example if the multiplier is 2000, then
  // 1.1 credits will result in 2200 basket tokens. If there are any fractional
  // amounts left over in this calculation when adding credits to a basket,
  // those fractional amounts will not get added to the basket.
  int64 multiplier = 2;
```

should be an int and not string too right?

amaurym: I think I may have understood: credit amounts can be fractional (e.g. `1.1` in the comment), but basket tokens (sdk.Coins) are integers. Multiplier allows to lose less of the decimal parts of ecocredits. Is that correct?

Still think it should be in MsgCreateBasket though

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -52,6 +53,30 @@ service Msg {
...
+  // CreateBasket creates a bank denom which wraps credits.
+  rpc CreateBasket(MsgCreateBasket) returns (MsgCreateBasketResponse);
+
+  // AddToBasket adds credits to a basket in return for basket tokens.
+  rpc AddToBasket(MsgAddToBasket) returns (MsgAddToBasketResponse);
+
+  // TakeFromBasket takes credits from a basket without regard for which
+  // credits they are. The credits will be auto-retired if retire_on_take
+  // is true. Credits will be chosen randomly using the previous block hash
+  // as a consensus source of randomness.
```
blushi, line 66: @aaronc in terms of implementation, would the following work:
- take the previous block hash and convert it to uint64
- given the total number of different credits within the basket `n`, the first credits that will get picked correspond to: hash modulo n (in terms of order)
- then if we need to take more credits, we get some from the next one and so on

proto/regen/ecocredit/v1alpha1/types.proto
aaronc: `And` is needed to combine multiple options (for instance class + data range) within a single filter.

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: > should be an int and not string too right?

Well in cosmos we generally use big integers and those are serialized as strings in protobuf

aaronc: > Ah I think I may have understood: credit amounts can be fractional (e.g. `1.1` in the comment), but basket tokens (sdk.Coins) are integers. Multiplier allows to lose less of the decimal parts of ecocredits. Is that correct?
> 
> Still think it should be in MsgCreateBasket though. We can then delete the comment
> 
> > the first criteria which applies to a credit will determine its multiplier in the basket.

It is only used in `MsgCreateBasket`. This design allows different multipliers for different types of credits within a basket. For instance, someone could say XYZ credit is really 80% as valuable as ABC credit and give them different multipliers within the same basket.

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: That makes sense 👍 . Feel free to suggest an update to the docs to specifiy this more clearly

proto/regen/ecocredit/v1alpha1/types.proto
technicallyty: spoke with cory and i think i just misunderstood the design, hence my additional [curly suggestion](https://github.com/regen-network/regen-ledger/pull/610/files/318d9d97d77a069e82e74298e0643c61d5ad8bcb#r740417038), and this [issue](https://github.com/regen-network/regen-ledger/pull/610/files/318d9d97d77a069e82e74298e0643c61d5ad8bcb#r741916566).

proto/regen/ecocredit/v1alpha1/types.proto
```diff
@@ -117,4 +117,62 @@ message CreditTypeSeq {
 
   // The sequence number of classes of the credit type
   uint64 seq_number = 2;
+}
+
+message Filter {
+  oneof sum {
+    And and = 1;
+    Or or = 2;
+  }
```
technicallyty, line 126:
```suggestion
message Filter {
  oneof sum {
    And and = 1;
    Or or = 2;
```

technicallyty: misunderstood the intent, this should be reverted back

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: renamed `AdmissionCriteria` to `BasketCriteria`. Is that any clearer?

technicallyty (approved):

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -427,3 +452,107 @@ message MsgAllowAskDenom {
...
+  // of 6 will mean that 10^6 units of a basket token should be displayed
+  // as one unit in user interfaces.
+  uint32 exponent = 4;
+
+  // basket_criteria is the criteria by which credits can be added to the
+  // basket. Basket criteria will be applied in order and the first criteria
+  // which applies to a credit will determine its multiplier in the basket.
+  repeated BasketCriteria basket_criteria = 5;
+
+  bool retire_on_take = 6;
```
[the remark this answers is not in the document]
aaronc: Sure 👍. Retirement makes sense as the default

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -52,6 +53,30 @@ service Msg {
 
   // AllowAskDenom is a governance operation which authorizes a new ask denom to be used in sell orders
   rpc AllowAskDenom (MsgAllowAskDenom) returns (MsgAllowAskDenomResponse);
+
+  // CreateBasket creates a bank denom which wraps credits.
```
amaurym, line 57:
```suggestion
  // CreateBasket creates a bank denom which wraps credits.
  // Baskets are managed and created by Regen Ledger governance (REGEN holders).
  // This means that basket creation and basket composition updates goes through
  // governance processes and module parameters.
```

adding 578bbbdc24a28cd8f0cfb11ea9c3f137c2a5358c

amaurym (approved):
pre-approving, pending on https://github.com/regen-network/regen-ledger/pull/610#discussion_r742161190 and https://github.com/regen-network/regen-ledger/pull/610#discussion_r746245166

proto/regen/ecocredit/v1alpha1/tx.proto
aaronc: But how is governance controlling this? Don't we need to modify the API.

robert-zaremba (approved):

Issue #602, opened 2021-10-27
Add proto query definitions for ecocredit marketplace functionality


ryanchristo:
## Summary

Ref: #505

Add proto query definitions for ecocredit marketplace functionality.

## Proposal

Add query definitions for the following queries:

- all sell orders
- get sell order (for a sell order id)
- sell orders by batch denom
- sell orders by address
- all buy orders
- buy order by buy order id
- all allowed asking denoms

```
  // SellOrders queries for all sell orders with pagination.
  rpc SellOrders(QuerySellOrdersRequest) returns (QuerySellOrdersResponse) {
    option (google.api.http).get =
        "/regen/ecocredit/v1alpha1/sell-orders";
  }

  // SellOrder queries for all sell orders with pagination.
  rpc SellOrder(QuerySellOrderRequest) returns (QuerySellOrderResponse) {
    option (google.api.http).get =
        "/regen/ecocredit/v1alpha1/sell-orders/{sell_order_id}";
  }

...

```
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #613, opened 2021-11-02
feat(x/ecocredit): Msg Server implementation for buy/sell functionality

clevinson:
## Description

Closes: #603 #614 #622

This pull request implements the following:

- proto type definitions for `SellOrder`, `BuyOrder`, and `AskDenom`
- proto event definitions for `EventSell`, `EventUpdateSellOrder`, `EventBuyOrderCreated`, `EventBuyOrderFilled`, and `EventAllowAskDenom`
- message service methods for `Sell`, `UpdateSellOrders`, `Buy`, and `AllowAskDenom`
- `ValidateBasic` for `MsgSell`, `MsgUpdateSellOrders`, `MsgBuy`, and `MsgAllowAskDenom`

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] followed the guidelines for [building modules](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules)
- [ ] included the necessary unit and integration [tests](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#testing)
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed `!` in the type prefix if API or client breaking change
- [ ] confirmed all author checklist items have been addressed 
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

PR #632, opened 2021-11-11
feat: marketplace Query API

technicallyty:
## Description

* Adds queries for sell/buy orders and denoms.
* Refactors out the previously nested Sell/Buy order messages into their own standalone messages in types.proto to reuse them in the query messages. 

Closes: #602 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1alpha1/types.proto
```diff
@@ -117,4 +117,100 @@ message CreditTypeSeq {
...
+}
+
+// BuyOrder is the content of a new buy order.
+message BuyOrder {
+
+  // address is the address of the buyer
+  optional string address = 1;
+
+  // buy_order_id is the id of the buy order
+  optional uint64 buy_order_id = 2;
```
technicallyty, line 159: not sure if this is the best way to do this. wanted to avoid duplicating BuyOrder and SellOrder messages for query/tx types. Doing this allows us to reuse and optionally fill in these values when its used in a query. When used as a tx, there is only one address at the top level used as the signer. 

Another option is to remove these optionals, and include an additional repeated string and repeated uint64 who's index will match that of the repeated sell/buy orders in queries.

buf.yaml
```diff
@@ -10,6 +10,7 @@ lint:
   except:
     - SERVICE_SUFFIX
     - RPC_REQUEST_STANDARD_NAME
+    - RPC_REQUEST_RESPONSE_UNIQUE
```
technicallyty, line 13: added this line here to prevent the linter from complaining about reusing response types. imo doesn't make sense to have unique response types for rpc that return the exact same structure

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -58,6 +59,127 @@ service Query {
     option (google.api.http).get =
         "/regen/ecocredit/v1alpha1/params";
   }
+
+  // SellOrders queries a paginated list of all sell orders
+  rpc SellOrders(QuerySellOrdersRequest) returns (QuerySellOrdersResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/marketplace/sell-orders";
```
aleem1314, line 65: Can we use a single query for `SellOrders` and `SellOrder`, and make `sell-order-id` optional parameter?

buf.yaml
aaronc: let's not do this sorry... https://docs.buf.build/lint/rules#rpc_request_standard_name-rpc_response_standard_name-rpc_request_response_unique

aaronc (dismissed):

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -432,15 +363,8 @@ message MsgAllowAskDenom {
...
-  string denom = 2;
-
-  // display_denom is the denom to display to the user and is informational
-  string display_denom = 3;
-
-  // exponent is the exponent that relates the denom to the display_denom and is
-  // informational
-  uint32 exponent = 4;
+  // denom_info contains coin denom information.
+  Denom denom_info = 2;
```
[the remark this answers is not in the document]
ryanchristo: This could also be `DenomInfo` but I think `AskDenom` is more consistent with the current API.

proto/regen/ecocredit/v1alpha1/types.proto
ryanchristo: Why would we make `address` and `buy_order_id` optional? These should both be required in the transaction message and included in the query response.

ryanchristo (changes requested):
Looking good!

The types are already included in #613. It might have made more sense to use that branch as the base rather than using master. Can you rebase off that branch?
 
It would be nice to keep the query methods in the same order as they are listed within the `Query` service. We do this in all of the other proto files for modules, which makes it easier to look up methods.

proto/regen/ecocredit/v1alpha1/types.proto
technicallyty: so address was made optional as the `MsgBuy` structure has the buyer address at the top level, and not in the `Order` message. I wanted to reuse the `Order` structure in the query, while being able to inject address for query purposes. 

perhaps the better way is either to make a new order structure for queries or reuse the old one and just have another repeated field who's index matches the index of the `Order` msg

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -58,6 +59,127 @@ service Query {
...
+  }
+
+  // SellOrder queries a sell order by its ID
+  rpc SellOrder(QuerySellOrderRequest) returns (QuerySellOrderResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/marketplace/sell-orders/id/{sell_order_id}";
```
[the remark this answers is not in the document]
technicallyty: wouldn't this break REST? im not sure it can distinguish which query is being called by disambiguating whether its an id, address, or denom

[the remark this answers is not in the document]
ryanchristo: Ah yes, I think you're right. In that case, we should leave it as is. :+1:

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -177,3 +212,129 @@ message QueryCreditTypesResponse {
...
+
+  // pagination defines an optional pagination for the request.
+  cosmos.base.query.v1beta1.PageRequest pagination = 2;
+}
+
+// QuerySellOrdersByDenomResponse is the Query/SellOrdersByDenom response type.
+message QuerySellOrdersByDenomResponse {
+
+  // seller_addresses contain addresses associated with sell orders in the sell_orders array
+  repeated string seller_addresses = 1;
```
ryanchristo, line 263: Same as above. Is this field necessary?

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -177,3 +212,129 @@ message QueryCreditTypesResponse {
...
+message QueryBuyOrdersRequest {
+  // pagination defines an optional pagination for the request.
+  cosmos.base.query.v1beta1.PageRequest pagination = 1;
+}
+
+// QueryBuyOrdersResponse is the Query/BuyOrders response type.
+message QueryBuyOrdersResponse {
+
+  // buyer_addresses contain addresses associated with buy orders in the buy_orders array
+  repeated string buyer_addresses = 1;
```
ryanchristo, line 301: Each `BuyOrder` should include the `buyer`. It looks like this was overlooked in #613. I think it would be preferred to add the buyer address to the buy order rather than having a separate field in the query response.

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -177,3 +212,129 @@ message QueryCreditTypesResponse {
...
+message QueryBuyOrderRequest {
+  // buy_order_id is the id of the buy order.
+  uint64 buy_order_id = 1;
+}
+
+// QueryBuyOrderResponse is the Query/BuyOrder response type.
+message QueryBuyOrderResponse {
+
+  // buyer_address is the address of the creator of the buy order
+  string buyer_address = 1;
```
ryanchristo, line 320: Same as above. We should include the `buyer` in the `BuyOrder` and consider removing this field.

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -296,31 +297,7 @@ message MsgSell {
...
-    // batch_denom. Each credit unit of the batch will be sold for at least the
-    // ask_price or more.
-    cosmos.base.v1beta1.Coin ask_price = 3;
-
-    // disable_auto_retire disables auto-retirement of credits which allows a
-    // buyer to disable auto-retirement in their buy order enabling them to
-    // resell the credits to another buyer.
-    bool disable_auto_retire = 4;
-  }
+  repeated SellOrder orders = 2;
```
ryanchristo, line 300: I'm not sure if we should remove `Order` and replace it with `SellOrder`. `SellOrder` includes `owner`, which is also included in `MsgSell` and we should avoid duplicating the `owner` field. Thoughts?

proto/regen/ecocredit/v1alpha1/tx.proto
```diff
@@ -367,54 +344,8 @@ message MsgBuy {
...
-    // to complete the purchase.
-    cosmos.base.v1beta1.Coin bid_price = 3;
-
-    // disable_auto_retire allows auto-retirement to be disabled. If it is set to true
-    // the credits will not auto-retire and can be resold assuming that the
-    // corresponding sell order has auto-retirement disabled. If the sell order
-    // hasn't disabled auto-retirement and the buy order tries to disable it,
-    // that buy order will fail.
-    bool disable_auto_retire = 4;
+  repeated BuyOrder orders = 2;
```
ryanchristo, line 347: I think this works as is but as mentioned above, I think we actually need to add an `owner` or `buyer` field to `BuyOrder` and then we should support the ability to query buy orders by address. This should have been done in #613 but maybe we can add that here. Thoughts?

proto/regen/ecocredit/v1alpha1/types.proto
ryanchristo: I think we should avoid making them optional for the purpose of reusing code. I think each of the query responses should return `SellOrder` and `BuyOrder` as they are defined in `types.proto` (`BuyOrder` should include the buyer address, which I overlooked in #613) and the `Orders` within `MsgSell` and `MsgBuy` should stay the same.

proto/regen/ecocredit/v1alpha1/tx.proto
technicallyty: yeah i think it makes sense to embed address into buy order. I'll add it in this PR

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -58,6 +58,41 @@ service Query {
...
+
+  // BuyOrders querires a paginated list of all buy orders
+  rpc BuyOrders(QueryBuyOrdersRequest) returns (QueryBuyOrdersResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/buy-orders";
+  }
+
+  // BuyOrder queries a buy order by its id
+  rpc BuyOrder(QueryBuyOrderRequest) returns (QueryBuyOrderResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/buy-orders/id/{buy_order_id}";
+  }
```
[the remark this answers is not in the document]
technicallyty: added 👍🏻

proto/regen/ecocredit/v1alpha1/tx.proto
technicallyty: yeah i'm not sure what would be the best path forward here. imo i think the big sacrifice is having to manage two structures (if we update SellOrder down the road, we're gonna have to remember to update Order as well). the downside to having both is message size + clunky api consuming, devs will have to embed address in each repeated Sell/BuyOrder

technicallyty: talked with Cory, decided its best to have the duplicate structure. i'll revert the changes and put the Order message back in

technicallyty:
would it be useful to have a query for buy orders based on the denom of the bid price?

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -58,6 +58,51 @@ service Query {
...
+
+  // SellOrders queries a paginated list of all sell orders
+  rpc SellOrders(QuerySellOrdersRequest) returns (QuerySellOrdersResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/sell-orders";
+  }
+
+  // SellOrdersByDenom queries a paginated list of all sell orders of a specific denom
+  rpc SellOrdersByDenom(QuerySellOrdersByDenomRequest) returns (QuerySellOrdersByDenomResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/sell-orders/denom/{denom}";
+  }
```
ryanchristo, line 75: `denom` might be a bit ambiguous and maybe we should explicitly call this `batch_denom` (the denom for the credit batch, not the denom for the ask price).

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -58,6 +58,51 @@ service Query {
...
+
+  // BuyOrders querires a paginated list of all buy orders
+  rpc BuyOrders(QueryBuyOrdersRequest) returns (QueryBuyOrdersResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/buy-orders";
+  }
+
+  // BuyOrdersByDenom queries a paginated list of buy orders by denom
+  rpc BuyOrdersByDenom(QueryBuyOrdersByDenomRequest) returns (QueryBuyOrdersByDenomResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/buy-orders/denom/{denom}";
+  }
```
ryanchristo, line 95: > would it be useful to have a query for buy orders based on the denom of the bid price?

`batch_denom` is not included in a buy order so this would be for the bid price denom. I think we can either remove this or we could make it more explicit (i.e. `bid_denom`). If we have a query for buy orders by bid price denom, we should probably have a query for sell orders by ask price denom.

ryanchristo (commented):
This looks great! A couple more comments.

> would it be useful to have a query for buy orders based on the denom of the bid price?

I could see it being useful to query buy orders by bid price denom and sell orders by ask price denom. It looks like you already have a query for this...? See comments below.

proto/regen/ecocredit/v1alpha1/query.proto
technicallyty: ok ill just leave it out for now, can always add it later down the road if it seems useful

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -58,6 +58,46 @@ service Query {
...
+
+  // SellOrders queries a paginated list of all sell orders
+  rpc SellOrders(QuerySellOrdersRequest) returns (QuerySellOrdersResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/sell-orders";
+  }
+
+  // SellOrdersByDenom queries a paginated list of all sell orders of a specific ecocredit denom
+  rpc SellOrdersByBatchDenom(QuerySellOrdersByBatchDenomRequest) returns (QuerySellOrdersByBatchDenomResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/sell-orders/denom/{batch_denom}";
```
ryanchristo, line 74: Should we be explicit about `batch-denom` in the URL as well?

```suggestion
    option (google.api.http).get = "/regen/ecocredit/v1alpha1/sell-orders/batch-denom/{batch_denom}";
```

proto/regen/ecocredit/v1alpha1/query.proto
```diff
@@ -58,6 +58,46 @@ service Query {
...
+  }
+
+  // BuyOrdersByAddress queries a paginated list of buy orders by creator address
+  rpc BuyOrdersByAddress(QueryBuyOrdersByAddressRequest) returns (QueryBuyOrdersByAddressResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/buy-orders/address/{address}";
+  }
+
+  // AllowedAskDenoms queries all denoms allowed to be set in the AskPrice of a sell order
+  rpc AllowedAskDenoms(QueryAllowedAskDenomsRequest) returns (QueryAllowedAskDenomsResponse) {
+    option (google.api.http).get = "/regen/ecocredit/v1alpha1/denoms";
```
ryanchristo, line 99: Should we be explicit about `ask-denoms` in the URL?

```suggestion
    option (google.api.http).get = "/regen/ecocredit/v1alpha1/ask-denoms";
```

ryanchristo (approved):
Nice work! This looks good to me. A couple suggestions but nothing blocking.

aleem1314 (approved):

technicallyty:
@aaronc can you give another review or unblock? 🙏🏻

PR #662, opened 2021-12-02
feat(x/ecocredit): add marketplace cli

ryanchristo:
## Description

Closes: #615

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/client/testsuite/query.go
```diff
@@ -485,3 +487,557 @@ func (s *IntegrationTestSuite) TestQueryParams() {
...
+			args:      []string{"foo"},
+			expErr:    true,
+			expErrMsg: "Error: accepts 0 arg(s), received 1",
+		},
+		{
+			name:      "valid",
+			args:      []string{},
+			expErr:    false,
+			expErrMsg: "",
+			expOrders: []*ecocredit.SellOrder{
```
technicallyty, line 576: could make these a top level constant so we dont have to copy pasta this sell order slice

x/ecocredit/client/testsuite/tx.go
```diff
@@ -1373,3 +1393,417 @@ func (s *IntegrationTestSuite) TestTxUpdateIssuers() {
...
+			name:      "too many args",
+			args:      []string{"foo", "bar"},
+			expErr:    true,
+			expErrMsg: "accepts 1 arg(s), received 2",
+		},
+		{
+			name: "missing batch denom",
+			args: append(
+				[]string{
+					"[{quantity: \"5\", ask_price: \"100regen\", disable_auto_retire: false}]",
```
technicallyty, line 1424: might be a little cleaner/straightforward to just json.Marshal this using the `encoding/json` package.

i dont think this works for something like MsgSell as it uses coin for AskPrice, but we could just make a mirrored test struct changing that field to a string. 

not sure if that change would be all that valuable, so interested on seeing thoughts on this

x/ecocredit/client/tx.go
```diff
@@ -504,3 +509,224 @@ Parameters:
...
+				askPrice, err := sdk.ParseCoinNormalized(order.AskPrice)
+				if err != nil {
+					return err
+				}
+
+				// set order with ask price as sdk.Coin
+				orders[i] = &ecocredit.MsgSell_Order{
+					BatchDenom:        order.BatchDenom,
+					AskPrice:          &askPrice,
+					Quantity:          order.Quantity,
```
technicallyty, line 564: we should verify denom and quantity

x/ecocredit/client/tx.go
```diff
@@ -504,3 +509,224 @@ Parameters:
...
+				askPrice, err := sdk.ParseCoinNormalized(update.NewAskPrice)
+				if err != nil {
+					return err
+				}
+
+				// set update with new ask price as sdk.Coin
+				updates[i] = &ecocredit.MsgUpdateSellOrders_Update{
+					SellOrderId:       sellOrderId,
+					NewAskPrice:       &askPrice,
+					NewQuantity:       update.NewQuantity,
```
technicallyty, line 638: we should check this value

x/ecocredit/client/tx.go
```diff
@@ -504,3 +509,224 @@ Parameters:
...
+				bidPrice, err := sdk.ParseCoinNormalized(order.BidPrice)
+				if err != nil {
+					return err
+				}
+
+				// set order with new bid price as sdk.Coin
+				orders[i] = &ecocredit.MsgBuy_Order{
+					Selection:         selection,
+					BidPrice:          &bidPrice,
+					Quantity:          order.Quantity,
```
technicallyty, line 717: verify here

x/ecocredit/client/query.go
```diff
@@ -223,3 +232,189 @@ $%s q %s params
...
+			c, ctx, err := mkQueryClient(cmd)
+			if err != nil {
+				return err
+			}
+			pagination, err := client.ReadPageRequest(cmd.Flags())
+			if err != nil {
+				return err
+			}
+			res, err := c.SellOrdersByAddress(cmd.Context(), &ecocredit.QuerySellOrdersByAddressRequest{
+				Address:    args[0],
```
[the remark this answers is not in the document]
ryanchristo: We validate in `SellOrdersByAddress` already and return the error here if invalid. No need to validate both here and in the query implementation itself. See other query commands for example.

x/ecocredit/client/query.go
```diff
@@ -223,3 +232,189 @@ $%s q %s params
...
+			c, ctx, err := mkQueryClient(cmd)
+			if err != nil {
+				return err
+			}
+			pagination, err := client.ReadPageRequest(cmd.Flags())
+			if err != nil {
+				return err
+			}
+			res, err := c.SellOrdersByBatchDenom(cmd.Context(), &ecocredit.QuerySellOrdersByBatchDenomRequest{
+				BatchDenom: args[0],
```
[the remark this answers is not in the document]
ryanchristo: We validate in `SellOrdersByBatchDenom` already and return the error here if invalid. No need to validate both here and in the query implementation itself. See other query commands for example.

x/ecocredit/client/query.go
```diff
@@ -223,3 +232,189 @@ $%s q %s params
...
+			res, err := c.BuyOrders(cmd.Context(), &ecocredit.QueryBuyOrdersRequest{
+				Pagination: pagination,
+			})
+			return print(ctx, res, err)
+		},
+	})
+}
+
+// QueryBuyOrdersByAddressCmd returns a query command that retrieves all buy orders by address with pagination.
+func QueryBuyOrdersByAddressCmd() *cobra.Command {
```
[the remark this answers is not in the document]
ryanchristo: Fixed in https://github.com/regen-network/regen-ledger/pull/662/commits/7b0fabe694183b86062e2ed20f721922beca00a5

x/ecocredit/client/testsuite/query.go
ryanchristo: Added it to the test suite struct as it relies on val address https://github.com/regen-network/regen-ledger/pull/662/commits/7b0fabe694183b86062e2ed20f721922beca00a5

x/ecocredit/client/tx.go
ryanchristo: We already do in the message implementation. I don't think we need to duplicate the verification here. See other tx commands for example and test cases within this pull request.

x/ecocredit/client/tx.go
ryanchristo: We already do in the message implementation. I don't think we need to duplicate the verification here. See other tx commands for example and test cases within this pull request.

x/ecocredit/client/tx.go
ryanchristo: We already do in the message implementation. I don't think we need to duplicate the verification here. See other tx commands for example and test cases within this pull request.

x/ecocredit/client/query.go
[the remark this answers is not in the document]
technicallyty: ah i see that now. perhaps we shouldnt though 🤔 i think there was a concern in another thread by robert about validating before spinning out a TX so we dont send bogus transactions that we could verify beforehand. 

maybe queries should get the same treatment? 🤷🏻

x/ecocredit/client/testsuite/tx.go
ryanchristo: Hmmm.... I might be misunderstanding what you're proposing but I don't think we need to create a separate struct and marshal it for the arguments within the test cases. I was following `Send`, `Cancel`, and `Retire` as examples. I think the argument here as a string is important as this is the same format we are expecting users to provide.

x/ecocredit/client/query.go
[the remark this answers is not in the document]
ryanchristo: Perhaps we should. Let's chat about this on our next call.

x/ecocredit/server/server.go
```diff
@@ -102,7 +102,7 @@ func newServer(storeKey sdk.StoreKey, paramSpace paramtypes.Subspace,
 	}
 	s.sellOrderByBatchDenomIndex, err = orm.NewIndex(sellOrderTableBuilder, SellOrderByBatchDenomIndexPrefix, func(value interface{}) ([]interface{}, error) {
 		denom := value.(*ecocredit.SellOrder).BatchDenom
-		return []interface{}{denom}, nil
+		return []interface{}{[]byte(denom)}, nil
 	}, []byte{})
```
blushi, line 106: we could also just use a string-based index since `denom` is a string

blushi (approved):
utACK, just one small comment.

x/ecocredit/server/server.go
ryanchristo: When attempting to query using REST, I was getting a panic from the response.

```
panic: unknown field "code" in ecocredit.QuerySellOrdersByBatchDenomResponse
```

Switching to bytes solved the issue. I thought maybe the format of the denom (use of `-`) was causing issues but never got to the bottom of the issue. Any ideas?

ryanchristo: When attempting to query using REST, I was getting a panic from the response.

```
panic: unknown field "code" in ecocredit.QuerySellOrdersByBatchDenomResponse
```

Switching to bytes solved the issue. I thought maybe the format of the denom (use of `-`) was causing issues but never got to the bottom of it. Any ideas?

blushi: this is probably because the last parameter was set to `[]byte{}` while you were returning `denom` as a string within the interface slice, those should be of the same underlying type

technicallyty (approved):

Issue #621, opened 2021-11-04
Handle auto-retire in ecocredit buy/sell functionality


clevinson:
- [ ] Add `retirement_location` to `MsgBuy` so that when auto-retire is set to true we know where to retire credits to
- [ ] Update msg server implementation to handle transfer of retired credits

Additionally, in #613 we refactored out a new method for `sendEcocredits` that takes in a `MsgSend_SendCredits` as a parameter. It might make sense to refactor `MsgSend_SendCredits` into its own proto message in `ecocredits/types.proto`, and change its API slightly? Let's evaluate this when we tackle this issue.

ryanchristo:
Unable to add `retirement_location` without resolving #646 first.

PR #670, opened 2021-12-10
feat(x/ecocredit): auto-retire credits on purchase

ryanchristo:
## Description

Closes: #621

Auto-retire credits when processing a buy order if auto-retire is enabled for both sell and buy order and return an error if auto-retire is enabled for a sell order but disabled for a buy order.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


ryanchristo:
pending #674

blushi:
@ryanchristo is this r4r again?

x/ecocredit/server/msg_server.go
```diff
@@ -777,12 +777,24 @@ func (s serverImpl) Buy(goCtx context.Context, req *ecocredit.MsgBuy) (*ecocredi
...
+			}
+
+			// set tradable or retired amount depending on auto-retire
+			if sellOrder.DisableAutoRetire && order.DisableAutoRetire {
+				credit.RetiredAmount = "0"
+				credit.TradableAmount = creditsToReceive.String()
+			} else {
+				credit.RetiredAmount = creditsToReceive.String()
+				credit.RetirementLocation = order.RetirementLocation
+				credit.TradableAmount = "0"
```
technicallyty, line 797: should we fail the buy order here in the case of a sell order with auto-retirement enabled and the buy order has it disabled? 

seems like the direction we wanted to go based on [comment](https://github.com/regen-network/regen-ledger/blob/ryan/621-buy-auto-retire/proto/regen/ecocredit/v1alpha2/types.proto#L202-L207)

ryanchristo: There's a condition on lines 780-783 that throws an error if enabled for the sell order and disabled for the buy order.

The current logic is as follows:
- disabled for sell order, disabled for buy order, successfully receive tradable credits
- enabled for sell order, enabled for buy order, successfully receive retired credits
- enabled for sell order, disabled for buy order, fail with error
- disabled for sell order, enabled for buy order, successfully receive retired credits

ryanchristo:
> @ryanchristo is this r4r again?

Yes, this pull request is ready for review.

x/ecocredit/server/msg_server.go
ryanchristo: There's also a test for this `name: "auto-retire required for sell order"`

proto/regen/ecocredit/v1alpha2/tx.proto
```diff
@@ -475,6 +475,10 @@ message MsgBuy {
     // disable_partial_fill disables the default behavior of partially filling
     // buy orders if the requested quantity is not available.
     bool disable_partial_fill = 5;
+
+    // retirement_location is the optional retirement location for the credits
+    // which will be used only if disable_auto_retire is false.
+    string retirement_location = 6;
```
blushi, line 481: we should probably add some validation check related to this location in `MsgBuy.ValidateBasic()`

robert-zaremba: I think the `ValidateBasic` should be strictly minimal . User should pay for all possible operations, and this is only possible in `DeliverTx`. hence, it's better to have 2 methods:
* ValidateBasic
* Validate -- which is run at the beginning of Msg handler / RPC method, and can use state information

x/ecocredit/server/msg_server.go
```diff
@@ -777,12 +777,24 @@ func (s serverImpl) Buy(goCtx context.Context, req *ecocredit.MsgBuy) (*ecocredi
 				return nil, err
 			}
 
+			// error if auto-retire is required for given sell order
+			if !sellOrder.DisableAutoRetire && order.DisableAutoRetire {
+				return nil, ecocredit.ErrInvalidBuyOrder.Wrapf("auto-retire required for sell order %d", sellOrder.OrderId)
```
robert-zaremba, line 782:
```suggestion
				return nil, ecocredit.ErrInvalidBuyOrder.Wrapf("auto-retire is required for sell order %d", sellOrder.OrderId)
```

proto/regen/ecocredit/v1alpha2/tx.proto
ryanchristo: Added a check to `ValidateBasic` that uses `ValidateLocation`. If a retirement location is provided, `ValidateBasic` will check if the string matches the location format.

I'm not sure I understand the purpose of the second method you are proposing @robert-zaremba. Why do we need to use state information to check the location again? The logic for setting the retirement amount and retirement location is already implemented. Is there another check that I am missing?

ryanchristo: See https://github.com/regen-network/regen-ledger/pull/670#discussion_r779768975

x/ecocredit/server/msg_server.go
```diff
@@ -777,12 +777,24 @@ func (s serverImpl) Buy(goCtx context.Context, req *ecocredit.MsgBuy) (*ecocredi
 				return nil, err
 			}
 
+			// error if auto-retire is required for given sell order
+			if !sellOrder.DisableAutoRetire && order.DisableAutoRetire {
```
technicallyty, line 781: i think it would be nice to add another quick check:

```go
if order.RetirementLocation == "" && sellOrder.DisableAutoRetire {
    return someError
}
```

ryanchristo: Good catch but retirement location is optional.

```
    // retirement_location is the optional retirement location for the credits
    // which will be used only if disable_auto_retire is false.
```

technicallyty: but if sellOrder has auto retire enabled, then the retirement location would be required

ryanchristo: I might have misunderstood. I was reading this as retirement location is always optional. If sell order has auto-retire enabled, the retirement location will be stored either as an empty string or a valid location. An empty location is not valid when executing `MsgRetire` so I think you are right. I'll add a check, thanks!

technicallyty (approved):
looks good!

aleem1314 (approved):
lgtm

atheeshp (approved):

Issue #683, opened 2022-01-10
ecocredit: sendCredits method fails on empty amount


technicallyty:
## Summary of Bug

The `sendCredits` method will error out if tradable or retired amounts are left as empty strings. The Dec package does not account for empty strings, nor does `sendCredits.

`Send`'s validate basic method currently does ensure the msg contains a dec compatible string, however we have code in the msg server that calls `sendEcocredits` directly (will not go through validate basic). and the basket methods currently are setup to as well.

We should either

- Update Dec package to account for empty strings (`""` -> `"0"`)
- Update the sendEcocredits method to account for empty strings (`""` -> `"0"`)

## Version

Master

## Steps to Reproduce

sendEcocredits and leave either `TradeableAmount` or `RetiredAmount` empty. 

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ryanchristo:
Dec package updated to account for empty strings in https://github.com/regen-network/regen-ledger/pull/804

Issue #634, opened 2021-11-15
Implement Baskets MsgServer


blushi:
## Summary

Based on the msg service definition from #610 and initial work from Lisbon sprint https://github.com/regen-network/regen-ledger/pull/618, add MsgServer implementation for baskets in `x/ecocredit`

Adding unit tests on the MsgServer will also be included in this PR.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

clevinson:
@technicallyty We've had several product conversations over the past few days this week that may impact the work you're doing on this issue.

For now, let's keep the MsgServer implementation focused on the core logic (storing basket tokens in x/bank, tracking balances in x/ecocredit, etc.). When it comes to implementing the filter criteria for baskets, let's prioritize filtering by `class_id` and `credit_type_name`, and leave the logical operators (AND / OR) + the rest of the filter criteria fields unimplemented for now -- until we have more clarity on the requirements.


Pick & Take from basket functionality should be good to implement as well.

technicallyty:
i have a draft PR up now. linked it to this issue. reposting the questions here for discussion:

### Questions
- Should we allow basket's to have no filter?
- TakeFromBasket only returns tradable amounts, but this method can also result in credits being retired. should we include a `RetiredAmount` in the response message?
- Say someone submits a TakeFromBasket tx with a requested amount of 3.123456789, and one of the batches in the basket has a precision of 6. What do we do in this case? 
- Should we cache the total number of credits in the basket table in ORM? this could save us a lot of computation time in `TakeFromBasket` where we iterate over all the coins, and fail if there aren't enough for the swap. cacheing the amount would allow us to fail early if a user requests more than the basket has.
- Should the basket curator be allowed to pick if picking is disabled?
- Should filters have depth limits? (to stop people from doing stuff like Filter_And[Filter_And[Filter_and[Filter_And[Filter_and...]]]])
- What is the basket display name for? how should it be validated?
- should we allow basket exponents of 0? this would result in a 1:1 of basket token -> basket credit.
- currently i loop over all the batches in the basket, then sort by start date. is this the best/only way?
- How should the basket's coin denom be formed? currently i just use the basket name, but i feel it could be abused by people naming their basket OSMO, and tricking folks. Perhaps we could prefix basket coins with `bskt` so we would have `bsktFooBasket`?

[...]

PR #686, opened 2022-01-12
feat: basket msg server

technicallyty:
## Description

Adds baskets

#### ~~Basket Addresses~~
~~Baskets now derive an address upon creation. This allows us to use our existing ecocredit query endpoints without having to add basket specific queries. Need to know the balance of a basket? call the existing `Balance` endpoint with the basket's address.~~

~~This may not be the most ideal solution as a potential attack vector would be to send credits to this address using the Send method, effectively bypassing the filter. we could give basket addresses a custom prefix `regenbskt1v32....` and have `Send` actively check for this. not too sure if that's ideal, so curious to hear thoughts on it.~~

#### Basket Coin Denom
Currently, I use the basket name as the token denom alone. This is placeholder for a better solution. The name is used as the coin denom for the basket's token. 

To avoid being confused with other tokens on regen network, i think basket tokens should have a prefix before the basket name as their token name. Similar `wBTC` for wrapped bitcoin, we could do something like `bsktFooBasket` for a token from `FooBasket`. This would be good for UX, especially in the case where a troublemaker might make a basket called `OSMO` and try to fool people with the basket token.

### Questions
- Should we allow basket's to have no filter?
- TakeFromBasket only returns tradable amounts, but this method can also result in credits being retired. should we include a `RetiredAmount` in the response message?
- Say someone submits a TakeFromBasket tx with a requested amount of 3.123456789, and one of the batches in the basket has a precision of 6. What do we do in this case? 
- Should we cache the total number of credits in the basket table in ORM? this could save us a lot of computation time in `TakeFromBasket` where we iterate over all the coins, and fail if there aren't enough for the swap. cacheing the amount would allow us to fail early if a user requests more than the basket has.
- Should the basket curator be allowed to pick if picking is disabled?
- Should filters have depth limits?
- What is the basket display name for? how should it be validated?
- should we allow basket exponents of 0? this would result in a 1:1 of basket token -> basket credit.
- currently i loop over all the batches in the basket, then sort by start date. is this the best/only way?

## Stuff that still needs to be implemented:
- filter that allows you to simply set the filter to `batches no older than 3 years ago`
- Events
- methods to update basket criteria
- basket creation fee gov param

Closes: #634 
Closes: #683 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


ryanchristo:
Some notes from our call in case you missed something:

Create basket:
- Basket criteria should be enforced.
- Maybe display name shouldn't be included in request. Maybe auto-generated from name and exponent (i.e. E6, E9). Display name would go into denom metadata. Bank denom metadata needs to be set.

Add Basket:
- Using depth adds complexity. Filter should be a single filter, not an array. Remove "for" loop for filters. Each case (or and and) would have their own "for" loop. In case of "and", you want to check all. In case of "or", you only need one match.
- This is horizontal depth, you want to have vertical depth. Depth limit should be enforced in validate criteria. We can use depth of 3 for now and revisit in separate issue later.

TakeFromBasket:
- Integrate ORM

robert-zaremba:
ON HOLD : blocked by orm code gen work in SDK.

technicallyty:
closing as we can just port #732

PR #700, opened 2022-01-24
feat: ecocredit ORM proto definitions 

technicallyty:
## Description

adds ORM proto definitions for x/ecocredit

Closes: n/a

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,213 @@
...
+    // metadata is any arbitrary metadata attached to the project.
+    bytes metadata = 5;
+}
+
+// BatchInfo represents the high-level on-chain information for a credit batch.
+message BatchInfo {
+
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 3,
+        primary_key: {fields: "batch_denom"}
```
aaronc, line 67: Let's actually add an uint64 auto increment primary key and make this an unique index. I think we'll want a short uint64 to reference batches in the order book and denoms are getting kind of long

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,213 @@
...
+import "cosmos/base/v1beta1/coin.proto";
+import "gogoproto/gogo.proto";
+import "google/protobuf/timestamp.proto";
+
+// ClassInfo represents the high-level on-chain information for a credit class.
+message ClassInfo {
+
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 1,
+        primary_key: {fields: "class_id"}
```
aaronc, line 15: Maybe let's add admin and credit type indexes?

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,213 @@
...
+    // The number of batches issued in this credit class.
+    uint64 num_batches = 6;
+}
+
+// ProjectInfo represents the high-level on-chain information for a project.
+message ProjectInfo {
+
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 2,
+        primary_key: {fields: "project_id"}
```
aaronc, line 42: How about issuer and class indexes?

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,213 @@
...
+
+    // the measurement unit (e.g. kg, ton, etc)
+    string unit = 3;
+
+    // the decimal precision
+    uint32 precision = 4;
+}
+
+// SellOrder represents the information for a sell order.
+message SellOrder {
```
aaronc, line 126: Let's move orders into a separate proto package

aaronc (commented):
Great start @technicallyty ! Left a few suggestions

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,213 @@
...
+    };
+
+    // class_id is the unique ID of credit class.
+    string class_id = 1;
+
+    // admin is the admin of the credit class.
+    string admin = 2;
+
+    // issuers are the approved issuers of the credit class.
+    repeated string issuers = 3;
```
aaronc, line 25: In keeping with 1NF (which I think we should), let's move this to its own join table

proto/regen/ecocredit/v1beta1/state.proto
technicallyty: are there docs anywhere for the syntax on this?

proto/regen/ecocredit/v1beta1/state.proto
technicallyty: i separated it like:

`ecocredit/v1beta1/marketplace/state.proto`


is that how we want it?

proto/regen/ecocredit/v1beta1/state.proto
technicallyty: https://github.com/regen-network/regen-ledger/pull/700/files#diff-302ae7137e2a8ea4a04301cf72f3df8311c99ea4716efaf59ae04291e030fb63R16-R17

proto/regen/ecocredit/v1beta1/state.proto
technicallyty: added: https://github.com/regen-network/regen-ledger/pull/700/files#diff-302ae7137e2a8ea4a04301cf72f3df8311c99ea4716efaf59ae04291e030fb63R56-R57

proto/regen/ecocredit/v1beta1/marketplace/state.proto
```diff
@@ -0,0 +1,96 @@
...
+import "cosmos/orm/v1alpha1/orm.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+// SellOrder represents the information for a sell order.
+message SellOrder {
+
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 1,
+        primary_key: {fields: "order_id"}
+        index: {id: 1 fields: "batch_denom"}
```
aaronc, line 14: how about owner index?

proto/regen/ecocredit/v1beta1/marketplace/state.proto
```diff
@@ -0,0 +1,96 @@
...
+    // resell the credits to another buyer.
+    bool disable_auto_retire = 6;
+}
+
+// BuyOrder represents the information for a buy order.
+message BuyOrder {
+
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 2,
+        primary_key: {fields: "buy_order_id"}
```
aaronc, line 45: buyer index too?

proto/regen/ecocredit/v1beta1/state.proto
aaronc: in orm.proto
```suggestion
        primary_key: {fields: "id", auto_increment: true}
        index: {fields: "batch_denom", unique: true}
```

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,139 @@
...
+
+    // The number of batches issued in this credit class.
+    uint64 num_batches = 5;
+}
+
+// ClassIssuers is a JOIN table for Class Info that stores the credit class issuers
+message ClassIssuers {
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 2,
+        primary_key: {fields: "class_id"}
```
aaronc, line 40:
```suggestion
        primary_key: {fields: "class_id,issuer"}
```

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,139 @@
...
+    };
+
+    // class_id is the unique ID of credit class.
+    string class_id = 1;
+
+    // issuers are the approved issuers of the credit class.
+    repeated string issuers = 2;
```
aaronc, line 47:
```suggestion
   string issuer = 2;
```

in 1NF we can't have any `repeated` fields

aaronc (commented):
Looking good. Don't we have sequences to track for generating class, project and batch IDs?

technicallyty:
> Looking good. Don't we have sequences to track for generating class, project and batch IDs?

right, so these would be separate messages/tables right? i think batches are covered now with the addition of the auto_increment id

aaronc:
> 

Yes separate tables, but I think not quite for batch ID because these batch IDs are unique per project and project IDs unique per class right?

proto/regen/ecocredit/v1beta1/state.proto
aaronc: Yep 👍

aaronc:
We also need balances and supply

technicallyty:
> Yes separate tables, but I think not quite for batch ID because these batch IDs are unique per project and project IDs unique per class right?

actually looks like we were tracking batch sequence in the class via the field `num_batches`. should we continue with this method?

i've added a project seq table, and credit/basket balance. looks like the supply is already tracked in `BatchInfo` via `total_amount`.

aaronc:
> > Yes separate tables, but I think not quite for batch ID because these batch IDs are unique per project and project IDs unique per class right?
> 
> actually looks like we were tracking batch sequence in the class via the field `num_batches`. should we continue with this method?
> 
> i've added a project seq table, and credit/basket balance. looks like the supply was already tracked in `BatchInfo` via `total_amount`.

I sort of have the preference that things that get updated somewhat frequently - like sequences and supply - should live on separate tables from things which rarely or never change. It's not a huge deal, but I want to avoid the pattern that is done some places in the SDK like x/auth where every acct seq update rewrites the whole account with vesting, pub key and all

technicallyty:
> > > Yes separate tables, but I think not quite for batch ID because these batch IDs are unique per project and project IDs unique per class right?
> > 
> > 
> > actually looks like we were tracking batch sequence in the class via the field `num_batches`. should we continue with this method?
> > i've added a project seq table, and credit/basket balance. looks like the supply was already tracked in `BatchInfo` via `total_amount`.
> 
> I sort of have the preference that things that get updated somewhat frequently - like sequences and supply - should live on separate tables from things which rarely or never change. It's not a huge deal, but I want to avoid the pattern that is done some places in the SDK like x/auth where every acct seq update rewrites the whole account with vesting, pub key and all

ok, added batch sequence and separate batch supply table

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,225 @@
...
+
+    // the measurement unit (e.g. kg, ton, etc)
+    string unit = 3;
+
+    // the decimal precision
+    uint32 precision = 4;
+}
+
+// CreditTypeSequence stores and increments the sequence number for credit types
+message CreditTypeSequence {
```
aaronc, line 129: What are these sequences being used for?

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,225 @@
...
+// ProjectSequence stores and increments the sequence number for projects
+message ProjectSequence {
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 7,
+        primary_key: {fields: "project_id"}
+    };
+
+    string project_id = 1;
+
+    uint64 id = 2;
```
aaronc, line 149:
```suggestion
    uint64 next_id = 2;
```

proto/regen/ecocredit/v1beta1/state.proto
technicallyty: i believe it was added in the case we want multiple types of "carbon" credits with different parameters, i.e carbon but measured in a different unit. in which case the abbreviation would still contain "C" and then a sequence for each additional type C0, C01, C02 etc

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,225 @@
...
+    // batch_denom is the denom of the credit batch
+    string batch_denom = 2;
+
+    message CreditBalance {
+        string tradable = 1;
+        string retired = 2;
+    }
+
+    // balance is the balance of both tradable and retired credits
+    CreditBalance balance = 3;
```
aaronc, line 171:
```suggestion
        string tradable = 1;
        string retired = 2;
```

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,225 @@
...
+
+    // basket_denom is the denom of the basket
+    string basket_denom = 1;
+
+    // batch_denom is the denom of the batch
+    string batch_denom = 2;
+
+    // balance is the amount of ecocredits held in the basket
+    string balance = 3;
+}
```
aaronc, line 189: Let's put this in a basket sub-package

proto/regen/ecocredit/v1beta1/state.proto
aaronc: that doesn't make sense 🤔 . I think we should nix it. the abbreviation is the primary key and this is controlled by governance

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,220 @@
...
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 9,
+        primary_key: {fields: "basket_denom,batch_denom"}
+    };
+
+    // basket_denom is the denom of the basket
+    string basket_denom = 1;
+
+    // batch_denom is the denom of the batch
+    string batch_denom = 2;
```
aaronc, line 180: can we reduce storage space using uint64 id's?

proto/regen/ecocredit/v1beta1/state.proto
technicallyty: would it still be auto-generated then?

aaronc: credit type was never meant to be auto-generated. it is chosen by governance. i'm confused by where this seq came from tbh

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,208 @@
...
+    };
+
+    // id is the id of the batch
+    uint64 id = 1;
+
+    // total_amount is the total number of active credits in the credit batch.
+    // Some of the issued credits may be cancelled and will be removed from
+    // total_amount and tracked in amount_cancelled. total_amount and
+    // amount_cancelled will always sum to the original credit issuance amount.
+    string total_amount = 2;
```
aaronc, line 202: Where are we tracking retired amount? That's different from cancelled

proto/regen/ecocredit/v1beta1/state.proto
aaronc: no, the abbreviation is chosen by governance. just `C` for instance. let's fix it

proto/regen/ecocredit/v1beta1/baskets/state.proto
```diff
@@ -0,0 +1,22 @@
...
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 9,
+        primary_key: {fields: "basket_denom,batch_denom"}
+    };
+
+    // basket_denom is the denom of the basket
+    string basket_denom = 1;
+
+    // batch_denom is the denom of the batch
+    string batch_denom = 2;
```
aaronc, line 18:
```suggestion
    uint64 batch_id = 2;
```

proto/regen/ecocredit/v1beta1/state.proto
technicallyty: it wasn't tracked in the original design. i can split the supply into tradable and retired

technicallyty: https://github.com/regen-network/regen-ledger/pull/700/commits/40bbe1ed417b578073d2f232230562cd54a61cf5

proto/regen/ecocredit/v1beta1/state.proto
technicallyty: https://github.com/regen-network/regen-ledger/pull/700/commits/a40947c8ec4f81850b3491bb44aab356be08f474

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,199 @@
...
+
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 4,
+        primary_key: {fields: "id", auto_increment: true}
+        index: {fields: "batch_denom", unique: true}
+        index: {id: 1 fields: "project_id"}
+        index: {id: 2 fields: "start_date"}
+    };
+
+    // id is a an incrementing integer to identify the batch
```
aaronc, line 85:
```suggestion
    // id is an auto-incrementing integer to succintly identify the batch
```

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,199 @@
...
+
+// EcocreditBalance stores each users credit balance
+message EcocreditBalance {
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 8,
+        primary_key: {fields: "address,batch_id"}
+    };
+
+    // address is the address of the credit holder
+    string address = 1;
```
aaronc, line 150: let's use `bytes`

proto/regen/ecocredit/v1beta1/state.proto
technicallyty: added an issue to remove the sequencing logic from the ecocredit module https://github.com/regen-network/regen-ledger/issues/707

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,199 @@
...
+    uint64 batch_id = 2;
+
+    // tradable is the tradable amount of credits
+    string tradable = 3;
+
+    // retired is the retired amount of credits
+    string retired = 4;
+}
+
+// BatchSequence tracks the sequence number for batches within a class
```
aaronc, line 162:
```suggestion
// BatchSequence tracks the sequence number for batches within a project
```

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,199 @@
...
+message BatchSequence {
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 10,
+        primary_key: {fields: "class_id"}
+    };
+
+    // class_id is the id of the class for a batch
+    uint64 class_id = 1;
```
aaronc, line 170:
```suggestion
    string project_id = 1;
```

aaronc (commented):
getting close! thanks for your patience!

proto/regen/ecocredit/v1beta1/state.proto
technicallyty: should we be using cosmos scalars? or just change to bytes for now

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -0,0 +1,199 @@
...
+
+    // next_id is the sequence number for the project
+    uint64 next_id = 2;
+}
+
+// EcocreditBalance stores each users credit balance
+message EcocreditBalance {
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 8,
+        primary_key: {fields: "address,batch_id"}
```
aaronc, line 146:
```suggestion
        primary_key: {fields: "address,batch_id"}
        index: {id: 1, fields: "batch_id,address"}
```

aaronc (approved):
LGTM. Thanks @technicallyty 🎉 !

aaronc:
Just need to fix conflicts

proto/regen/ecocredit/v1beta1/marketplace/state.proto
```diff
@@ -0,0 +1,98 @@
...
+
+// SellOrder represents the information for a sell order.
+message SellOrder {
+
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 1,
+        primary_key: {fields: "order_id"}
+        index: {id: 1 fields: "batch_denom"}
+        index: {id: 2 fields: "owner"}
+    };
```
ryanchristo, line 16: Following #671, we need to include the `expiration` message field and add an expiration index.

proto/regen/ecocredit/v1beta1/marketplace/state.proto
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// BuyOrder represents the information for a buy order.
+message BuyOrder {
+
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 2,
+        primary_key: {fields: "buy_order_id"}
+        index: {id: 1 fields: "buyer"}
+    };
```
ryanchristo, line 48: Following #671, we need to include the expiration message field and add an expiration index.

proto/regen/ecocredit/v1beta1/marketplace/state.proto
```diff
@@ -0,0 +1,98 @@
...
+    // the credits will not auto-retire and can be resold assuming that the
+    // corresponding sell order has auto-retirement disabled. If the sell order
+    // hasn't disabled auto-retirement and the buy order tries to disable it,
+    // that buy order will fail.
+    bool disable_auto_retire = 6;
+
+    // disable_partial_fill disables the default behavior of partially filling
+    // buy orders if the requested quantity is not available.
+    bool disable_partial_fill = 7;
+}
```
ryanchristo, line 98: We need to add `AllowAskDenom` here as well.

proto/regen/ecocredit/v1beta1/baskets/state.proto
```diff
@@ -0,0 +1,22 @@
...
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 9,
+        primary_key: {fields: "basket_denom,batch_id"}
+    };
+
+    // basket_denom is the denom of the basket
+    string basket_denom = 1;
+
+    // batch_id is the denom of the batch
+    uint64 batch_id = 2;
```
ryanchristo, line 18:
```suggestion
    // batch_id is the id of the credit batch
    uint64 batch_id = 2;
```

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -3,25 +3,198 @@ syntax = "proto3";
...
+    string tradable = 3;
+
+    // retired is the retired amount of credits
+    string retired = 4;
+}
+
+// BatchSequence tracks the sequence number for batches within a project
+message BatchSequence {
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 10,
```
clevinson, line 166:
```suggestion
        id: 9,
```

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -3,25 +3,198 @@ syntax = "proto3";
...
+    string project_id = 1;
+
+    // num_batches is a sequence number incrementing on each issued batch
+    uint64 num_batches = 2;
+}
+
+// BatchSupply tracks the supply of a credit batch
+message BatchSupply {
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 11,
```
clevinson, line 180:
```suggestion
        id: 10,
```

proto/regen/ecocredit/v1beta1/baskets/state.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+
+package regen.ecocredit.v1beta1.baskets;
+
+import "cosmos/orm/v1alpha1/orm.proto";
+
+// BasketBalance stores the amount of credits from a batch in a basket
+message BasketBalance {
+    option (cosmos.orm.v1alpha1.table) = {
+        id: 9,
```
clevinson, line 10:
```suggestion
        id: 1,
```

clevinson: Should this be `1`? As its the only table in this file

proto/regen/ecocredit/v1beta1/marketplace/state.proto
technicallyty: https://github.com/regen-network/regen-ledger/pull/700/commits/22491e9d5d575fadaa875d201750bf95933ea5c9

proto/regen/ecocredit/v1beta1/marketplace/state.proto
technicallyty: https://github.com/regen-network/regen-ledger/pull/700/commits/22491e9d5d575fadaa875d201750bf95933ea5c9

proto/regen/ecocredit/v1beta1/marketplace/state.proto
technicallyty: https://github.com/regen-network/regen-ledger/pull/700/commits/22491e9d5d575fadaa875d201750bf95933ea5c9

ryanchristo: Looks like the expiration index is still missing..

proto/regen/ecocredit/v1beta1/marketplace/state.proto
ryanchristo: Same here..

proto/regen/ecocredit/v1beta1/marketplace/state.proto
technicallyty: oh whoops, think i missed the index part and only added it as a field 😅 its in there now!

ryanchristo (approved):

proto/regen/ecocredit/v1beta1/marketplace/state.proto
```diff
@@ -0,0 +1,129 @@
...
+
+    // Selection defines a buy order selection.
+    message Selection {
+
+        // sum defines the type of selection.
+        oneof sum {
+            // sell_order_id is the sell order ID against which the buyer is trying to buy.
+            // When sell_order_id is set, this is known as a direct buy order because it
+            // is placed directly against a specific sell order.
+            uint64 sell_order_id = 1;
```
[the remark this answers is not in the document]
technicallyty: i think that would be out of the scope of this PR. this structure was copied from existing proto defs. so i believe this concern would warrant its own issue

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -3,25 +3,198 @@ syntax = "proto3";
...
+        id: 7,
+        primary_key: {fields: "project_id"}
+    };
+
+    // project_id is the id of the project
+    string project_id = 1;
+
+    // next_id is the sequence number for the project
+    uint64 next_id = 2;
+}
```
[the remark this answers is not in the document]
technicallyty: so the format of the project id is classID + sequence number. i think actually the project_id field should be class_id. but in any case, the 2 fields are needed

proto/regen/ecocredit/v1beta1/marketplace/state.proto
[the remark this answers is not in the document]
aaronc: why a list?

proto/regen/ecocredit/v1beta1/state.proto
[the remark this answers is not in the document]
aaronc: it should be class_id

proto/regen/ecocredit/v1beta1/marketplace/state.proto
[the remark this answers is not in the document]
robert-zaremba: @aaronc to select multiple orders to fill.

proto/regen/ecocredit/v1beta1/state.proto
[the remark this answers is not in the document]
robert-zaremba: ah, so `ProjectSequence` is a type for a key.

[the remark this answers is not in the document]
robert-zaremba: @technicallyty if you will be updating the `project_id` field, let's update the `message` documentation as well.

proto/regen/ecocredit/v1beta1/marketplace/state.proto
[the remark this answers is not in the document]
aaronc: this model is just based on buying from a specific sell order. it's not a batch operation. it's like a shopping cart where you buy one product from one vendor

PR #717, opened 2022-02-01
refactor(x/ecocredit): rename bank_denom to base_currency

robert-zaremba:
What do you think about using `base_currency` instead of `bank_denom`?

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


aaronc:
The idea with bank is to differentiate it more from batch denom. Base and batch doesn't seem to do that as well. How about base currency or currency denom?

blushi:
> The idea with bank is to differentiate it more from batch denom. Base and batch doesn't seem to do that as well. How about base currency or currency denom?

Agreed, coin denom could also work.

technicallyty:
> > The idea with bank is to differentiate it more from batch denom. Base and batch doesn't seem to do that as well. How about base currency or currency denom?
> 
> Agreed, coin denom could also work.

+1 for coin denom, pretty clear that way

robert-zaremba:
"coin denom" is purple prose. Also, in SDK `coin = {denom, amount}`, so I would mix it. I like currency more than denom.

My favorite is (in this order):
+ base_currency
+ base_denom

@dtkong , @glandua, @clevinson any thoughts?

dtkong:
hmm, how about the possibility of using `token` ? eg  `base_token` , `token_denom` ? 
`base_denom` isn't particularly evocative to me. 
`base_currency` is okay. currency implies there might be currency exchange though? 
i slightly dislike `coin_denom`; I don't love the idea of `coin` as it reminds me of skeuomorphic interfaces.

proto/regen/ecocredit/marketplace/v1beta1/state.proto
```diff
@@ -113,16 +113,16 @@ message BuyOrder {
...
         id: 3
-        primary_key: {fields: "bank_denom"}
+        primary_key: {fields: "base_denom"}
         index: {id: 1 fields: "display_denom", unique: true}
     };
 
-    // denom is the bank denom to allow (ex. ibc/GLKHDSG423SGS)
-    string bank_denom = 1;
+    // base is the bank denom to allow (ex. ibc/GLKHDSG423SGS)
+    string base_denom = 1;
```
robert-zaremba, line 125: could be `base_currency` or `base_token`

aaronc: base_currecny or currency_denom. But honestly I don't see this as a big deal

aaronc: base_denom is less clear to me though

ryanchristo:
It looks like we never came to an agreement here. Maybe we should stick with `bank_denom` and close this pull request?

robert-zaremba:
I think we oscillate around `base_currency`.

proto/regen/ecocredit/marketplace/v1/state.proto
```diff
@@ -125,12 +125,12 @@ message BuyOrder {
 message AllowedDenom {
   option (cosmos.orm.v1alpha1.table) = {
     id : 3
-    primary_key : {fields : "bank_denom"}
+    primary_key : {fields : "base_currency"}
     index : {id : 1 fields : "display_denom", unique : true}
   };
 
   // denom is the bank denom to allow (ex. ibc/GLKHDSG423SGS)
```
ryanchristo, line 132:
```suggestion
  // base_currency is the bank denom to allow (ex. ibc/GLKHDSG423SGS)
```

ryanchristo: The description should probably be updated as well but I'm not sure how to word it because I think "bank denom" is an accurate description.

ryanchristo (commented):
My vote is `bank_denom` because this field represents a "denom" for a token that is minted using the "bank" module and it provides a nice contrast to `batch_denom`.

If we update "denom" to "currency", would we need to reconsider `display_denom`, `allow_ask_denom`, `basket_denom`, and/or `batch_denom`? I don't think we want to change any of these fields nor should we deviate from our current naming convention.

If others think this is an improvement, I can adjust, but to me this seems like an unnecessary change and it may cause more confusion to both users and our developers.

aaronc:
> My vote is `bank_denom` because this field represents a "denom" for a token that is minted using the "bank" module and it provides a nice contrast to `batch_denom`.

I generally agree

robert-zaremba:
> My vote is bank_denom because this field represents a "denom" for a token that is minted using the "bank" module and it provides a nice contrast to batch_denom.

This is make sense form the technical point of view, and only assuming there won't be any other non-bank currency in the future (changing the logic to accept more token types is easier than changing the proto name and type).
But, IMO, base currency is way more clear to the market than bank denom.

aaronc:
> > My vote is bank_denom because this field represents a "denom" for a token that is minted using the "bank" module and it provides a nice contrast to batch_denom.
> 
> This is make sense form the technical point of view, and only assuming there won't be any other non-bank currency in the future (changing the logic to accept more token types is easier than changing the proto name and type). But, IMO, base currency is way more clear to the market than bank denom.

I can't imagine a non-bank currency because bank is the general way tokens work and we're only talking about regular cosmos tokens. Seems like @technicallyty @ryanchristo and I prefer `bank_denom`. Sorry @robert-zaremba, but let's close this

robert-zaremba:
OK :)

Issue #689, opened 2022-01-19
App wiring: add ecocredit module to module manager


ryanchristo:
## Summary

Following #671, we need to wire up `BeginBlocker` for the ecocredit module.

As far as I understand, we need to include the ecocredit module when initializing the module manager and then include the module when setting the order for begin blockers. The ecocredit module does not implement the `AppModule` interface and in attempting to do so there is a conflict with `RegisterServices` - we use a custom `Configurator` in the ecocredit module with a different signature.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

technicallyty:
we should rename the function defined in the interface here: https://github.com/regen-network/regen-ledger/blob/master/types/module/server/module.go#L11

imo doesn't make much sense to have this method named as such as it collides with the sdk's module interface as you've stated above.

aaronc:
we should just get the app wiring work done with and be done with `AppModule`

ryanchristo:
Looking into adding BeginBlocker to Configurator as possible solution for the time being.

PR #720, opened 2022-02-03
feat: add begin and end blocker support to module/server

robert-zaremba:
## Description

Closes: #689

Initially I started implementing begin and end blocker support through new `Configurator` register functions (similarly to `RegisterGenesisHandlers). However this  approach has downsides:
* it hides modules, so it's not possible to customize the execution order
* it further away from the SDK design
* would require yet another small rework of the ecocredit Module by removing the keeper and moving the whole logic to the keeper itself (the server) - which is mixing module logic with the data and execution layer.

So, instead I added Begin and End blocker functions to the Regen `module/server.Module`, which uses modules directly to execute the begin / end blockers. 
I had to update the module to be mutable - the keeper assignment in [RegisterServices](https://github.com/regen-network/regen-ledger/blob/26cb432c298245804dd56360437f7f96977cd39b/x/ecocredit/module/module.go#L65) wouldn't  work otherwise.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/module/module.go
```diff
@@ -35,23 +35,23 @@ type Module struct {
 }
 
 // NewModule returns a new Module object.
-func NewModule(paramSpace paramtypes.Subspace, accountKeeper ecocredit.AccountKeeper, bankKeeper ecocredit.BankKeeper) Module {
+func NewModule(paramSpace paramtypes.Subspace, accountKeeper ecocredit.AccountKeeper, bankKeeper ecocredit.BankKeeper) *Module {
```
robert-zaremba, line 38: Pointer receiver is required to handle keeper registration in `RegisterServices`

aaronc (commented):
It doesn't seem like this does any particular ordering, other than registration order which is deterministic. Maybe that's okay since it's a short term fix. It would be good to actually make ecocredit have a begin and end blocker even if they don't do anything yet.

robert-zaremba:
@aaronc x/ecocredit has a registered Begin Blocker (in this PR). It doesn't have end blocker. I can add an empty one.

>  Maybe that's okay since it's a short term fix. 

I don't think we need it (there is only one module, and even with cosmowasm, registration order will work good enough). Once we will Cosmos SDK 0.46 we should migrate to `AppModule`

aaronc:
the simulation issue does seem to be related to the ecocredit begin blocker

x/ecocredit/server/expiration.go
```diff
@@ -26,16 +28,19 @@ func (s serverImpl) PruneOrders(ctx sdk.Context) error {
...
 		err = s.sellOrderTable.Delete(ctx, sellOrder.OrderId)
 		if err != nil {
+			logger.Error("problem with sell order delete", "err", err)
 			return err
 		}
 	}
 
 	buyOrdersIter, err := s.buyOrderByExpirationIndex.PrefixScan(ctx, minTime, blockTime)
 	if err != nil {
+		logger.Error("problem with buy order iteratorr", "err", err)
```
technicallyty, line 43:
```suggestion
		logger.Error("problem with buy order iterator", "err", err)
```

technicallyty (commented):
could you add a test for these changes?

aleem1314:
@robert-zaremba I think simulations are failing because of duplicate prefixes for orm [tables](https://github.com/regen-network/regen-ledger/blob/master/x/ecocredit/server/server.go#L22-L32).

ryanchristo (approved):
Looks good to me. One question. Pre-approving.

x/ecocredit/server/server.go
```diff
@@ -22,23 +22,23 @@ const (
...
 	BatchInfoTablePrefix     byte = 0x6
 
-	ProjectInfoTablePrefix    byte = 0x12
-	ProjectInfoTableSeqPrefix byte = 0x13
-	ProjectsByClassIDIndex    byte = 0x14
-	BatchesByProjectIndex     byte = 0x15
+	ProjectInfoTablePrefix    byte = 0x10
+	ProjectInfoTableSeqPrefix byte = 0x11
+	ProjectsByClassIDIndex    byte = 0x12
+	BatchesByProjectIndex     byte = 0x13
```
aaronc, line 28: I don't think this file should have changed in this PR. Looks like something got messed up

aaronc (changes requested):
LGTM. But there are some unrelated changes that need to be reverted. Once that's done ready to approve

x/ecocredit/server/server.go
robert-zaremba: This prevents the sim tests to pass.

robert-zaremba: basically there is a conflict between the prefixes.

robert-zaremba: this got broken in https://github.com/regen-network/regen-ledger/pull/647

aaronc (approved):
approving as @robert-zaremba  clarified that the changes to store key prefixes are to fix conflicts introduced in #647

Issue #707, opened 2022-01-25
ecocredit: remove credit type sequences


technicallyty:
## Summary

in #700 it was found that credit type sequences didn't make sense and should be removed. Abbreviations should be controlled entirely by governance. 

This can probably be done once ORM work is introduced

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #696, opened 2022-01-22
x/ecocredit: migrate ecocredit storage model to ORM


technicallyty:
## Summary

We want to migrate the ecocredit storage model to use the ORM. This allows us to define indexes for easier and more efficient querying such as querying batches by start date, as needed for `TakeFromBasket`. 

This is similar to #684 but just changes the backend. we could still create some streamlined functions like bank has.

when this is finished, we should uncomment the interface guard to ensure we impl the MsgServer interface 100%.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #723, opened 2022-02-03
refactor(ecocredit)!: migrate core ecocredit msg_server to use ORM

technicallyty:
## Description

beginning of ecocredit ORM refactor

Closes: #696 

Closes: #707 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/server/testsuite/suite.go
```diff
@@ -50,14 +54,13 @@ func NewIntegrationTestSuite(fixtureFactory testutil.FixtureFactory, paramSpace
 
 func (s *IntegrationTestSuite) SetupSuite() {
```
technicallyty, line 55: I think we should consider refactoring this massive integration test into smaller, bite sized unit tests, and then just have a big integration scenario test. 

it's a bit confusing to figure my way around this code. the batch info test is broken and the one thats failing has a hard coded batch denom to search for. i'm not really sure what the best approach to fixing that test would be other than obliterating the existing test structure to fit my needs..

should we create an issue to move these smaller tests into unit tests? @aaronc @ryanchristo @robert-zaremba thoughts?

ryanchristo: I agree. A lot of these tests are written like unit tests already. I think it makes sense to separate them out and to simply the integration tests in a separate issue. You should be able to trace the batch denom issue without obliterating the existing structure. Would be happy to take a look if you get stuck.

proto/regen/ecocredit/v1beta1/state.proto
```diff
@@ -93,6 +94,9 @@ message ProjectInfo {
 
     // metadata is any arbitrary metadata attached to the project.
     bytes metadata = 5;
+
+    // issuer is the address that is allowed to issue credits from this project
+    string issuer = 6;
```
aaronc, line 99: We actually don't want to do this restriction here and instead want any credit class issuer to be able to issue
```suggestion
```

proto/regen/ecocredit/v1alpha2/query.proto
```diff
@@ -23,6 +23,11 @@ service Query {
         "/regen/ecocredit/v1alpha2/classes/{class_id}";
   }
 
+  rpc ClassIssuers(QueryClassIssuersRequest) returns (QueryClassIssuersResponse) {
```
aaronc, line 26: Can we move all these changes to v1beta1 and leave v1alpha1 as is for now?

proto/regen/ecocredit/basket/v1beta1/tx.proto
```diff
@@ -4,7 +4,7 @@ package regen.ecocredit.basket.v1beta1;
 
 import "regen/ecocredit/basket/v1beta1/types.proto";
 
-option go_package = "github.com/regen-network/regen-ledger/x/ecocredit/v1beta1";
+option go_package = "github.com/regen-network/regen-ledger/x/ecocredit/basket/v1beta1";
```
technicallyty, line 7: code wasn't being generated for these packages, i assume they were being overwritten by core which gets written last? all code from these sub packages get generated now

x/ecocredit/server/core/query_server.go
```diff
@@ -0,0 +1,619 @@
...
+	"google.golang.org/grpc/status"
+)
+
+// Classes queries for all credit classes with pagination.
+func (s serverImpl) Classes(ctx context.Context, request *v1beta1.QueryClassesRequest) (*v1beta1.QueryClassesResponse, error) {
+	if request == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "empty request")
+	}
+	if request.Pagination == nil {
+		request.Pagination = &query.PageRequest{}
```
technicallyty, line 24: if pagination is nil, should i just be passing nil to the `List` call? or is the empty PageRequest ok?

ryanchristo (commented):
Looking great. Nice work! Still working through the review. A few comments to start.

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1193 @@
...
+		return nil, err
+	}
+	if class.Admin != req.Admin {
+		return nil, sdkerrors.ErrUnauthorized.Wrapf("expected admin %s, got %s", class.Admin, req.Admin)
+	}
+
+	// delete the old issuers
+	if err = s.stateStore.ClassIssuerStore().DeleteBy(ctx, ecocreditv1beta1.ClassIssuerClassIdIssuerIndexKey{}.WithClassId(class.Name)); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
technicallyty: the primary key is made up of the classID+issuer_address, so it wouldn't work in this case. if it had a simple row ID  primary key with class+issuer indexes, i could do that - i'm not sure on the benchmarks of either case though, could look into some metrics for ORM ops if we value it

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1193 @@
...
+const gasCostPerIteration = uint64(10)
+
+// CreateClass creates a new class of ecocredit
+//
+// The admin is charged a fee for creating the class. This is controlled by
+// the global parameter CreditClassFee, which can be updated through the
+// governance process.
+func (s serverImpl) CreateClass(ctx context.Context, req *v1beta1.MsgCreateClass) (*v1beta1.MsgCreateClassResponse, error) {
+	regenCtx := types.UnwrapSDKContext(ctx)
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
```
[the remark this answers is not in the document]
technicallyty: fixed in https://github.com/regen-network/regen-ledger/pull/723/commits/20695f8b260b3b738107f25c93d8f82a5d6a3602#diff-170c3cd03161feb06e17c134c27d9cf4bb4af8e07f428df2606bde3482eec5b2L29

proto/regen/ecocredit/v1beta1/events.proto
```diff
@@ -6,60 +6,66 @@ option go_package = "github.com/regen-network/regen-ledger/x/ecocredit/v1beta1";
 
 // EventCreateClass is an event emitted when a credit class is created.
 message EventCreateClass {
+  // row_id is the id of the database row entry
+  uint64  row_id = 1;
```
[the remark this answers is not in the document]
technicallyty: hmm i think i added it in the assumption we might add `QueryFooByRowID` later down the line.. not sure though, @aaronc is this something we expect to do?

types/math/dec.go
```diff
@@ -33,6 +32,9 @@ var dec128Context = apd.Context{
 }
 
 func NewDecFromString(s string) (Dec, error) {
+	if s == "" {
+		s = "0"
+	}
```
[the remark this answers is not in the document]
technicallyty: i found a few places where were we would explicitly check for the empty string case and then set it to "0" anyway, so it felt like a good way to reduce that.

are there any specific cases where we'd prefer to error here instead of default to 0?

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1193 @@
...
+		ProjectId:  projectInfo.Id,
+		BatchDenom: batchDenom,
+		Metadata:   req.Metadata,
+		StartDate:  timestamppb.New(req.StartDate.UTC()),
+		EndDate:    timestamppb.New(req.EndDate.UTC()),
+	})
+	if err != nil {
+		return nil, err
+	}
+	newBatchID := rowID
```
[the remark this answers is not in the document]
technicallyty: fixed

x/ecocredit/server/core/operations.go
```diff
@@ -0,0 +1,20 @@
+package core
+
+import (
+	"github.com/cosmos/cosmos-sdk/types/module"
+	simtypes "github.com/cosmos/cosmos-sdk/types/simulation"
+)
+
+// WeightedOperations returns all the ecocredit module operations with their respective weights.
+// TODO: sim refactor PR
+func (s serverImpl) WeightedOperations(simState module.SimulationState) []simtypes.WeightedOperation {
```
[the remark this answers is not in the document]
technicallyty: created an issue and linked in code https://github.com/regen-network/regen-ledger/issues/728

proto/regen/ecocredit/v1beta1/events.proto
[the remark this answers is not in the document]
aaronc: Let's omit it for now as it's more for internal usage

proto/regen/ecocredit/v1beta1/events.proto
```diff
@@ -6,60 +6,66 @@ option go_package = "github.com/regen-network/regen-ledger/x/ecocredit/v1beta1";
...
 
   // admin is the admin of the credit class.
-  string admin = 2;
+  string admin = 3;
 }
 
 // EventCreateProject is an event emitted when a project is created.
 message EventCreateProject {
+  // row_id is the id of the database row entry
+  uint64  row_id = 1;
```
aaronc, line 22: Ditto

proto/regen/ecocredit/v1beta1/events.proto
```diff
@@ -6,60 +6,66 @@ option go_package = "github.com/regen-network/regen-ledger/x/ecocredit/v1beta1";
...
   // project_location is the location of the project.
   // Full documentation can be found in MsgCreateProject.project_location.
-  string project_location = 4;
+  string project_location = 5;
 }
 
 // EventCreateBatch is an event emitted when a credit batch is created.
 message EventCreateBatch {
+  // row_id is the id of the database row entry
+  uint64  row_id = 1;
```
aaronc, line 41: Ditto

types/math/dec.go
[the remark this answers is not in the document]
aaronc: If we want to handle "", no need to parse anything just return a zero value. 

Seems like we'd check for "" elsewhere if we actually want a billable case so I think this default is okay

x/ecocredit/denom.go
```diff
@@ -76,3 +77,16 @@ func ValidateDenom(denom string) error {
 	}
 	return nil
 }
+
+// GetClassIdFromBatchDenom returns the classID in a batch denom
+func GetClassIdFromBatchDenom(denom string) string {
```
aaronc, line 82: Test?

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1200 @@
...
+		return nil, err
+	}
+
+	var params ecocredit.Params
+	s.paramSpace.GetParamSet(sdkCtx, &params)
+	if params.AllowlistEnabled && !s.isCreatorAllowListed(sdkCtx, params.AllowedClassCreators, adminAddress) {
+		return nil, sdkerrors.ErrUnauthorized.Wrapf("%s is not allowed to create credit classes", adminAddress.String())
+	}
+
+	err = s.chargeCreditClassFee(sdkCtx, adminAddress)
```
aaronc, line 42: we should add a `fee` field to `MsgCreateClass`. Fees should never just happen. let's add a follow-up issue

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1200 @@
...
+		if creditType.Name == creditTypeName {
+			return *creditType, nil
+		}
+	}
+	return v1beta1.CreditType{}, sdkerrors.ErrInvalidType.Wrapf("%s is not a valid credit type", creditTypeName)
+}
+
+func (s serverImpl) getAllCreditTypes(ctx sdk.Context) []*v1beta1.CreditType {
+	var params ecocredit.Params
+	s.paramSpace.GetParamSet(ctx, &params)
```
aaronc, line 1162: credit types we definitely don't want in params. it should just be in the `CreditType` table. i would say let's drop all param stuff now and add gov support later

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1200 @@
...
+		CreditType: req.CreditTypeName,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, issuer := range req.Issuers {
+		issuer, _ := sdk.AccAddressFromBech32(issuer)
+		if err = s.stateStore.ClassIssuerStore().Insert(ctx, &ecocreditv1beta1.ClassIssuer{
+			ClassId: classID,
```
aaronc, line 71: should we be using row id here?

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1200 @@
...
+
+	err = s.assertClassIssuer(ctx, classID, req.Issuer)
+	if err != nil {
+		return nil, err
+	}
+
+	projectID := req.ProjectId
+	if projectID == "" {
+		found := false
+		for !found {
```
aaronc, line 107: why do we need a loop here? something seems not quite right here

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1200 @@
...
+		return nil, sdkerrors.ErrUnauthorized.Wrapf("expected admin %s, got %s", class.Admin, req.Admin)
+	}
+
+	// delete the old issuers
+	if err = s.stateStore.ClassIssuerStore().DeleteBy(ctx, ecocreditv1beta1.ClassIssuerClassIdIssuerIndexKey{}.WithClassId(class.Name)); err != nil {
+		return nil, err
+	}
+
+	// add the new issuers
+	for _, issuer := range req.Issuers {
```
aaronc, line 474: it seems like `MsgUpdateClassIssuers` should have `add` and `remove` fields rather than just replacing all the old issuers

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1200 @@
...
+	classInfo.Metadata = req.Metadata
+	if err = s.stateStore.ClassInfoStore().Update(ctx, classInfo); err != nil {
+		return nil, err
+	}
+	return &v1beta1.MsgUpdateClassMetadataResponse{}, err
+}
+
+// Sell creates new sell orders for credits
+// TODO: impl with ORM https://github.com/regen-network/regen-ledger/issues/728
+func (s serverImpl) Sell(ctx context.Context, sell *marketplacev1beta1.MsgSell) (*marketplacev1beta1.MsgSellResponse, error) {
```
aaronc, line 505: i think we can just delete all this marketplace stuff and not have this implement marketplacev1beta1.MsgServer at all

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1200 @@
...
+	//	Exponent:     req.Exponent,
+	//})
+	//if err != nil {
+	//	return nil, err
+	//}
+	//
+	//return &ecocredit.MsgAllowAskDenomResponse{}, nil
+}
+
+func (s serverImpl) CreateBasket(ctx context.Context, basket *basketv1beta1.MsgCreateBasket) (*basketv1beta1.MsgCreateBasketResponse, error) {
```
aaronc, line 894: ditto just delete

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1200 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// assertClassIssuer makes sure that the issuer is part of issuers of given classID.
+// Returns ErrUnauthorized otherwise.
+func (s serverImpl) assertClassIssuer(goCtx context.Context, classID, issuer string) error {
+	it, err := s.stateStore.ClassIssuerStore().List(goCtx, ecocreditv1beta1.ClassIssuerClassIdIssuerIndexKey{}.WithClassId(classID))
```
aaronc, line 927: we don't need to iterate, we can just do a `Has(classId, issuer)` I'm pretty sure

aaronc (commented):
Implementation generally looks good. A few things to clean-up. Some of them can be in follow-ups as they're related to existing design choices

x/ecocredit/server/core/msg_server.go
technicallyty: ah right, fixed it.

x/ecocredit/server/core/msg_server.go
technicallyty: deleted

x/ecocredit/server/core/msg_server.go
technicallyty: added

x/ecocredit/server/core/msg_server.go
technicallyty: it was actually your suggestion in the original projects PR 😄 - https://github.com/regen-network/regen-ledger/pull/647#discussion_r766151278

x/ecocredit/server/core/msg_server.go
technicallyty: added to follow-ups issue https://github.com/regen-network/regen-ledger/issues/728

x/ecocredit/server/core/msg_server.go
technicallyty: switched to use table

x/ecocredit/server/core/msg_server.go
```diff
@@ -0,0 +1,1200 @@
...
+func (s serverImpl) CreateClass(ctx context.Context, req *v1beta1.MsgCreateClass) (*v1beta1.MsgCreateClassResponse, error) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	// Charge the admin a fee to create the credit class
+	adminAddress, err := sdk.AccAddressFromBech32(req.Admin)
+	if err != nil {
+		return nil, err
+	}
+
+	var params ecocredit.Params
+	s.paramSpace.GetParamSet(sdkCtx, &params)
```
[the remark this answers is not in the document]
technicallyty: https://github.com/regen-network/regen-ledger/issues/729

x/ecocredit/server/core/query_server.go
aaronc: empty page request is okay

x/ecocredit/server/core/msg_server.go
aaronc: we need to loop to avoid a clash but the way this loop is structured is just confusing

technicallyty: do you have any suggestions to make it clearer? i had thought it was pretty straightforward but open to suggestions

technicallyty:
this PR is now more in line with the architecture used for #732. There's quite a lot of code in here, so I can opt to split this into chunks and submit multiple, smaller PR's. 

I've implemented the msg server and query servers for v1beta1 proto files with unit tests. It is not wired up to the existing msg_server yet, as that require moving around more things and making an already large diff even larger. 

Would appreciate any feedback/ideas on how to proceed from here! @aaronc @ryanchristo @clevinson :-)

ryanchristo:
This pull request is being broken up into smaller pull requests.

blushi:
> This pull request is being broken up into smaller pull requests.

should we just close this then?

PR #762, opened 2022-02-11
feat(x/ecocredit): Msg Create implementation

robert-zaremba:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


aaronc:
Remember to target `release/v2.2.x`

x/ecocredit/server/basket/create.go
```diff
@@ -3,10 +3,41 @@ package basket
...
+	denom := msg.Name
+	sender, err := sdk.AccAddressFromBech32(msg.Curator)
+	if err != nil {
+		return nil, err
+	}
+
+	err = k.stateStore.BasketStore().Insert(ctx, &basketv1.Basket{
+		BasketDenom:       denom,
+		DisableAutoRetire: msg.DisableAutoRetire,
+		CreditTypeName:    msg.CreditTypeName,
```
technicallyty, line 28: should check if this credit type exists or not, as well as if the exponent in the request is >= the credit type precision

x/ecocredit/server/basket/create.go
```diff
@@ -3,10 +3,41 @@ package basket
...
+	}
+	// TODO: need to decide about the denom creation
+	denom := msg.Name
+	sender, err := sdk.AccAddressFromBech32(msg.Curator)
+	if err != nil {
+		return nil, err
+	}
+
+	err = k.stateStore.BasketStore().Insert(ctx, &basketv1.Basket{
+		BasketDenom:       denom,
```
technicallyty, line 26: will need to check if this basket denom already exists

x/ecocredit/server/basket/create.go
```diff
@@ -3,10 +3,41 @@ package basket
...
-	//TODO implement me
-	panic("implement me")
+func (k Keeper) Create(ctx context.Context, msg *basket.MsgCreate) (*basket.MsgCreateResponse, error) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	fee := k.ecocreditKeeper.GetCreateBasketFee(ctx)
+	if err := msg.Validate(fee); err != nil {
+		return nil, err
+	}
+	// TODO: need to decide about the denom creation
+	denom := msg.Name
```
aaronc, line 19: let's just do this for now maybe:
```suggestion
	denom := fmt.Sprintf("eco/%s", msg.Name)
```

then on Monday let's sync with @clevinson and @glandua

x/ecocredit/server/basket/create.go
```diff
@@ -3,10 +3,41 @@ package basket
...
+		DisableAutoRetire: msg.DisableAutoRetire,
+		CreditTypeName:    msg.CreditTypeName,
+		DateCriteria:      msg.DateCriteria.ToApi(),
+		Exponent:          msg.Exponent,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	k.bankKeeper.SendCoinsFromAccountToModule(sdkCtx, sender, feeModuleAccName, fee)
```
aaronc, line 36: is this better than burning?

x/ecocredit/server/basket/create.go
```diff
@@ -3,10 +3,41 @@ package basket
...
+		return nil, err
+	}
+	// TODO: need to decide about the denom creation
+	denom := msg.Name
+	sender, err := sdk.AccAddressFromBech32(msg.Curator)
+	if err != nil {
+		return nil, err
+	}
+
+	err = k.stateStore.BasketStore().Insert(ctx, &basketv1.Basket{
```
aaronc, line 25: so you'll also need to insert an entry into `BasketClassStore` for each class

aaronc (commented):
Good work @robert-zaremba 👍 . This looks pretty close

x/ecocredit/server/basket/create.go
aaronc: okay, let's keep as is

clevinson: I don't think we can do this this way as it will break invariants.

Funding the community account should (I'm pretty sure) go through `distributionKeeper.FundCommunityPool()`, as the distribution keeper also tracks internally the balance of the community pool independently from the actual balance in the module account.

x/ecocredit/server/basket/create.go
aaronc: the unique constraint on `name` should handle this right?

x/ecocredit/basket/msg_create.go
```diff
@@ -57,8 +57,17 @@ func (m MsgCreate) ValidateBasic() error {
 }
 
 // Validate additional validation with access to the state data.
-func (m MsgCreate) Validate(minFee sdk.Coin) error {
-	// TODO: update function to do stateful validation
+// minFee must be sorted.
+func (m MsgCreate) Validate(minFee sdk.Coins) error {
+	// Note if user is adding too much denoms then we fails as well.
+	if len(minFee) != len(m.Fee) {
+		return errBadReq.Wrapf("Wrong denom set, expected %v", m.Fee)
```
aaronc, line 64: aliasing `sdkerrors.ErrInvalidateRequest` with `errBadReq` does not enhance readability IMHO

aaronc:
I finished this up so we can hit our timeline

x/ecocredit/server/basket/create.go
```diff
@@ -3,10 +3,122 @@ package basket
...
+	}
+	if !found {
+		return sdkerrors.ErrInvalidRequest.Wrapf("can't find credit type %s", creditType)
+	}
+	return nil
+}
+
+func (k Keeper) indexAllowedClasses(ctx context.Context, basketID uint64, allowedClasses []string) error {
+	for _, class := range allowedClasses {
+		// TODO: we should be able to check if class exists without deserializing it.
```
aaronc, line 107: Yeah we can directly check the kv store. But I would consider any optimization of Create premature since I would be surprised if it gets called even 5 times in production

aaronc (approved):

x/ecocredit/basket/msg_create.go
```diff
@@ -57,8 +57,17 @@ func (m MsgCreate) ValidateBasic() error {
 }
 
 // Validate additional validation with access to the state data.
-func (m MsgCreate) Validate(minFee sdk.Coin) error {
-	// TODO: update function to do stateful validation
+// minFee must be sorted.
+func (m MsgCreate) Validate(minFee sdk.Coins) error {
```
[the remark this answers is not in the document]
robert-zaremba: I know, the interface methods will be removed. But this is what we have now.
The idea of `Validate` is to do all stateful validation, that's why I didn't call it `ValidateFee`.

x/ecocredit/server/basket/create.go
robert-zaremba: It just smells bad if we can't use Has which is available in ORM.
I've added it

x/ecocredit/server/basket/create.go
```diff
@@ -3,10 +3,120 @@ package basket
...
+		BasketDenom: denom,
+		Curator:     msg.Curator,
+	})
+
+	return &basket.MsgCreateResponse{BasketDenom: denom}, err
+}
+
+// assertCreditTypeExists returns error if a given credit type doesn't exist or it's precision
+// is bigger then the requested exponent.
+func assertCreditTypeExists(ctx context.Context, k EcocreditKeeper, creditType string, exponent uint32) error {
```
ryanchristo, line 78: Maybe `validateCreditType` given we are checking more than whether or not the credit type exists.

ryanchristo (approved):

x/ecocredit/basket/msg_create.go
```diff
@@ -56,9 +56,13 @@ func (m MsgCreate) ValidateBasic() error {
 	return m.Fee.Validate()
 }
 
-// Validate additional validation with access to the state data.
-func (m MsgCreate) Validate(minFee sdk.Coin) error {
-	// TODO: update function to do stateful validation
+// ValidateCreateFee additional validation with access to the state data.
+// minFee must be sorted.
+func ValidateCreateFee(m *MsgCreate, minFee sdk.Coins) error {
```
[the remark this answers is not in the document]
robert-zaremba: so the idea was to make all validation here. I break down the some parts of the Create RPC in the keeper methods. Eg `indexAllowedClasses`.

x/ecocredit/server/basket/create.go
```diff
@@ -3,10 +3,120 @@ package basket
...
+		if c.Abbreviation == creditType {
+			found = true
+			if c.Precision > exponent {
+				return sdkerrors.ErrInvalidRequest.Wrapf(
+					"exponent %d must be >= credit type precision %d",
+					exponent,
+					c.Precision,
+				)
+			}
+			break
```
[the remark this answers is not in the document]
robert-zaremba: good idea.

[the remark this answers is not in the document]
robert-zaremba: done

x/ecocredit/server/basket/create.go
```diff
@@ -3,10 +3,120 @@ package basket
...
+	basketv1 "github.com/regen-network/regen-ledger/api/regen/ecocredit/basket/v1"
+	"github.com/regen-network/regen-ledger/types"
+	"github.com/regen-network/regen-ledger/x/ecocredit"
+	"github.com/regen-network/regen-ledger/x/ecocredit/basket"
 )
 
-func (k Keeper) Create(ctx context.Context, create *baskettypes.MsgCreate) (*baskettypes.MsgCreateResponse, error) {
-	//TODO implement me
-	panic("implement me")
+func (k Keeper) Create(ctx context.Context, msg *basket.MsgCreate) (*basket.MsgCreateResponse, error) {
```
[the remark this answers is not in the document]
robert-zaremba: done

x/ecocredit/server/basket/create.go
```diff
@@ -3,10 +3,120 @@ package basket
...
+	err = k.distKeeper.FundCommunityPool(rgCtx.Context, fee, sender)
+	if err != nil {
+		return nil, err
+	}
+	if err = assertCreditTypeExists(ctx, k.ecocreditKeeper, msg.CreditTypeName, msg.Exponent); err != nil {
+		return nil, err
+	}
+
+	// TODO: need to decide about the denom creation
+	denom := msg.Name
```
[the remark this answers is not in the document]
robert-zaremba: in separate PR - so we  will all align without blocking this PR

x/ecocredit/server/basket/create.go
robert-zaremba: OK

robert-zaremba: done

x/ecocredit/basket/msg_create.go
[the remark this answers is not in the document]
aaronc: in general I would prefer that this code just lives in the `Create` method itself or at least in the same package rather than calling into a separate package. remember that in the future we won't be able to define any code in the api module where the types are generated

x/ecocredit/server/basket/create.go
```diff
@@ -3,10 +3,120 @@ package basket
...
+// is bigger then the requested exponent.
+func assertCreditTypeExists(ctx context.Context, k EcocreditKeeper, creditType string, exponent uint32) error {
+	res, err := k.CreditTypes(ctx, &ecocredit.QueryCreditTypesRequest{})
+	if err != nil {
+		return err
+	}
+
+	found := false
+	for _, c := range res.CreditTypes {
+		if c.Abbreviation == creditType {
```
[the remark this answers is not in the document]
aaronc: yes, but i think this is a mistake. abbreviation is/should be the primary key of credit type. shall we update the `MsgCreate` proto to make this more clear?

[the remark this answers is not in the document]
robert-zaremba: I will update proto to make it clear.

PR #774, opened 2022-02-15
feat(x/ecocredit): remove MsgCreate.prefix and add Basket.name

aaronc:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/basket/v1/state.proto
```diff
@@ -23,17 +24,22 @@ message Basket {
     // basket_denom is the basket bank denom.
     string basket_denom = 2;
 
+    // name is the unique name of the basket specified in MsgCreate. Basket
+    // names must be unique across all credit types and choices of exponent
+    // above and beyond the uniqueness constraint on basket_denom.
+    string name = 3;
```
robert-zaremba, line 30: can we automatically create an "empty" index? Essentially a set of keys, without associating any structure with it through ORM. Otherwise maybe it will make more sense to do it "manually"?

robert-zaremba (approved):
utACK. Personally, I don't think need to employ ORM and adding not needed attribute to the state to create a set.

proto/regen/ecocredit/basket/v1/state.proto
aaronc: what do you mean by empty?

aaronc:
> utACK. Personally, I don't think need to employ ORM and adding not needed attribute to the state to create a set.

what is the issue with that though? this will also allow lookups by name which I think is actually useful

robert-zaremba:
> what is the issue with that though? this will also allow lookups by name which I think is actually useful

1. if we want a lookup by name then ok
2. if we only want to check uniqueness then we don't need to pollute the state.

proto/regen/ecocredit/basket/v1/state.proto
robert-zaremba: if the only thing we want is a set of names, then we would do it by mapping `name -> 1`, without adding the name to the struct and doing `name -> primary key` association.

robert-zaremba: I'm using 1 since we can't use empty value. In go it would be `name -> empty struct`

robert-zaremba:
>  we expect a handful of baskets ever I think (maybe 10-100 if that)

fair enough

>  polluting state. How else would we do it besides adding something to state?

doing mapping `name -> 1` without storing `name` in the Basket object.

technicallyty (approved):

aaronc:
> >  we expect a handful of baskets ever I think (maybe 10-100 if that)
> 
> 
> 
> fair enough
> 
> 
> 
> >  polluting state. How else would we do it besides adding something to state?
> 
> 
> 
> doing mapping `name -> 1` without storing `name` in the Basket object.

You can use the orm to create a table with a single field to do that. But it is kind of nice to be able to relate that to the actual denom IMHO

proto/regen/ecocredit/basket/v1/state.proto
aaronc: we can use an empty array, but the orm has this all covered still

Issue #726, opened 2022-02-07
Bond module


tonda100:
# MVP User Stories

These stories will be implemented in the minimal viable product.

## Emit Bond

As an bond issuer I want to be able to emit (issue) a bond batch, so that I am able to resell the bonds to the investors.

## Sell Bond

As an bond holder I want to be able sell the bond, so that I can liquidate its value.

(Used also for the underwriting - first transfer to the investor based on a contract.)

# Additional User Stories

These stories will be implemented in next phase(s) to acheive full functionality.

## Issue Carbon Credits

As an band AND credit issuer, I want to issue credits connected to the bonds, so there is a proof of link.

*Possibly can be also done by individual credit batch issuance and subsequent linking batch and bond.*

## Mature Bond

As an issuer, I want to be able to mature the bond (finish the bond lifecycle), so the bond holders can liquidate its value.

## Put Bond

As an bond holder, I want to put the bond, so that I can liquidate its value before maturity.

## Call Bond

As an issuer, I want to call the bond, so I can buy from the current bond owner the bonds, before the maturity.

# Attributes

## Bond Emission

Bond emission is single issuance of the bonds of one issuer granted the authority to issue bonds related to one project issued at a single point in time.

<table>
<colgroup>
<col style="width: 33%" />
<col style="width: 33%" />
<col style="width: 33%" />
</colgroup>
<tbody>
<tr class="odd">
<td style="text-align: left;"><p><strong>Attribute name</strong></p></td>
<td style="text-align: left;"><p><strong>Attribute type</strong></p></td>
<td style="text-align: left;"><p><strong>Note</strong></p></td>
</tr>
<tr class="even">
<td style="text-align: left;"><p>Emission denom</p></td>
<td style="text-align: left;"></td>
<td style="text-align: left;"><p>unique identifier of the bond emmision</p></td>
</tr>
<tr class="odd">
<td style="text-align: left;"><p>Bond name</p></td>
<td style="text-align: left;"><p>text</p></td>
<td style="text-align: left;"><p>display name of the bond</p></td>
</tr>
<tr class="even">
<td style="text-align: left;"><p>Issuer</p></td>
<td style="text-align: left;"><p>wallet</p></td>
<td style="text-align: left;"></td>
</tr>
<tr class="odd">
<td style="text-align: left;"><p>Emission value</p></td>
<td style="text-align: left;"><p>number</p></td>
[...]

PR #801, opened 2022-02-24
feat(x/bond): Proto files

horavamartin:
## Description

Closes: #726 

Proto files for the Bond module.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/proto/regen/bond/v1alpha1/genesis.proto
+++ b/proto/regen/bond/v1alpha1/genesis.proto
@@ -1,0 +1,16 @@
+syntax = "proto3";
+package regen.bond.v1alpha1;
+
+import "gogoproto/gogo.proto";
+import "regen/bond/v1alpha1/params.proto";
+import "regen/bond/v1alpha1/types.proto";
+
+
+option go_package = "github.com/regen-network/regen-ledger/x/bond";
+
+// GenesisState defines the bond module's genesis state.
+message GenesisState {
+  Params params = 1 [(gogoproto.nullable) = false];
+
+  repeated BondInfo bond_info = 2;
+}
--- a/proto/regen/bond/v1alpha1/query.proto
+++ b/proto/regen/bond/v1alpha1/query.proto
@@ -1,0 +1,46 @@
+syntax = "proto3";
+package regen.bond.v1alpha1;
+
+import "gogoproto/gogo.proto";
+import "google/api/annotations.proto";
+import "cosmos/base/query/v1beta1/pagination.proto";
+import "regen/bond/v1alpha1/params.proto";
+import "regen/bond/v1alpha1/types.proto";
+
+option go_package = "github.com/regen-network/regen-ledger/x/bond";
+
+// Query defines the gRPC query service.
+service Query {
+
+  // Parameters queries the parameters of the module.
+  rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
+    option (google.api.http).get = "/regen/bond/v1alpha1/params";
+  }
+
+  // Parameters queries the parameters of the module.
+  rpc Bonds(QueryBondsRequest) returns (QueryBondsResponse) {
+    option (google.api.http).get = "/regen/bond/v1alpha1/bonds";
+  }
+}
+
+// QueryParamsRequest is request type for the Query/Params RPC method.
+message QueryParamsRequest {}
+
+// QueryParamsResponse is response type for the Query/Params RPC method.
+message QueryParamsResponse {
+  // Params holds all the parameters of this module.
+   Params params = 1 [(gogoproto.nullable) = false];
+}
+
+message QueryBondsRequest {
+  // Pagination defines the pagination in the response.
+  cosmos.base.query.v1beta1.PageRequest pagination = 1;
+}
+
+message QueryBondsResponse {
+  // Bonds info.
+  repeated BondInfo bonds = 1;
+
+  // Pagination defines the pagination in the response.
+  cosmos.base.query.v1beta1.PageResponse pagination = 2;
+}
--- a/proto/regen/bond/v1alpha1/query.proto
+++ b/proto/regen/bond/v1alpha1/query.proto
@@ -28,16 +28,21 @@ message QueryParamsRequest {}
 
 // QueryParamsResponse is response type for the Query/Params RPC method.
 message QueryParamsResponse {
+
   // Params holds all the parameters of this module.
    Params params = 1 [(gogoproto.nullable) = false];
 }
 
+// QueryBondsRequest is request type for the Query/Bonds RPC method.
 message QueryBondsRequest {
+
   // Pagination defines the pagination in the response.
   cosmos.base.query.v1beta1.PageRequest pagination = 1;
 }
 
+// QueryBondsResponse is response type for the Query/Bonds RPC method.
 message QueryBondsResponse {
+
   // Bonds info.
   repeated BondInfo bonds = 1;
 
--- a/proto/regen/bond/v1alpha1/tx.proto
+++ b/proto/regen/bond/v1alpha1/tx.proto
@@ -1,0 +1,85 @@
+syntax = "proto3";
+package regen.bond.v1alpha1;
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/timestamp.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+option go_package = "github.com/regen-network/regen-ledger/x/bond";
+
+// Msg defines the Msg service.
+service Msg {
+
+  // IssueBond issues a new bond
+  rpc IssueBond(MsgIssueBond) returns (MsgIssueBondResponse);
+
+  // SellBond sells a bond to a new owner
+  rpc SellBond(MsgSellBond) returns (MsgIssueBondResponse);
+}
+
+// MsgIssueBond is the Msg/IssueBond request type.
+message MsgIssueBond {
+
+  // emission_denom is the unique id of emission
+  string emission_denom = 1;
+
+  // bond name
+  string name = 2;
+
+  // account address of current holder of the bond
+  string holder = 3;
+
+  // face value
+  string face_value = 4;
+
+  // currency of the face value
+  string face_currency = 5;
+
+  // date of issuance
+  google.protobuf.Timestamp issuance_date = 6 [(gogoproto.stdtime) = true];
+
+  // date of maturity
+  google.protobuf.Timestamp maturity_date = 7 [(gogoproto.stdtime) = true];
+
+  // coupon rate
+  string coupon_rate = 8;
+
+  // coupon frequency
+  string coupon_frequency = 9;
+
+  // project code
+  string project = 10;
+
+  // metadata encoded in base64
+  string metadata = 11;
+}
+
+// MsgIssueBondResponse is the Msg/IssueBond response type.
+message MsgIssueBondResponse {
+
+  // id is the unique id of the bond
+  string id = 1;
+}
+
+// MsgIssueBond is the Msg/IssueBond request type.
+message MsgSellBond {
+
+  // id is the unique id of the bond to sell
+  string id = 1;
+
+  // buyer of the bond
+  string buyer = 2;
+
+  // amount of bond to sell
+  string amount = 3;
+
+  // holder of the bond
+  string holder = 4;
+}
+
+// MsgSellBondResponse is the Msg/MsgSellBond response type.
+message MsgSellBondResponse {
+
+  // id is the unique id of the newly created bond
+  repeated string id = 1;
+}
--- a/proto/regen/bond/v1alpha1/types.proto
+++ b/proto/regen/bond/v1alpha1/types.proto
@@ -1,0 +1,62 @@
+syntax = "proto3";
+package regen.bond.v1alpha1;
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/regen-network/regen-ledger/x/bond";
+
+// BatchInfo represents the high-level on-chain information for a bond.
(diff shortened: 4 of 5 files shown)


proto/regen/bond/v1alpha1/tx.proto
```diff
@@ -0,0 +1,84 @@
...
+  string coupon_rate = 8;
+
+  // coupon frequency
+  string coupon_frequency = 9;
+
+  // project code
+  string project = 10;
+
+  // metadata encoded in base64
+  string metadata = 11;
```
ryanchristo, line 53:
```suggestion
  // metadata encoded in base64
  bytes metadata = 11;
```

proto/regen/bond/v1alpha1/types.proto
```diff
@@ -0,0 +1,85 @@
+syntax = "proto3";
+package regen.bond.v1alpha1;
+
+option go_package = "github.com/regen-network/regen-ledger/x/bond";
+
+// BatchInfo represents the high-level on-chain information for a bond.
+message BondInfo {
```
ryanchristo, line 7:
```suggestion
// BondInfo represents the high-level on-chain information for a bond.
message BondInfo {
```

proto/regen/bond/v1alpha1/types.proto
```diff
@@ -0,0 +1,85 @@
...
+  CouponFrequency coupon_frequency = 12;
+
+  // bond status
+  Status status = 13;
+
+  // project code to which the bond belongs
+  string project = 14;
+
+  // metadata encoded in base64
+  string metadata = 15;
```
ryanchristo, line 52:
```suggestion
  bytes metadata = 15;
```

proto/regen/bond/v1alpha1/types.proto
```diff
@@ -0,0 +1,85 @@
...
+  string face_currency = 7;
+
+  // date of bond creation
+  string creation_date = 8;
+
+  // date of issuance
+  string issuance_date = 9;
+
+  // date of maturity
+  string maturity_date = 10;
```
ryanchristo, line 37:
```suggestion
  // date of maturity
  google.protobuf.Timestamp maturity_date = 10 [(gogoproto.stdtime) = true];
```

proto/regen/bond/v1alpha1/types.proto
```diff
@@ -0,0 +1,85 @@
...
+  string face_value = 6;
+
+  // currency of the face value
+  string face_currency = 7;
+
+  // date of bond creation
+  string creation_date = 8;
+
+  // date of issuance
+  string issuance_date = 9;
```
ryanchristo, line 34:
```suggestion
  // date of issuance
  google.protobuf.Timestamp issuance_date = 9 [(gogoproto.stdtime) = true];
```

proto/regen/bond/v1alpha1/types.proto
```diff
@@ -0,0 +1,85 @@
...
+  string parent_bond = 5;
+
+  // face value
+  string face_value = 6;
+
+  // currency of the face value
+  string face_currency = 7;
+
+  // date of bond creation
+  string creation_date = 8;
```
ryanchristo, line 31:
```suggestion
  // date of bond creation
  google.protobuf.Timestamp creation_date = 8 [(gogoproto.stdtime) = true];
```

ryanchristo: Also, is it necessary to store the creation date? The transaction will have a timestamp.

proto/regen/bond/v1alpha1/tx.proto
```diff
@@ -0,0 +1,84 @@
...
+
+  // buyer of the bond
+  string buyer = 2;
+
+  // amount of bond to sell
+  string amount = 3;
+
+  // holder of the bond
+  string holder = 4;
+}
```
ryanchristo, line 77: I'm not entirely clear on how this will work without an order of some kind being stored on chain. A seller would need to create an order and a buyer would need to be able to fulfill that order.

ryanchristo (commented):
Looking good. A few comments and suggestions to start.

aaronc:
The core thing I think we need to get clear on is whether a bond token can simply be an `x/bank` token or not, in which case `x/bond` just needs to store some basic metadata.

proto/regen/bond/v1alpha1/genesis.proto
```diff
@@ -0,0 +1,20 @@
...
+
+
+option go_package = "github.com/regen-network/regen-ledger/x/bond";
+
+// GenesisState defines the bond module's genesis state.
+message GenesisState {
+
+  // Params contains the updateable global parameters for use with the x/params
+  // module
+  Params params = 1 [(gogoproto.nullable) = false];
```
robert-zaremba, line 16: No point to create an empty type here.

proto/regen/bond/v1alpha1/types.proto
robert-zaremba, line 7: Can we call it Bond, rather than Bondinfo?

proto/regen/bond/v1alpha1/genesis.proto
```diff
@@ -0,0 +1,20 @@
...
+
+  // bond_info is the list of bond info.
+  repeated BondInfo bond_info = 2;
```
robert-zaremba, line 19:
```suggestion
  repeated BondInfo bonds = 2;
```

proto/regen/bond/v1alpha1/query.proto
```diff
@@ -0,0 +1,51 @@
...
+  // Parameters queries the parameters of the module.
+  rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
+    option (google.api.http).get = "/regen/bond/v1alpha1/params";
+  }
+
+  // Parameters queries the parameters of the module.
+  rpc Bonds(QueryBondsRequest) returns (QueryBondsResponse) {
+    option (google.api.http).get = "/regen/bond/v1alpha1/bonds";
+  }
+}
```
robert-zaremba, line 24: We should be able to query a bond by ID

proto/regen/bond/v1alpha1/tx.proto
```diff
@@ -0,0 +1,84 @@
...
+
+  // SellBond sells a bond to a new owner
+  rpc SellBond(MsgSellBond) returns (MsgSellBondResponse);
+}
+
+// MsgIssueBond is the Msg/IssueBond request type.
+message MsgIssueBond {
+
+  // emission_denom is the unique id of emission
+  string emission_denom = 1;
```
robert-zaremba, line 23: how about calling it `bond_denom` instead introducing new entity (emission)?

robert-zaremba: or maybe  `bond_id`? .... but wait... maybe it's worth to consider it as an NFT? Let's move that discussion to a  discussion: https://github.com/regen-network/regen-ledger/discussions/815

proto/regen/bond/v1alpha1/tx.proto
```diff
@@ -0,0 +1,84 @@
...
+
+  // bond name
+  string name = 2;
+
+  // account address of current holder of the bond
```
robert-zaremba, line 28:
```suggestion
  // account address of current bond holder
```

proto/regen/bond/v1alpha1/tx.proto
robert-zaremba: If it's base64 encoded, then it should be string. 
BTW: with proto, we don't need to do base64 encoding, so using bytes directly should work.

proto/regen/bond/v1alpha1/tx.proto
robert-zaremba: yeah, or kind of a multisig: both seller and buyer need to sign the message.

Spec is missing to validate this design.

proto/regen/bond/v1alpha1/types.proto
```diff
@@ -0,0 +1,85 @@
+syntax = "proto3";
+package regen.bond.v1alpha1;
+
+option go_package = "github.com/regen-network/regen-ledger/x/bond";
+
+// BatchInfo represents the high-level on-chain information for a bond.
+message BondInfo {
+
+  // id of the bond
+  string id = 1;
```
robert-zaremba, line 10: How the `id` will be constructed? If this will be a sequence number, then uint64 should be sufficient.

proto/regen/bond/v1alpha1/types.proto
```diff
@@ -0,0 +1,85 @@
...
+  string creation_date = 8;
+
+  // date of issuance
+  string issuance_date = 9;
+
+  // date of maturity
+  string maturity_date = 10;
+
+  // coupon rate
+  string coupon_rate = 11;
```
robert-zaremba, line 40: please add more info, how the rate is encoded? Is it a decimal number (we have Dec type)? Or an integer?

proto/regen/bond/v1alpha1/types.proto
```diff
@@ -0,0 +1,85 @@
...
+    STATUS_UNSPECIFIED = 0;
+
+    // Active
+    STATUS_ACTIVE = 1;
+
+    // Closed
+    STATUS_CLOSED = 2;
+
+    // Matured
+    STATUS_MATURED = 3;
```
robert-zaremba, line 67: I don't think we need to prefix each value name with `STATUS_`: https://developers.google.com/protocol-buffers/docs/proto3#enum

proto/regen/bond/v1alpha1/types.proto
```diff
@@ -0,0 +1,85 @@
...
+    COUPON_FREQUENCY_UNSPECIFIED = 0;
+
+    // Monthly
+    COUPON_FREQUENCY_MONTHLY = 1;
+
+    // Quarterly
+    COUPON_FREQUENCY_QUARTERLY = 2;
+
+    // Annually
+    COUPON_FREQUENCY_ANNUALLY = 3;
```
robert-zaremba, line 83: ditto

proto/regen/bond/v1alpha1/query.proto
```diff
@@ -0,0 +1,51 @@
...
+import "gogoproto/gogo.proto";
+import "google/api/annotations.proto";
+import "cosmos/base/query/v1beta1/pagination.proto";
+import "regen/bond/v1alpha1/params.proto";
+import "regen/bond/v1alpha1/types.proto";
+
+option go_package = "github.com/regen-network/regen-ledger/x/bond";
+
+// Query defines the gRPC query service.
+service Query {
```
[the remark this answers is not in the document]
horavamartin: Added query for a single bond.

proto/regen/bond/v1alpha1/types.proto
horavamartin: I have added the `STATUS_` prefix because the Lint check failed:
```
{"path":"proto/regen/bond/v1alpha1/types.proto","start_line":60,"start_column":5,"end_line":60,"end_column":12,"type":"ENUM_VALUE_PREFIX","message":"Enum value name \"MATURED\" should be prefixed with \"STATUS_\"."} 
```

proto/regen/bond/v1alpha1/types.proto
horavamartin: There was also a lint error:
```
{"path":"proto/regen/bond/v1alpha1/types.proto","start_line":66,"start_column":5,"end_line":66,"end_column":13,"type":"ENUM_VALUE_PREFIX","message":"Enum value name \"ANNUALLY\" should be prefixed with \"COUPON_FREQUENCY_\"."}
```

proto/regen/bond/v1alpha1/types.proto
horavamartin: It is generated as UUID. It is not a number.

proto/regen/bond/v1alpha1/genesis.proto
horavamartin: Removed.

proto/regen/bond/v1alpha1/types.proto
horavamartin: And also in every module in Regen prefixes are used in enum values.

proto/regen/bond/v1alpha1/types.proto
robert-zaremba: OK, let's add a comment to the proto. BTW: why not using a sequence number? I sure uint64 is big enough.

proto/regen/bond/v1alpha1/tx.proto
```diff
@@ -0,0 +1,84 @@
...
+
+  // IssueBond issues a new bond
+  rpc IssueBond(MsgIssueBond) returns (MsgIssueBondResponse);
+
+  // SellBond sells a bond to a new owner
+  rpc SellBond(MsgSellBond) returns (MsgSellBondResponse);
+}
+
+// MsgIssueBond is the Msg/IssueBond request type.
+message MsgIssueBond {
```
robert-zaremba, line 20: Related to https://github.com/regen-network/regen-ledger/discussions/815#discussioncomment-2305205,
Let's consider the following alternative for this RPC:
* `MsgNewBondEmission` - will create new unique denom D and N bonds with denom D and precision = 0 (exponent = 0). It will also create Emission object, storing the bond value V, face currency, maturity, interest rate etc... 
* Instead of `MsgIssueBond` we can do `MsgPrimaryBondSell` -> selling X bond form given emission to the borrower. It will fail if the amount of bonds left is smaller than X. Borrower has to borrow `X*V` face currency, to the issuer.

proto/regen/bond/v1alpha1/types.proto
horavamartin: We could use sequence number. Would be better?

Issue #829, opened 2022-03-03
Convert metadata fields from bytes to string


ryanchristo:
## Summary

When querying credit classes, we should display the metadata as a decoded string rather than encoded bytes.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #863, opened 2022-03-08
refactor(x/ecocredit): change metadata to raw string 

technicallyty:
## Description

refactors core metadata to a raw string. 

Closes: #829 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


aaronc:
The idea was actually to change the field type to string not just re-encode

types/metadata.go
```diff
@@ -0,0 +1,14 @@
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+func DecodeMetadata(metadataStr string) ([]byte, error) {
+	b, err := base64.StdEncoding.DecodeString(metadataStr)
+	if err != nil {
+		return nil, sdkerrors.ErrInvalidRequest.Wrap("metadata is malformed, proper base64 string is required")
+	}
+	return b, nil
+}
```
ryanchristo, line 14: Might not be necessary but should we add a test for this?

x/ecocredit/server/core/create_batch_test.go
```diff
@@ -42,7 +42,7 @@ func TestCreateBatch_Valid(t *testing.T) {
 				RetiredAmount:  "3.4",
 			},
 		},
-		Metadata:  nil,
+		Metadata: "",
```
ryanchristo, line 45:
```suggestion
		Metadata:  "",
```

x/ecocredit/server/core/create_batch_test.go
```diff
@@ -100,7 +100,7 @@ func TestCreateBatch_BadPrecision(t *testing.T) {
 				TradableAmount: "10.1234567891111",
 			},
 		},
-		Metadata:  nil,
+		Metadata: "",
```
ryanchristo, line 103:
```suggestion
		Metadata:  "",
```

x/ecocredit/server/core/update_class_test.go
```diff
@@ -194,21 +194,21 @@ func TestUpdateClass_Metadata(t *testing.T) {
...
 		ClassId:  "C01",
-		Metadata: []byte("barfoo"),
+		Metadata: "barfoo",
 	})
 	assert.NilError(t, err)
 
 	class, err := s.stateStore.ClassInfoStore().Get(s.ctx, 1)
 	assert.NilError(t, err)
-	assert.DeepEqual(t, []byte("barfoo"), class.Metadata)
+	assert.DeepEqual(t, "barfoo", class.Metadata)
```
ryanchristo, line 211:
```suggestion
	assert.Equal(t, "barfoo", class.Metadata)
```

ryanchristo (commented):
Looking good. Should we add query tests like you did for class info and classes but for batch info, batches, project info, and projects? Also, we need to decode in the CLI commands.

```
$ ./build/regen q ecocredit classes
classes:
- admin: regen16md38uw5z9v4du2dtq4qgake8ewyf36u6qgfza
  class_id: C01
  credit_type:
    abbreviation: C
    name: carbon
    precision: 6
    unit: metric ton CO2 equivalent
  issuers:
  - regen16md38uw5z9v4du2dtq4qgake8ewyf36u6qgfza
  metadata: cmVnZW4=
  num_batches: "0"
pagination:
  next_key: null
  total: "0"
```

types/metadata.go
technicallyty: added

ryanchristo:
> Looking good. Should we add query tests like you did for class info and classes but for batch info, batches, project info, and projects? Also, we need to decode in the CLI commands.

Just took another look and it looks like these two items have not been address. More concerned with the CLI query commands than the tests but the test might be nice to have.

x/ecocredit/client/tx.go
```diff
@@ -107,7 +107,7 @@ Parameters:
 			if args[2] == "" {
 				return errors.New("base64_metadata is required")
 			}
-			b, err := decodeMetadata(args[2])
+			b, err := types.DecodeMetadata(args[2])
```
blushi, line 110: why do we need to decode the metadata now?
it can just be provided as any type of string and doesn't need to be decoded since it'll be stored as string too

blushi (changes requested):

x/ecocredit/client/tx.go
ryanchristo: This was a misunderstanding on my part and it was not clear in the issue. I was under the impression that we wanted to display the decoded metadata and was not aware that the intention was to support any type of string as well. The issue was proposed when Mark was asking about which credit class was which and shared class query results.

For example...

```
class:
  ...
  metadata: cmVnZW4=
```

Would be...

```
class:
  ...
  metadata: regen
```

Do you think it would still be valuable to decode the metadata if base64 encoded string and return string otherwise? Or are you proposing that we display the base64 encoded string? cc @blushi @aaronc

blushi: I believe we shouldn't have any based64 encoding/decoding involved anymore now that we just deal with strings 
just as for the metadata fields in x/group and x/gov

technicallyty: so the changes in this PR just affect the new V1/ORM stuff, and so far none of the client CLI stuff has been updated for this (CLI still targeting v1alpha2). i'll open a tracking issue for this

technicallyty: https://github.com/regen-network/regen-ledger/issues/876

blushi: oh ok I see

technicallyty: i will update the core however to get rid of decoding, i also had thought the point was to keep base64 strings at the storage level, and decode at the query level. will update shortly

blushi: yeah indeed we don't need the decoding in the queries

aaronc:
Can we change the PR title?

ryanchristo (approved):

blushi (approved):

Issue #821, opened 2022-03-03
Implement escrow model for sell orders


ryanchristo:
## Summary

For now, we have decided to implement an escrow model for sell orders. At a future date, we might want to consider an escrow module for buy orders but this is out-of-scope for this issue.

We will also need to add a field to the balance in the proto files for "escrowed balance" or "in sell orders". The total balance will include the escrow balance and the tradable balance.

PR #878, opened 2022-03-12
feat(marketplace): escrow sell order credits

technicallyty:
## Description

- adds `Escrowed` field to both `BatchBalance` and `BatchSupply`
- updates `Sell` to escrow the credits in the order
- updates `Sell` test to check balances/supplies are updated
- updates `Sell` to emit `EventSell`

Closes: #821 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


ryanchristo (changes requested):
The escrowed credits need to be returned to the seller upon expiration: https://github.com/regen-network/regen-ledger/blob/master/x/ecocredit/server/expiration.go#L23-L35. Also, we should add a message for cancel sell order, allowing the seller to cancel their order at any time to reclaim their escrowed credits, which we can do in a separate pull request (#880). Looking good otherwise.

x/ecocredit/server/marketplace/sell.go
```diff
@@ -92,3 +100,62 @@ func (k Keeper) getOrCreateMarketId(ctx context.Context, creditTypeAbbrev, bankD
...
+	}
+	newEscrowed, err := math.SafeAddBalance(escrowed, sellQty)
+	if err != nil {
+		return fmt.Errorf("escrowed balance: %s, sell order request: %s - %w", escrowed.String(), sellQty.String(), err)
+	}
+	// set the new balance
+	bal.Tradable = newTradable.String()
+	bal.Escrowed = newEscrowed.String()
+
+	// save
```
robert-zaremba, line 132:
```suggestion
```

x/ecocredit/server/marketplace/sell.go
```diff
@@ -92,3 +100,62 @@ func (k Keeper) getOrCreateMarketId(ctx context.Context, creditTypeAbbrev, bankD
...
+		return err
+	}
+	tradable, err := math.NewDecFromString(bal.Tradable)
+	if err != nil {
+		return err
+	}
+	// subtract the desired sell amount from tradable balance
+	newTradable, err := math.SafeSubBalance(tradable, sellQty)
+	if err != nil {
+		return fmt.Errorf("tradable balance: %s, sell order request: %s - %w", tradable.String(), sellQty.String(), err)
```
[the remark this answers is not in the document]
robert-zaremba: Also, if we use `%v` then we don't need to call `String()`

x/ecocredit/server/marketplace/sell.go
```diff
@@ -92,3 +100,62 @@ func (k Keeper) getOrCreateMarketId(ctx context.Context, creditTypeAbbrev, bankD
...
+	supTradable, err = math.SafeSubBalance(supTradable, sellQty)
+	if err != nil {
+		return err
+	}
+	supEscrow, err = math.SafeAddBalance(supEscrow, sellQty)
+	if err != nil {
+		return err
+	}
+	supply.EscrowedAmount = supEscrow.String()
+	supply.TradableAmount = supTradable.String()
```
robert-zaremba, line 159: @aaronc shouldn't we have another type to serialize big numbers in the state? One type for response, another type for state.

robert-zaremba (approved):
utACK. Left few suggestions.

proto/regen/ecocredit/v1/state.proto
```diff
@@ -195,6 +195,9 @@ message BatchBalance {
 
     // retired is the retired amount of credits
     string retired = 4;
+
+    // escrowed is the amount of credits locked up in escrow for the marketplace.
+    string escrowed = 5;
```
aaronc, line 200: Is escrowed there right user facing name? Maybe `in_sell_orders`?

proto/regen/ecocredit/v1/state.proto
```diff
@@ -220,4 +223,8 @@ message BatchSupply {
     // cancelled, effectively undoing the issuance. The sum of total_amount and
     // amount_cancelled will always equal the original credit issuance amount.
     string cancelled_amount = 4;
+
+    // escrowed_amount is the amount of credits in the batch that have been
+    // locked up in escrow for use in the marketplace.
+    string escrowed_amount = 5;
```
aaronc, line 229: Do we really need to have a separate supply here? I guess maybe that's a useful invariant?

x/ecocredit/server/marketplace/sell.go
aaronc: I don't think so. State is directly exposed to clients and in the absence of any standard that clients can decode, string is the best choice. If someone defines a standard that's well specified this could change in the future

proto/regen/ecocredit/v1/state.proto
technicallyty: is there a specific concern you have regarding users and the word `escrowed`? i felt like it was a good fit - `in_sell_orders` seems a bit verbose

proto/regen/ecocredit/v1/state.proto
technicallyty: it seemed right to have it as the credits aren't necessarily tradable nor retired. if anything, it might be useful to be able to query a batch to see how many credits are in the market?

proto/regen/ecocredit/v1/state.proto
ryanchristo: I could go either way. `in_sell_orders` is more specific and potentially more user friendly but provided the supporting documentation, `escrowed` seems ok to me as well.

x/ecocredit/server/marketplace/prune.go
```diff
@@ -0,0 +1,87 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	marketplacev1 "github.com/regen-network/regen-ledger/api/regen/ecocredit/marketplace/v1"
+	"github.com/regen-network/regen-ledger/types/math"
+	"google.golang.org/protobuf/types/known/timestamppb"
+	"time"
+)
+
+// PruneSellOrders is an EndBlock function that moves escrowed credits back into their tradable balance and deletes orders
+// that have expired.
+func (k Keeper) PruneSellOrders(ctx context.Context) error {
```
ryanchristo, line 14: Why is this separate from the existing `PruneOrders` that includes both buy and sell orders? And this should be for `BeginBlocker` not `EndBlocker`, correct? This is how `PruneOrders` is wired up at the moment. This is only being called in your test and we are still deleting and not properly returning escrowed credits.

ryanchristo: https://github.com/regen-network/regen-ledger/blob/70d0d40d3ac95d53b9420792cce168e4ccb04649/x/ecocredit/server/expiration.go#L23-L35

ryanchristo (changes requested):
`PruneOrders` was not updated and I believe we want to include pruning of expired buy and sell orders in `BeginBlocker` not `EndBlocker`.

x/ecocredit/server/marketplace/prune.go
technicallyty: > not properly returning escrowed credits.

https://github.com/regen-network/regen-ledger/blob/70d0d40d3ac95d53b9420792cce168e4ccb04649/x/ecocredit/server/marketplace/prune.go#L29 ?

ryanchristo: It's not wired up in `BeginBlocker` like `PruneOrders` is. We are still deleting sell orders without returning escrowed funds.

ryanchristo: The implemention looks good but it needs to be wired up as is `PruneOrders`.

technicallyty: > Why is this separate from the existing PruneOrders

not sure im following, is this a concern with where the function is placed?

> that includes both buy 

buy orders are not stored in state at the moment, so im not sure i see the value in adding them at this time. 

> This is only being called in the added test and we are still deleting and not properly returning escrowed credits. 

well, none of these PR's are truly wired in yet. This is just impl + unit tests. I think that concern should be addressed when we tackle integration tests.

ryanchristo: > is this a concern with where the function is placed?

Not a concern about where the function is placed. `PruneOrders` is currently wired up in `BeginBlocker`.

> well, none of these PR's are truly wired in yet

`PruneOrders` is already wired up in `BeginBlocker`.

> buy orders are not stored in state at the moment

Correct but the pruning buy orders was already implemented. We can delete if need be or comment out.

ryanchristo: https://github.com/regen-network/regen-ledger/blob/70d0d40d3ac95d53b9420792cce168e4ccb04649/x/ecocredit/server/expiration.go#L23-L35

proto/regen/ecocredit/v1/state.proto
aaronc: How about you make a call @ryanchristo

x/ecocredit/server/marketplace/prune.go
ryanchristo: Thanks for the clarification in discord. Sounds like we have more migrations to do for what I'm requesting.

proto/regen/ecocredit/v1/state.proto
ryanchristo: Ok, `escrowed` it is. More aligned with `tradable` and `retired`.

ryanchristo (approved):
Ok, unblocking and approving now that we are tracking in #898. Nice work!

Issue #728, opened 2022-02-08
ecocredit refactor follow-ups


technicallyty:
## Summary

after #696 we will need to:

- [x] refactor marketplace code - @technicallyty 
- - [x] this should probably include refactoring some of the proto endpoints as we've separated the packages. currently we have `proto/regen/ecocredit/market/v1beta1/` and the endpoints are `regen/ecocredit/v1beta1/sell-orders` but these should be changed to `regen/ecocredit/market/v1beta1/sell-orders`
- [ ] refactor server integration tests - #995
- [x] Explicitly set fees in MsgCreateClass #912

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ryanchristo:
We have separate tracking issues for "refactor simulations" (#920 and #921) so I removed it from here. We should be able to "refactor operations" and "use index delete in expired orders pruning" at this point.

I'm not sure I understand what needs to be done for "refactor tests" - is this related to rewriting acceptance tests using gherkin syntax and then implementing those tests using gocuke? If so, we have separate issues tracking the initial rewriting that needs to be done (#893, #894, #895, #896) and we can open up followup issues for implementing once we have the acceptance tests rewritten. We should also prioritize migrations and features so that we are ready for a v4.0 upgrade if we need added support for bridging polygon assets sooner than later. We could also do some of this work during the audit/testing phase. I think it would be safe to remove "refactor tests" from the list here unless you had something else in mind.

technicallyty:
> I'm not sure I understand what needs to be done for "refactor tests" 

yeah this was a pretty early issue in the re-design/re-factoring, i believe my thoughts at the time were around unit testing. but with that mostly done on a per-PR basis, i think the biggest refactoring of tests would be cleaning up our current integration tests. i imagine things will be broken once the submodules are wired up to the main module

PR #912, opened 2022-03-19
feat(ecocredit): specify fee in `CreateClass`

technicallyty:
## Description

- adds fee to CreateClass

ref #728 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/server/core/create_class.go
```diff
@@ -32,8 +32,20 @@ func (k Keeper) CreateClass(goCtx context.Context, req *core.MsgCreateClass) (*c
...
+	if !ok {
+		return nil, sdkerrors.ErrInvalidRequest.Wrapf("could not convert %s to %T", req.Fee.Amount, sdk.Int{})
+	}
+	if reqFeeAmt.LT(feeAmt) {
+		return nil, sdkerrors.ErrInsufficientFee.Wrapf("expected %v for fee, got %v", feeAmt, reqFeeAmt)
+	}
+
 	// Charge the admin a fee to create the credit class
-	err = k.chargeCreditClassFee(sdkCtx, adminAddress, params.CreditClassFee)
+	err = k.chargeCreditClassFee(sdkCtx, adminAddress, sdk.Coins{sdk.Coin{Denom: req.Fee.Denom, Amount: feeAmt}})
```
technicallyty, line 48: force the fee to the amount specified by the params, rather than the amount passed, seemed like the proper thing to do?

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -57,6 +58,10 @@ message MsgCreateClass {
 
   // credit_type_abbrev describes the abbreviation of a credit type (e.g. "C", "BIO").
   string credit_type_abbrev = 4;
+
+  // fee specifies the fee to pay for the class creation.
+  // acceptable fees for creating a class can be found in the governance parameters for the ecocredit module.
```
ryanchristo, line 63:
```suggestion
  // fee specifies the fee to pay for the creation of the credit class.
  // acceptable fees for creating a credit class can be found in the governance parameters for the ecocredit module.
```

x/ecocredit/server/core/create_class.go
```diff
@@ -32,8 +32,20 @@ func (k Keeper) CreateClass(goCtx context.Context, req *core.MsgCreateClass) (*c
...
+	feeAmt := params.CreditClassFee.AmountOf(req.Fee.Denom)
+	if feeAmt.IsZero() {
+		return nil, sdkerrors.ErrInvalidRequest.Wrapf("%s is not allowed to be used in credit class fees", req.Fee.Denom)
+	}
+	reqFeeAmt, ok := sdk.NewIntFromString(req.Fee.Amount)
+	if !ok {
+		return nil, sdkerrors.ErrInvalidRequest.Wrapf("could not convert %s to %T", req.Fee.Amount, sdk.Int{})
+	}
+	if reqFeeAmt.LT(feeAmt) {
+		return nil, sdkerrors.ErrInsufficientFee.Wrapf("expected %v for fee, got %v", feeAmt, reqFeeAmt)
```
ryanchristo, line 44: Might be nice to include the denom here. The error is:
```
expected 20 for fee, got 10
```

ryanchristo (approved):
Looks good to me. A couple minor suggestions. Pre-approving.

x/ecocredit/server/core/create_class.go
technicallyty: https://github.com/regen-network/regen-ledger/pull/912/commits/67d36d01d196abb1c9d66d5b8b5a264a277bc41a#diff-d5c0619debc0b72ca4732464b5573adcf46173b2de79598e66d43da5a2c47202R44

added 👍🏻

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -57,6 +58,10 @@ message MsgCreateClass {
 
   // credit_type_abbrev describes the abbreviation of a credit type (e.g. "C", "BIO").
   string credit_type_abbrev = 4;
+
+  // fee specifies the fee to pay for the creation of the credit class.
+  // acceptable fees for creating a credit class can be found in the governance parameters for the ecocredit module.
+  Fee fee = 5;
```
robert-zaremba, line 64: 1. why someone would pay a different fee than the required one? Maybe we can use only `string denom` here and charge the required fee. 
2. If we really need to use both denom and amount, then let's use `Coin` type and remove Fee message. We can make it required (it's very small, so we in fact optimize the performance) to avoid null checks.

proto/regen/ecocredit/v1/types.proto
```diff
@@ -28,3 +28,13 @@ message Params {
...
+
+// Fee represents a bank coin fee.
+message Fee {
+
+  // denom is the denom of the coin (i.e. uregen)
+  string denom = 1;
+
+  // amount is the amount of fees to pay in the specified denom.
+  string amount = 2;
+}
```
robert-zaremba, line 40: We don't need that type. Let's use `Coin` instead.

x/ecocredit/core/msg_create_class.go
```diff
@@ -46,6 +46,17 @@ func (m *MsgCreateClass) ValidateBasic() error {
 		}
 	}
 
+	if m.Fee != nil {
```
robert-zaremba, line 49: ok, if it's required, and a struct is small (even if we use Coin), then we can add option for `nullable = false`.

robert-zaremba (commented):
I would prefer that we use use a bare denom instead of Fee/ Coin. And definitely let's remove the Fee type.

proto/regen/ecocredit/v1/tx.proto
technicallyty: 1. i believe the intent was to ensure users aren't being charged fees that they are not explicitly setting. this was a concern i had also considered, so might be nice to get more opinions in here 

2. i changed it to coin, but im not sure about the nullable thing. pulsar doesn't support that (not sure if thats planned either?). it might be best to leave out the option knowing we we will switch fully to pulsar eventually

x/ecocredit/server/core/create_class.go
robert-zaremba: yes, related to my earlier review. I think we should adjust.

proto/regen/ecocredit/v1/tx.proto
robert-zaremba: since we are automatically adjusting, then maybe let's update the docs if we want to keep this attribute.

robert-zaremba (approved):

Issue #924, opened 2022-03-23
Extend Credit Batch functionality to allow dynamic minting


robert-zaremba:
## Summary

Add possibility to mint new credits once the batch is created.

## Problem Definition

Currently, Batch issuance and initial allocation is defined during Batch creation, and it's not possible to create new credits after that.
If we would like to apply this model for a bridge use case, then it will create additional fragmentation and negatively draw on user experience. 
Let's consider the following scenario. Alice transfers 1 X-TCO2 token (which represent a specific project and vintage, X is and ERC20 smart contract representing that ) to Regen Ledger. 1h later, she is happy to see that everything worked and transfer 100'000 tokens from the same X-TCO2. 2h later Bob transfer 50'000 of same tokens to Regen ledger. While all that tokens were fungible in Polygon, in Regen, in the current model, it would create 3 batches, and the resulting credits would not be fungible between the batches.

## Proposal

1. Create a `rpc BatchMint(MsgBatchMint) returns (MsgBatchMintResponse)` method. Only Credit Class issuer can mint new credits. In practice, authz will be used to authorize other account (a bridge account) to mint. Authorization will have a credit amount limit.
    ```
    message MsgBatchMint {
      // address authorized to issue new credits in a given batch, signer of the msg
     string issuer = 1;
     // issuance are the credits issued in the batch.
     repeated BatchIssuance issuance = 2;
     OriginTx origin_tx= 3;
     // reference note for accounting, will be passed to an event
     string `note` = 3;

      message OriginTx {
        // type of the transaction originating the mint process. Eg: Polygon, Ethereum, Verra,
        string type = 1;
        // the id of a transaction based on a type (tx id, serial number) 
        string id = 2;
      }
    }
    ```

2. Add new attribute to `Batch` (and `MsgCreateBatch`):

    ```
    // maximum amount of credits that can be minted. Optional, when empty then there is no 
    // maximal supply. When the supply (tradeable_amount + retired_amount + escrowed_amount)
    // equals to max_credits the no more credits can be minted.
    // Zero is not a correct value.
    string max_credits.
    ````

    When a batch is locked, no new credits can be minted (this should be the default).

____

#### For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

clevinson:
I think at a high level we first need to ask ourselves if fungibility of credits within a credit vintage is a desired feature.

[...]

PR #937, opened 2022-03-25
feat(x/ecocredit): update batch proto to support dynamic minting

robert-zaremba:
## Description

Closes: #924

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1/state.proto
```diff
@@ -10,224 +10,232 @@ option go_package = "github.com/regen-network/regen-ledger/x/ecocredit/core";
...
+  string retired_amount = 3;
+
+  // cancelled_amount is the number of credits in the batch that have been
+  // cancelled, effectively undoing the issuance. The sum of total_amount and
+  // amount_cancelled will always equal the original credit issuance amount.
+  string cancelled_amount = 4;
+
+  // escrowed_amount is the amount of credits in the batch that have been
+  // locked up in escrow for use in the marketplace.
+  string escrowed_amount = 5;
```
robert-zaremba, line 240: auto formatted

x/ecocredit/go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/tendermint/tendermint v0.34.15
 	github.com/tendermint/tm-db v0.6.7
 	github.com/thanhpk/randstr v1.0.4
+	golang.org/x/mod v0.4.2
```
robert-zaremba, line 24: no idea why this was added

proto/regen/ecocredit/v1/state.proto
ryanchristo: Can we finish the proto lint pull request and make the formatting changes there (#721)?

Some of the files updated in this pull request are not related to the proposed changes (e.g. basket proto files). Also, the space before the colon does not appear to be a standard proto style guideline, is this something you added?

ryanchristo: > there is no space before the colon

You are adding spaces before the colon. Maybe that's what you meant to say.

```
primary_key : {fields : "id", auto_increment : true}
```

If this is the format we want to use, we should make sure this is consistent for everyone pushing code (#721). This is new to everyone else, so would be nice to make sure we are all aligned and using the same style.

clevinson:
Can we reduce this PR to just updated proto files (no codegen files, and no stubs implementing interfaces, etc.)? 

I think if its only a PR with proto files for now then it will be a great companion to the #924 issue so we can talk concretely about what an implementation hypothetically would look like.

robert-zaremba:
Reducing this PR only to proto updates.

proto/regen/ecocredit/v1/state.proto
```diff
@@ -136,6 +136,11 @@ message BatchInfo {
 
   // issuance_date is the timestamp when the credit batch was issued.
   google.protobuf.Timestamp issuance_date = 7;
+
+  // enable_future_minting tells if it's possible to mint new credits in
+  // the future. Once the enable_future_minting is set to false, it can't be
+  // toggled any more.
+  bool enable_future_minting = 8;
 }
```
robert-zaremba, line 144: if we have `enable_future_minting`, `max_supply` is not needed any more. so I removed `max_supply`.

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -165,6 +180,39 @@ message MsgCreateBatchResponse {
...
+  bool enable_future_minting = 5;
+
+  message OriginTx {
+    // type of the transaction originating the mint process. Eg: Polygon,
+    // Ethereum, Verra,
+    string typ = 1;
+    // the id of a transaction based on a type (tx id, serial number)
+    string id = 2;
+  }
+}
```
robert-zaremba, line 211: Both, the `note` and `origin_tx` will service the accounting purpose.

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -23,6 +23,16 @@ service Msg {
   // be distributed to recipients in either tradable or retired form.
   rpc CreateBatch(MsgCreateBatch) returns (MsgCreateBatchResponse);
 
+  // MintBatch issues new token in a given batch.
+  // Currently the issuer must be the Credit Class issuer, or an address
```
aaronc, line 27: The issuer should be the issuer of the original batch. Not any credit class issuer

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -23,6 +23,16 @@ service Msg {
...
+  // autorized by him. The request will fail if the batch enable_future_minting
+  // was previously set to false.
+  // NOTE: this method is only for bridge purpose. It must not be used
```
aaronc, line 30:
```suggestion
  // NOTE: this method is only for bridging credits from other registries. It should not be used
```

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -23,6 +23,16 @@ service Msg {
...
+  // for issuing native credits on Regen. More specifically, we
+  // enable minting more credits in an existing batch, when the batch
```
aaronc, line 32:
```suggestion
  // enable minting more credits in an existing batch when the batch
```

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -23,6 +23,16 @@ service Msg {
...
+  // represents a batch - vintage originally registered in other chain.
```
aaronc, line 33:
```suggestion
  // represents a batch - vintage originally registered an another chain or registry. 
```

aaronc (commented):
Generally looks good. I think minting permissions should be more restrictive

robert-zaremba:
+ changed `enable_future_minting` to `sealed`
+ updated docs
+ added `issuer` to BatchInfo

proto/regen/ecocredit/v1/state.proto
aaronc: I actually think the default should be sealed. So if we want a different name let's use `open`

proto/regen/ecocredit/v1/tx.proto
robert-zaremba: done

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -165,6 +179,38 @@ message MsgCreateBatchResponse {
...
 }
 
+// MsgMintBatch is a request type for MintBatch Msg RPC.
+message MsgMintBatch {
+  // Issuer must equal to the batch.issuer address.
+  // Signer of the msg.
+  string issuer = 1;
+
+  // issuance are the credits issued in the batch.
+  repeated BatchIssuance issuance = 2;
```
clevinson, line 189: If there are multiple `BatchIssuance`s for a given `MsgMintBatch` is it expected that they all have the same `origin_tx`?

We could alternatively evaluate nesting `origin_tx` inside BatchIssuance message, so that an on-chain minting event can batch together multiple issuances (coming from multiple different bridge transactions). This may be suboptimal though, as leveraging events becomes quite cumbersome and annoying if in the current event system if we have multiple event attributes corresponding to different origin bridge transactions for one SDK tx.

Also- do we need to support `origin_tx` information in `MsgCreateBatch()` in cases where a bridge action triggers a new batch creation? Or do we expect that the bridge creator does not mint any credits when they create a new batch for a bridge (so all minting of credits via a bridge always happen through a `MsgMintBatch()` call).

I see a few options:

### All issuances from bridges must happen through `MsgMintBatch()` 

In this case, we don't need `origin_tx` information in `MsgCreateBatch()`, and we should make sure we have some on-chain validation that `MsgCreateBatch()` fails if `sealed == false` and `issuance != nil`.

### Allow issuances from bridges in `MsgCreateBatch()`

In this case, we should add `origin_tx` into `MsgCreateBatch()`

robert-zaremba:
rename sealed -> unlocked

aaronc (approved):

clevinson (commented):
Thanks for these updates @robert-zaremba !

A few suggestions in-line, and additionally I have one larger question about your proposal for utilizing events for accounting purposes.

In our current bridge strategy, the bridge operator is responsible (either manually or via an automated way) for initiating `MintBatch()` to issue newly bridged credits to a batch.  And this operator is the only person who can mint new credits to this batch.

Still, it would be nice to have some safety mechanisms that prevent the bridge operator from accidentally bridging credits twice.  If a bridge operator crashes, and is restarting and pulling a queue of bridge events on Polygon via querying EVM events, then how does the bridge operator ensure that it doesn't mint credits that it has already minted? Certainly it can query events for an `origin_tx` to ensure that a given Polygon Tx has not been bridged over yet, but what if it is communicating with a malicious API node? Or what if the full-node that the bridge is querying is just falling behind in processing blocks (which is quite often a possibility)?

I think it might be preferable to have some on-chain logic that enforces no duplicate mints get created for a given `origin_tx`. I know that it's been discussed in the past the possibility of having the state machine have access to events, but I also remember this being pushed back against by some folks on the SDK team -- so I'm not assuming that this will happen.

Given that, I think it would be more secure and preferable to store `origin_tx` in state (maybe as a simple field in `BatchInfo` (or a separate table entirely) called `origin_txs` that stores a list of all `origin_tx` fields as they get processed, so that when processing a new `MsgMintBatchRequiest()`, the state machine would verify that the given `origin_tx` has not been submitted yet.

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -133,30 +143,34 @@ message MsgCreateBatch {
   // quantified and verified.
   google.protobuf.Timestamp end_date = 6 [ (gogoproto.stdtime) = true ];
 
-  // BatchIssuance represents the issuance of some credits in a batch to a
-  // single recipient.
-  message BatchIssuance {
+  // If unlocked is set to true, we will seal the batch and disable the
```
clevinson, line 146:
```suggestion
  // If unlocked is set to false, we will seal the batch and disable the
```

proto/regen/ecocredit/v1/state.proto
```diff
@@ -136,6 +136,14 @@ message BatchInfo {
...
   // issuance_date is the timestamp when the credit batch was issued.
   google.protobuf.Timestamp issuance_date = 7;
+
+  // issuer is an account address which created the batch and which is
+  // authorized to issue more credits if sealed=false.
+  string issuer = 8;
+
+  // sealed tells if it's possible to mint new credits in the future.
+  // Once the sealed is set to true, it can't be toggled any more.
+  bool sealed = 9;
```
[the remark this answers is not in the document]
clevinson: Looks like this was already addressed. thx!

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -23,6 +23,16 @@ service Msg {
   // be distributed to recipients in either tradable or retired form.
   rpc CreateBatch(MsgCreateBatch) returns (MsgCreateBatchResponse);
 
+  // MintBatch issues new token in a given batch.
+  // Currently the issuer must be the Credit Class issuer, or an address
+  // autorized by him. The request will fail if the batch enable_future_minting
+  // was previously set to false.
```
clevinson:
```suggestion
  // autorized by him. The request will fail if the batch's `unlocked` field
  // is set to false.
```

aaronc:
I agree about origin tx uniqueness, but please let's use a join table rather than a list in keeping with 1NF

proto/regen/ecocredit/v1/state.proto
robert-zaremba: It's in the state, so we control the defaults. Good naming is a challenge.
If we want to rename this, then let's use `open` everywhere.

proto/regen/ecocredit/v1/tx.proto
robert-zaremba: > Also- do we need to support origin_tx information in MsgCreateBatch() in cases where a bridge action triggers a new batch creation?

Good point. yes!

robert-zaremba: I've updated the `MsgCreatBatch` - so bridge will create a batch and mint at the same time. It will be consistent with the minting we have now. Subsequent mints will use MsgMintBatch

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -23,6 +23,16 @@ service Msg {
...
 
+  // MintBatch issues new token in a given batch.
+  // The issuer must be the account who created (or delegated using x/authz),
+  // the batch.
+  // The request will fail if the batch is sealed.
+  // NOTE: this method is only for bridge purpose. It must not be used
+  // for issuing native credits on Regen. More specifically, we
+  // enable minting more credits in an existing batch, when the batch
+  // represents a vintage originally registered in another chain.
+  rpc MintBatch(MsgMintBatch) returns (MsgMintBatchResponse);
```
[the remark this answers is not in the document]
robert-zaremba: yes, doesn't sound very English. How about `MintBatchCredits`?

robert-zaremba:
+ updated events (see comment in PR)
+ added orig tx index
+ added OrigTx and note to `MsgCreateBatch`
+ rename MintBatch to MintBatchCredits
+ use open instead of unlocked.

proto/regen/ecocredit/v1/types.proto
```diff
@@ -34,3 +34,13 @@ message Params {
...
 }
+
+// OriginTx is a reference to an external transaction or an operation
+// related to an action on Regen Ledger.
+message OriginTx {
+  // type of the transaction originating the mint process. Eg: Polygon,
+  // Ethereum, Verra,
+  string typ = 1;
+  // the id of a transaction based on a type (tx id, serial number)
+  string id = 2;
```
robert-zaremba, line 45: Normally I would prefer to  not storing everything. We have events to keep that information. But events are not guaranteed to be permanent. 
So ideally we don't store everything here and we sort out the events in SDK.

What do you think?

proto/regen/ecocredit/v1/state.proto
```diff
@@ -236,3 +244,24 @@ message BatchSupply {
...
+  string tx_id = 1;
+
+  // type of the transaction originating the mint process. Eg: Polygon,
+  // Ethereum, Verra,
+  string typ = 2;
+
+  string note = 3;
+
+  // batch_denom is the unique ID of the credit batch.
+  string batch_denom = 4;
```
robert-zaremba, line 266: Normally I would prefer to not storing everything. We have events to keep that information. But events are not guaranteed to be permanent.
So ideally we don't store everything here and we sort out the events in SDK.

What do you think?

robert-zaremba: in fact, since I'm adding this fields here, I've slimmed down the `EventMintBatchCredits` event

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -23,6 +24,17 @@ service Msg {
...
+  // MintBatchCredits issues new token in a given batch.
+  // The issuer must be the account who created (or delegated using x/authz),
+  // the batch.
+  // The request will fail if the batch is not open.
+  // NOTE: this method is only for bridge purpose. It must not be used
+  // for issuing native credits on Regen. More specifically, we
+  // enable minting more credits in an existing batch, when the batch
+  // represents a vintage originally registered in another chain.
+  rpc MintBatchCredits(MsgMintBatchCredits)
+      returns (MsgMintBatchCreditsResponse);
```
ryanchristo, line 36: Can you comment this line out for the time being and then we can uncomment when implementing (#973). Otherwise anyone who makes proto changes will face errors because this message is not yet implemented.

ryanchristo:
```suggestion
  // rpc MintBatchCredits(MsgMintBatchCredits)
      returns (MsgMintBatchCreditsResponse);
```

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -133,30 +145,43 @@ message MsgCreateBatch {
...
+  string note = 9;
+}
 
-    // retired_amount is the number of credits in this issuance that are
-    // effectively retired by the issuer on receipt. Decimal values are
-    // acceptable.
-    string retired_amount = 3;
+// BatchIssuance represents the issuance of some credits in a batch to a
+// single recipient.
+message BatchIssuance {
```
ryanchristo, line 164: This will also break tests:

```
core/msg_create_batch_test.go:32:18: undefined: MsgCreateBatch_BatchIssuance
```

Can you update any references of `MsgCreateBatch_BatchIssuance` to `BatchIssuance` before merging?

ryanchristo:
When writing the proto API for new features, we should either include dummy implementation methods or comment out any new rpc methods that will cause tests to fail after generating the proto files.

Ideally changes to the existing API would not be made in the same pull request, which would prevent the need to update any code in the existing implementation. In cases like this where `BatchIssuance` is being pulled out of `BatchInfo`, we should generate the proto files and make sure the necessary changes are made in the existing implementation.

A potential process for designing new features starting with proto API updates:

1. a pull request is created with only proto API definitions
2. the pull request is reviewed and the design is approved
3. proto files are generated and, if necessary, minimal changes are made to the existing implementation, which will allow proto files to be generated in other pull requests without breaking tests

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -165,6 +190,34 @@ message MsgCreateBatchResponse {
...
+  // batch_denom is the unique ID of the credit batch.
+  string batch_denom = 2;
+
+  // issuance are the credits issued in the batch.
+  repeated BatchIssuance issuance = 3;
+
+  // open=true keeps the batch unlocked for future minting.
+  // Othrwise we will seal the batch after the mint.
+  // Throws an error if the batch is already sealed.
+  bool open = 4;
```
clevinson, line 208: Could we update this to actually be a separate message? I think in most use cases (thinking of bridges), the issuer won't actually ever want to seal the batch.  Adding this here requires the user to explicitly set this value every time they call `MsgMintBatchCredits`, and if they forget, then it would lock the batch and prevent the issuer from being able to issue more credits to that vintage (assuming we have the appropriate restrictions on project <> credit type duplicate issuances).

I think I'd rather see something like:
```proto
message MsgSealBatch {
  string issuer = 1;
}
```

clevinson: @aaronc I know you had initially proposed this `bool open = 4;` what do you think of this upgrade?

aaronc: I don't think it makes that big of a difference. This should all be done automatically anyway. But I don't object

clevinson:
I think the proto gen stubs suggested from @ryanchristo makes sense. 

Im happy to give a concept ACK on the general design now- the only api change i'd still like to see is a MsgSealBatch message separated out. 

@robert-zaremba can you add go stubs and address the related go changes to prevent tests from breaking?

proto/regen/ecocredit/v1/tx.proto
robert-zaremba: Yesterday we had a session with Cory about the process and we found that it would make sense to separate the toggle functionality to a separate message. So if nobody objects I will make the update.

robert-zaremba:
Sorry but we are going back and forth without being constructive. 2 weeks ago I started with basic implementation for Msg methods and wanted to add dummy implementation for the RPC implementation while waiting for reviews (which would prevent troubles when someone generates the code and introduces additional changes). It got a criticism (https://github.com/regen-network/regen-ledger/pull/937#issuecomment-1079395808and and and discord) to ONLY do proto code (so NO code generation and NO other methods). 

The tests are not breaking because this PR doesn't include generated code.

I prefer to merge this PR and do other PR with implementation to not hold on this PR any more.

robert-zaremba:
I've added `MsgSealBatch`

ryanchristo (approved):

clevinson (approved):

clevinson:
Sorry about the back & forth @robert-zaremba, but I also remembered a bit more than dummy implementation in what you originally was putting forward (I think I remember some implementation work as well). Happy to have this merged now, and let's discuss maybe on one of our next dev calls to make sure we're not going back and forth too much with process around proto PRs as design.

glandua:
Great work Robert.  Although I was originally pretty restrained about the dynamic batch approach, I am quite pleased how this has turned out.

robert-zaremba:
@ryanchristo , @clevinson there was:

> basic implementation for Msg methods

robert-zaremba:
thanks for merging! :beers:

Issue #167, opened 2020-11-12
Add additional indexes and queries to x/ecocredit module


aaronc:
It would improve UX to be able to make the following queries:

- [x] credit batches by class - #885
- [x] credit batches by issuer - #955 
- [x] credit classes by admin https://github.com/regen-network/regen-ledger/pull/954
- [x] credit balances by account for all batch denoms #942 

We should do pagination the way it is done in the SDK.

technicallyty:
I've done #885 as a separate RPC endpoint, but im wondering if it might be better to do it all under the existing `Batches` with a oneof for which specific field to query by (projectId, classId, issuer..) thoughts? @ryanchristo @aaronc

ryanchristo:
We should avoid using `oneof` as far as I understand (see #790 and #872).

ryanchristo:
I guess query will not be affected by this though.

ryanchristo:
I think we should keep queries separate but we can discuss on our next product call as part of #909.

ryanchristo:
`issuer` is part of `BatchInfo` in `v1alpha1` (live on chain) and I think it should be part of `v1` as well. Adding issuer will not add another state migration and I think removing it might have been a mistake or misunderstanding.

I would go with (re)adding an issuer field and index to `BatchInfo`.

PR #942, opened 2022-03-25
feat(x/ecocredit): query all credit balances based on address

technicallyty:
## Description

- adds a query to get all credit balances of an address
- changes the response type of the existing balance query to use BatchBalance from state.proto

ref: #167 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1/query.proto
```diff
@@ -244,12 +250,28 @@ message QueryBalanceRequest {
...
+  // pagination defines an optional pagination for the request.
+  cosmos.base.query.v1beta1.PageRequest pagination = 2;
+}
+
+// QueryBalancesResponse is the Query/Balances response type.
+message QueryBalancesResponse {
 
-  // retired_amount is the decimal number of retired credits.
-  string retired_amount = 2;
+  // balances are a list of balances from different credit batches that the user holds.
```
ryanchristo, line 270:
```suggestion
  // balances are a list of balances from different credit batches that the account holds.
```

ryanchristo (approved):
Looks good to me. A couple minor suggestions.

x/ecocredit/server/core/query_balances.go
```diff
@@ -0,0 +1,49 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (k Keeper) Balances(ctx context.Context, req *core.QueryBalancesRequest) (*core.QueryBalancesResponse, error) {
+	acc, err := sdk.AccAddressFromBech32(req.Account)
+	if err != nil {
+		return nil, err
+	}
+
+	pg, err := ormutil.GogoPageReqToPulsarPageReq(req.Pagination)
```
robert-zaremba, line 20: not directly related to this PR, I think this function name should have "Slow" suffix.

x/ecocredit/server/core/query_balances.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	it, err := k.stateStore.BatchBalanceTable().List(ctx, api.BatchBalanceAddressBatchIdIndexKey{}.WithAddress(acc), ormlist.Paginate(pg))
+	if err != nil {
+		return nil, err
+	}
+
+	balances := make([]*core.BatchBalance, 0, 10)
```
robert-zaremba, line 30: why starting with cap = 10?

robert-zaremba (approved):
Looks good. I thought we wanted to do separate PR with proto only ;)

x/ecocredit/server/core/query_balances.go
technicallyty: yeah that makes sense - it does call another `slow` method under the hood

x/ecocredit/server/core/query_balances.go
technicallyty: i was thinking it might be good to add some arbitrary cap to avoid slice allocations. would you advise a smaller/bigger number or just not at all? maybe we could add a `default` constant if we want to define a capacity?

robert-zaremba: No idea what would be the most common one, that's why I asked ;) Maybe let's decrease to 8 and add a comment?
By default, go will double the capacity if it needs.

Issue #678, opened 2022-01-03
Add state migrations for ecocredit module


aleem1314:
## Summary

Add state migrations for ecocredit module from version 1 to 2 (Regen Ledger v3.0 to v4.0).

We need to include migrations for the following:

- batch project
- batch issuance date
- basket curator

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #715, opened 2022-02-01
x/ecocredit v1 Readiness Checklist


aaronc:
## x/ecocredit v1 Module Readiness Checklist

This checklist is to be used for tracking the final internal audit of new Cosmos SDK modules prior to inclusion in a published release.

### Release Candidate Checklist

The following checklist should be gone through once the module has been fully implemented. This audit should be performed directly on `master`, or preferably on a `alpha` or `beta` release tag that includes the module.

The module **should not** be included in any Release Candidate tag until it has passed this checklist.

- [x] API audit (at least 1 person) (@ryanchristo)
  - [x] Are Msg and Query methods and types well-named and organized?
    - #1161
    - #1168
    - #1192
    - #1197
    - #1198
  - [x] Is everything well documented (inline godoc as well as `/spec/` folder in module directory)
- [x] State machine audit (at least 2 people) (@aleem1314, @technicallyty, @ryanchristo).
  - [x] Read through MsgServer code and verify correctness upon visual inspection
  - [x] Ensure all state machine code which could be confusing is properly commented
  - [x] Make sure state machine logic matches Msg method documentation
  - [x] Ensure that all state machine edge cases are covered with tests and that test coverage is sufficient (at least 90% coverage on module code)
      - #1155
      - #1160
      - #1170
      - #1202
  - [x] Assess potential threats for each method including spam attacks and ensure that threats have been addressed sufficiently. This should be done by writing up threat assessment for each method
  - [x] Assess potential risks of any new third party dependencies and decide whether a dependency audit is needed
- [x] State migration audit and testing using a temporary test network (@kaustubhkapatral, @aleem1314)
  - [x] Read through state migration and verify correctness upon visual inspection
  - [x] Spin up a temporary test network and populate with data and then perform upgrade with mainnet chain id
  - [x] Spin up a temporary test network and populate with data and then perform upgrade with redwood chain id
- [x] Completeness audit, fully implemented with tests (at least 1 person) (@aleem1314, @ryanchristo)
  - [x] Genesis import and export of all state
    - #1195
  - [x] Query services
      - #1168
      - #1178
      - #1197
      - #1198
  - [x] CLI methods
    - #1155
    - #1176
    - #1199
  - [x] All necessary migration scripts are present
    - #1180
    - #1185
    - #1188
    - #1205
    - #1209

### Published Release Checklist

[...]

PR #957, opened 2022-03-30
feat(ecocredit): add state migrations

aleem1314:
## Description

Closes: #678 

* migrate credit classes and issuers to ORM v1
* create projects for existing classes
* migrate credit batches to ORM v1
* migrate balance & supply to ORM v1
* migrate credit type, project, batch sequence tables to ORM v1
* wire ecocredit migrations with `server.Manager`.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/migrations/v3/core.go
```diff
@@ -0,0 +1,385 @@
...
+	// TODO: update with actual data
+	projects := []*api.ProjectInfo{
+		{
+			Name:            "P01",
+			ClassId:         1,
+			Admin:           sdk.AccAddress("cosmos154hmhstk3gpkv2ndec8zkjkc5c3svutcqcswne"),
+			ProjectLocation: "AB-CDE FG1 345",
+			Metadata:        "metadata",
+		},
+	}
```
ryanchristo, line 138: Is it possible to iterate over credit batches rather than manually inserting project information?

- For `Name`, we could track the number of projects per credit class and automatically add the project name using the same format when creating projects (i.e. `C0101` is the first project in the `C01` credit class).
- For `ClassId` the batch provides the class id.
- For `Admin`, the batch provides the issuer.
- For `ProjectLocation`, the batch provides the project location.
- For `Metadata`, we could leave blank and maybe we could squeeze in another pull request adding the ability to update the metadata field #979

We would also need to skip adding a new project if all fields (except for the project name) match a project that has already been inserted. Does this sound possible?

x/ecocredit/migrations/v3/core.go
```diff
@@ -0,0 +1,385 @@
...
+			return err
+		}
+
+		bInfo := api.BatchInfo{
+			ProjectId:    projectIDsMap[batchInfo.ClassId],
+			BatchDenom:   batchInfo.BatchDenom,
+			Metadata:     string(batchInfo.Metadata),
+			StartDate:    timestamppb.New(*batchInfo.StartDate),
+			EndDate:      timestamppb.New(*batchInfo.EndDate),
+			IssuanceDate: nil, // TODO: add issuance date
```
ryanchristo, line 205: Is there something blocking this todo?

x/ecocredit/migrations/v3/core.go
```diff
@@ -0,0 +1,385 @@
...
+	store := sdkCtx.KVStore(storeKey)
+	if err = migrateBalances(store, ss, ctx, batchIDsMap); err != nil {
+		return err
+	}
+
+	if err = migrateSupply(store, ss, ctx, batchIDsMap); err != nil {
+		return err
+	}
+
+	// TODO: migrate params if needed #729
```
ryanchristo, line 260: I don't think v0.46 will make it into this release but good to have this here.

x/ecocredit/module/module.go
```diff
@@ -162,7 +162,7 @@ func (a Module) GetTxCmd() *cobra.Command {
 }
 
 // ConsensusVersion implements AppModule/ConsensusVersion.
-func (Module) ConsensusVersion() uint64 { return 2 }
+func (Module) ConsensusVersion() uint64 { return 3 }
```
ryanchristo, line 165: I think we can actually leave this at `2`. We did not use the consensus version in the last upgrade.

https://github.com/regen-network/regen-ledger/blob/v3.0.0/app/stable_appconfig.go#L51-L52

ryanchristo (commented):
This is looking great. I think we can avoid manually adding projects, which may not be necessary for mainnet (no credit batches exist on chain yet) but it would be nice to support for redwood.

x/ecocredit/migrations/v3/core.go
ryanchristo: Oof... Yea, not sure how we should do this. We don't want this to be user controlled and we can't query the event, so we would have to manually update it (...and we would need to do redwood separately). Adding this to #980.

ryanchristo: Oof... Yea, this will probably need to be done manually along with the curator address for baskets. We don't want this to be user controlled and we can't query the event. Added this item to #980.

aleem1314: Yes, This needs to be done manually. Can we leave it empty for existing credit batches?

x/ecocredit/migrations/v3/core.go
aleem1314: > Is it possible to iterate over credit batches rather than manually inserting project information?

Yes. We can do this.

types/module/server/manager.go
```diff
@@ -199,6 +205,21 @@ func (mm *Manager) InitGenesis(ctx sdk.Context, genesisData map[string]json.RawM
 	return res
 }
 
+// InitGenesis performs state migrations for registered modules.
```
aleem1314, line 208:
```suggestion
// RunMigrations performs state migrations for registered modules.
```

ryanchristo (commented):
On our product call today, we discussed how we want to handle state migrations for projects. I still think this automatic migration is a potential approach, which will reduce the amount of manual operations required and support the migrations on redwood (which would require separate manual operations).

There was a concern brought up by @clevinson and @aaronc about the potential of multiple projects with the same location. Although I understand the concern, I don't see this as a major issue given that we are the only credit class creators (and credit issuers) and if we do issue any credits prior to the upgrade, we have the ability to provide distinct project locations for each credit batch.

That being said, we may need to revert the automatic migrations I requested for projects. @clevinson and @aaronc can you confirm? If so, my mistake and I apologize for the extra work.

Edit: For now, it sounds like we want to keep the automatic project migrations but we will also need to perform manual updates following the state migrations for each chain.

x/ecocredit/migrations/v3/core.go
```diff
@@ -0,0 +1,404 @@
...
+			pInfo, err := pItr.Value()
+			if err != nil {
+				return err
+			}
+
+			if pInfo.ClassId == classNameIDMap[batchInfo.ClassId] {
+				projectExists = true
+				projectID = pInfo.Id
+				break
+			}
```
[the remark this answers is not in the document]
ryanchristo: This comment was from before the product call. Pending a response from @clevinson and @aaronc.

types/module/server/manager.go
```diff
@@ -199,6 +205,21 @@ func (mm *Manager) InitGenesis(ctx sdk.Context, genesisData map[string]json.RawM
 	return res
 }
 
+// RunMigrations performs state migrations for registered modules.
+func (mm *Manager) RunMigrations(ctx sdk.Context, cdc codec.Codec) error {
+	for _, h := range mm.migrationHandlers {
```
aaronc, line 210: This needs to be sorted to run deterministically... maybe at this point we should just use the SDK's `AppModule` to not risk these kinds of issues

aaronc (changes requested):

types/module/server/manager.go
ryanchristo: There's only one handler, which is for the ecocredit module. In that case, is this still an issue?

x/ecocredit/migrations/v3/core.go
```diff
@@ -0,0 +1,404 @@
...
+	orm "github.com/regen-network/regen-ledger/orm"
+)
+
+type batchMapT struct {
+	Id              uint64
+	AmountCancelled string
+}
+
+// MigrateState performs in-place store migrations from v3.0 to v4.0.
+func MigrateState(sdkCtx sdk.Context, storeKey storetypes.StoreKey,
```
technicallyty, line 23: should we also include the params migration here? that would close #965. It should just take the existing params, convert them to v1 proto, and then set the default AskDenom param (https://github.com/regen-network/regen-ledger/pull/969).

types/module/server/manager.go
aaronc: Best to just do it now and not have to worry

x/ecocredit/migrations/v3/core.go
aleem1314: > should we also include the params migration here? that would close #965. It should just take the existing params, convert them to v1 proto, and then set the default AskDenom param (#969).

We don't have any breaking changes in params (It is just a version bump). For `allowed_ask_denoms` we can set them in the upgrade handler.

x/ecocredit/migrations/v3/core.go
```diff
@@ -0,0 +1,385 @@
...
+		}
+		if err := ss.ClassSequenceTable().Save(ctx, &api.ClassSequence{
+			CreditType:  ctype.Abbreviation,
+			NextClassId: ctype.SeqNumber,
+		}); err != nil {
+			return err
+		}
+
+		// delete credit type sequence from old store
+		if err := creditTypeSeqTable.Delete(sdkCtx, &ctype); err != nil {
```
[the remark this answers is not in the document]
aaronc: in the SDK orm we can use `DeleteBy` to accomplish the same thing. For the old orm, I don't know

app/stable_appconfig.go
```diff
@@ -56,7 +56,26 @@ func setCustomOrderEndBlocker() []string {
...
+
+	// TODO: set `AllowedAskedDenoms` ecocredit param #969
+	// // run x/ecocredit state migrations
+	// if err := app.smm.RunMigrations(ctx, app.AppCodec()); err != nil {
+	// 	return nil, err
+	// }
+	// toVersion[ecocredit.ModuleName] = ecocreditmodule.Module{}.ConsensusVersion()
+
+	// return toVersion, nil
+	// })
```
ryanchristo, line 76: I think we can hold off on implementing this here and point to #980 as a followup.

types/module/server/manager.go
```diff
@@ -199,6 +206,28 @@ func (mm *Manager) InitGenesis(ctx sdk.Context, genesisData map[string]json.RawM
 	return res
 }
 
+// RunMigrations performs state migrations for registered modules.
+func (mm *Manager) RunMigrations(ctx sdk.Context, cdc codec.Codec) error {
+	keys := make([]string, 0, len(mm.migrationHandlers))
+	for k := range mm.migrationHandlers {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
ryanchristo, line 215: Might be nice to add a comment here as to why we are sorting (i.e. to prevent non-determinism).

x/ecocredit/migrations/v3/core.go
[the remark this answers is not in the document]
ryanchristo: After further discussing the @clevinson on the product call, we decided this would be the best approach to support both mainnet and redwood and then if we needed to perform any additional operations for mainnet (#980), we would handle those updates in the upgrade handler. Once this change is made and the related test is updated, this pull request has my approval.

x/ecocredit/migrations/v3/core.go
```diff
@@ -0,0 +1,404 @@
...
+			}
+
+			name := FormatProjectID(batchInfo.ClassId, projectSeq)
+			id, err := ss.ProjectInfoTable().InsertReturningID(ctx,
+				&api.ProjectInfo{
+					Name:            name,
+					Admin:           admin,
+					ClassId:         classID,
+					ProjectLocation: batchInfo.ProjectLocation,
+					Metadata:        "", // TODO: add metadata
```
ryanchristo, line 191: I think we can remove this todo or replace it with a comment about metadata being handled in the upgrade handler on a per chain basis.

x/ecocredit/migrations/v3/core.go
```diff
@@ -0,0 +1,404 @@
...
+			projectID = id
+		}
+
+		bInfo := api.BatchInfo{
+			ProjectId:    projectID,
+			BatchDenom:   batchInfo.BatchDenom,
+			Metadata:     string(batchInfo.Metadata),
+			StartDate:    timestamppb.New(*batchInfo.StartDate),
+			EndDate:      timestamppb.New(*batchInfo.EndDate),
+			IssuanceDate: nil, // TODO: add issuance date
```
ryanchristo, line 206: I think we can remove this todo or replace it with a comment about issuance date being handled in the upgrade handler on a per chain basis.

ryanchristo (commented):
The updates look good to me. Just one more change requested from me in regards to how the existence of projects is determined. We need to check both the class id and project location when determining if a project already exists.

x/ecocredit/migrations/v3/core.go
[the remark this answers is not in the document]
aleem1314: updated [3cd4bb1](https://github.com/regen-network/regen-ledger/pull/957/commits/3cd4bb11930724de96e955890ddf9eef0d5261d1#diff-f31c5489648094d493af1a8cfe6393dd8292a82f4e0902c5b981d3126bf36aeaR165)

ryanchristo (approved):
Looks good to me. Opened #993 as a followup.

ryanchristo:
@aaronc the change you requested has been made. Is this something we can unblock now?

technicallyty (approved):
looks good to me, nice work! @aleem1314 👍🏻 

do we need an issue to test this manually and/or with integration test? looks like the test is calling the migration directly, so we'll need a way to ensure it's all wired up correctly 😄

ryanchristo:
> do we need an issue to test this manually and/or with integration test? looks like the test is calling the migration directly, so we'll need a way to ensure it's all wired up correctly smile

We won't be able to test manually until we have the upgrade handlers implemented (#980). Also added a section to the ecoredit module readiness checklist for testing state migrations against a temporary test network (#715).

PR #960, opened 2022-03-30
feat(x/ecocredit): introduce core params API

technicallyty:
## Description

- introduces param structure to state.proto
- introduces rpc methods to update params
- places dummy impl in params.go to appease interface guard

Closes: n/a

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] followed the guidelines for [building modules](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules)
- [ ] included the necessary unit and integration [tests](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#testing)
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed `!` in the type prefix if API or client breaking change
- [ ] confirmed all author checklist items have been addressed 
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

PR #962, opened 2022-03-30
feat(x/ecocredit): introduce basket params API

technicallyty:
## Description

- introduces param structure to state.proto
- introduces rpc methods to update params
- places dummy impl in params.go to appease interface guard

Closes: n/a

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/basket/v1/state.proto
```diff
@@ -86,3 +86,19 @@ message BasketBalance {
...
+    id : 13,
+    primary_key : {fields : "denom"}
+  };
+
+  // denom is the denom of the fee required to create a basket.
+  string denom = 1;
+
+  // amount is the amount of the fee required to create a basket.
+  string amount = 2;
+}
```
ryanchristo, line 104: As mentioned in #960 in relation to the credit class fee, I think we only want to support a single fee. We named the parameter `basket_creation_fee` and copied the implementation for credit class fee but I don't actually recall any conversations about supporting multiple fees.

```suggestion
// BasketFee is a valid coin denom and amount that may be used as the fee
// for basket creation.
message BasketFee {
  // This table is controlled via governance.
  option (cosmos.orm.v1alpha1.singleton) = {
    id : 13,
    primary_key : {fields : "denom"}
  };

  // fee is the fee required to create a basket.
  cosmos.base.v1beta1.Coin fee = 1
}
```

proto/regen/ecocredit/basket/v1/tx.proto
```diff
@@ -20,6 +20,10 @@ service Msg {
   // Take takes credits from a basket starting from the oldest
   // credits first.
   rpc Take(MsgTake) returns (MsgTakeResponse);
+
+  // UpdateClassFee is a governance method that allows for the addition and removal of fees to be used for the
+  // basket creation fee.
+  rpc UpdateBasketFee(MsgUpdateBasketFee) returns (MsgUpdateBasketFeeResponse);
```
ryanchristo, line 26:
```suggestion
  // UpdateClassFee a governance method that updates the basket creation fee.
  rpc UpdateBasketFee(MsgUpdateBasketFee) returns (MsgUpdateBasketFeeResponse);
```

proto/regen/ecocredit/basket/v1/tx.proto
```diff
@@ -154,3 +158,20 @@ message MsgTakeResponse {
...
+  // root_address is the address of the signer.
+  // This MUST equal the address of the gov module for the tx to succeed.
+  string root_address = 1;
+
+  // add_fees defines a list of coins to append to the list of fees that can be used in the basket creation fee.
+  repeated cosmos.base.v1beta1.Coin add_fees = 2;
+
+  // remove_fee_denoms defines a list of denoms to remove from the list of basket creation fees.
+  repeated string remove_fee_denoms = 3;
+}
```
ryanchristo, line 174: As mentioned above, I think we want to support a single basket fee and this should update that single fee.

ryanchristo (commented):
As mentioned in #960, it looks like we may not update to Cosmos SDK v0.46 for Regen Ledger v4.0. We can still add the proto updates and the message implementations but we won't be able to migrate until after v4.0.

Depending on what we do for credit class fee, we should do the same here for basket fee.

Also, make sure you run make proto-format.

proto/regen/ecocredit/basket/v1/state.proto
technicallyty: using Coin works as long as it's not the primary key, ORM can't handle imported messages as PK's. i think the denom/amount separation is necessary

robert-zaremba, line 104: Ideally would be if we don't do `a -> a` pattern (saving in a key concatenation of the value)

robert-zaremba (commented):
I would prefer if we avoid `a -> a` pattern

proto/regen/ecocredit/basket/v1/state.proto
aaronc: we don't need a primary key with a singleton. it just has one instance

technicallyty: right - so then we aren't concerned with fees in other denoms? if we're just going with the amount of $regen only, we could just store the amount string

aaronc: We just want one fee (could be multiple coins). So just a singleton with repeated Coins

ryanchristo:
closing in favor of #1349

PR #963, opened 2022-03-30
refactor(x/ecocredit): update market params API

technicallyty:
## Description

- updates the marketplace params proto definitions

Closes: n/a

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/marketplace/v1/tx.proto
```diff
@@ -25,8 +25,8 @@ service Msg {
   // Buy creates credit buy orders.
   rpc Buy(MsgBuy) returns (MsgBuyResponse);
 
-  // AllowAskDenom is a governance operation which authorizes a new ask denom to
-  // be used in sell orders
+  // AllowAskDenom is a governance operation which authorizes new ask denoms and/or removes denoms to
+  // be used in sell orders.
   rpc AllowAskDenom(MsgAllowAskDenom) returns (MsgAllowAskDenomResponse);
```
ryanchristo, line 30: I think the original intention here was to add an ask denom, i.e. "to allow it". We should probably have two separate methods, one for adding and one for removing and we should only allow one at a time. I think we should avoid batch updates to governance params and each addition/removal warrants its own governance proposal.

ryanchristo (commented):
As mentioned in #960 and #962, it looks like we may not update to Cosmos SDK v0.46 for Regen Ledger v4.0. We can still add the proto updates and the message implementations but we will need to find a workaround for allowed ask denoms until the v0.46 update.

In line with my comments in those two pull requests, I think we should avoid batch updates for governance methods as each addition/removal should warrant its own proposal.

proto/regen/ecocredit/marketplace/v1/tx.proto
technicallyty: yeah i think it makes sense to remove batching. governance can do batch updates if needed anyway, as the new v0.46 gov updates allows for multiple messages to execute from a single proposal

proto/regen/ecocredit/marketplace/v1/tx.proto
```diff
@@ -195,21 +195,32 @@ message MsgBuyResponse {
...
 
+
 // MsgAllowAskDenom is the Msg/AllowAskDenom request type.
 message MsgAllowAskDenom {
-  // root_address is the address of the governance account which can authorize
-  // ask denoms
+
+  // root_address is the address of the signer.
+  // This MUST equal the address of the gov module for the tx to succeed.
   string root_address = 1;
```
robert-zaremba, line 204: root sounds ambiguous. Why not `sender` or `signer`?

proto/regen/ecocredit/marketplace/v1/tx.proto
robert-zaremba: TBH, I like the idea of packing this into one proposal (as it's done now). Imagine there are 10 denoms, and you want to change them to 10 other denoms. That would require 20 proposals. 
At the same time, if there was no clarity during the signaling, one can still do small changes using a proposal with only one edit.

proto/regen/ecocredit/marketplace/v1/tx.proto
```diff
@@ -195,21 +195,32 @@ message MsgBuyResponse {
...
 
-  // denom is the denom to allow (ex. ibc/GLKHDSG423SGS)
-  string denom = 2;
+  // add_denoms specifies denoms to add to the allowed denom list
```
robert-zaremba, line 206:
```suggestion
  // add_denoms specifies denoms to add to the allowed denom list.
  // It's OK to include here a denom which is already allowed. 
  // Returns an error if a union of `add_denom` and `remove_denom` is not empty.
```

proto/regen/ecocredit/marketplace/v1/tx.proto
```diff
@@ -195,21 +195,32 @@ message MsgBuyResponse {
...
+  repeated DenomInfo add_denoms = 2;
+
+  // remove_denoms specifies denoms to remove from the allowed denom list
```
robert-zaremba, line 209:
```suggestion
  // remove_denoms specifies denoms to remove from the allowed denom list.
  // Returns an error if a union of `add_denom` and `remove_denom` is not empty.
```

proto/regen/ecocredit/marketplace/v1/tx.proto
```diff
@@ -195,21 +195,32 @@ message MsgBuyResponse {
...
+  // DenomInfo contains the information related to a bank denom
+  message DenomInfo {
+    // denom is the denom to allow (ex. ibc/GLKHDSG423SGS)
+    string denom = 1;
 
-  // exponent is the exponent that relates the denom to the display_denom and is
-  // informational
-  uint32 exponent = 4;
+    // display_denom is the denom to display to the user and is informational
+    string display_denom = 2;
```
robert-zaremba, line 218: not sure why this is needed? I think we should use the default display from the bank module.

proto/regen/ecocredit/marketplace/v1/tx.proto
```diff
@@ -195,21 +195,32 @@ message MsgBuyResponse {
...
+
+    // exponent is the exponent that relates the denom to the display_denom and
+    // is informational
+    uint32 exponent = 3;
```
robert-zaremba, line 222: again, why denormalizing and not using the information from bank?

robert-zaremba (commented):
I would prefer that we use information from the bank instead of denormalizing it in the `allow denom list`

ryanchristo:
Closing per #1342 and #1346

Issue #976, opened 2022-04-04
update ecocredit genesis to use v1 types


aleem1314:
## Summary

the ecocredit module is currently using v1alpha2 state for genesis import and export.

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #977, opened 2022-04-04
refactor(ecocredit): update genesis import & export to use v1 api

aleem1314:
## Description

Closes: #976 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/module/module_test.go
```diff
@@ -13,6 +14,8 @@ func TestHybridORMLegacyGenesis(t *testing.T) {
 	cdc := codec.NewProtoCodec(codectypes.NewInterfaceRegistry())
 	bz := m.DefaultGenesis(cdc)
 	require.NotNil(t, bz)
+	fmt.Println(string(bz))
+	require.True(t, false)
```
aleem1314, line 18:
```suggestion
```

x/ecocredit/genesis.go
```diff
@@ -1,172 +0,0 @@
-package ecocredit
-
-import (
-	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
-	"github.com/regen-network/regen-ledger/types/math"
-)
-
-// Validate performs basic validation for each credit-batch,
-// it returns an error if credit-batch tradable or retired supply
-// does not match the sum of all tradable or retired balances
```
aleem1314, line 10: Do we need batch balance = batch supply validation? To validate the genesis state we need a proto message for each type.

```
message Classes {
  repeated ClassInfo classes = 1;
}
```

ryanchristo: Hmmm... I'm not sure how to best handle this. cc @aaronc @technicallyty @robert-zaremba

aaronc:
If we need to check invariants in validation, we should just import to an in memory orm db and run the invariants. Does that make sense?

x/ecocredit/genesis.go
technicallyty: going off what aaron mentioned below, i think we should be able to put the raw json inside an in-mem ORM db, then use iterate over the batch balance/supply tables and check invariants that way

x/ecocredit/core/genesis.go
```diff
@@ -0,0 +1,267 @@
...
+		if err != nil {
+			return err
+		}
+
+		if _, ok := abbrevToPrecision[class.CreditType]; !ok {
+			return sdkerrors.ErrNotFound.Wrapf("credit type not exist for %s abbreviation", class.CreditType)
+		}
+	}
+
+	classIds := make(map[uint64]uint64) // map of projectID to classID
```
technicallyty, line 79: i like the convention you used above (with abbrevToPrecision) of naming map type variables <key>To<Value>. could we also do that here?

x/ecocredit/core/genesis.go
```diff
@@ -0,0 +1,267 @@
...
+			return err
+		}
+
+		if _, exists := classIds[val.Id]; exists {
+			continue
+		}
+		classIds[val.Id] = val.ClassId
+	}
+
+	decimalPlaces := make(map[uint64]uint32) // map of batchID to precision
```
technicallyty, line 98: same here ☝🏻

x/ecocredit/core/genesis.go
```diff
@@ -0,0 +1,267 @@
...
+	bItr, err := ss.BatchInfoTable().List(ormCtx, api.BatchInfoPrimaryKey{})
+	if err != nil {
+		return err
+	}
+	defer bItr.Close()
+
+	for bItr.Next() {
```
technicallyty, line 105: comments for what each of these loop iterations are doing would be nice 😄

x/ecocredit/core/genesis.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	if err := validateSupply(calSupplies, supplies); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func calculateSupply(ctx context.Context, decimalPlaces map[uint64]uint32, ss api.StateStore, calSupply map[uint64]math.Dec) error {
```
technicallyty, line 183: similar to above, it would be nice to have the variable form KeyToValue for maps, its not immediately clear to devs what the key is supposed to be in this function

x/ecocredit/core/genesis.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+	}
+
+	return nil
+}
+
+func validateSupply(calSupply, supply map[uint64]math.Dec) error {
+	for denom, cs := range calSupply {
+		if s, ok := supply[denom]; ok {
+			if s.Cmp(cs) != 0 {
```
technicallyty, line 239:
```suggestion
			if s.Cmp(cs) != math.EqualTo {
```

can use this const to make it a lil more readable

technicallyty (commented):
this is great work 👏🏻 just a few comments/suggestions

x/ecocredit/core/genesis_test.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/regen-network/regen-ledger/x/ecocredit/core"
+)
+
+func TestValidateGenesis(t *testing.T) {
+	x := `{
+		"regen.ecocredit.v1.BatchBalance":[
+			{"address":"gydQIvR2RUi0N1RJnmgOLVSkcd4=","batch_id":"1","tradable":"90.003","retired":"9.997","escrowed":""}
+		],
```
ryanchristo, line 16: Might be nice to format some of these objects to help with readability.

ryanchristo:
```suggestion
		"regen.ecocredit.v1.BatchBalance": [
			{
				"address": "gydQIvR2RUi0N1RJnmgOLVSkcd4=",
				"batch_id": "1",
				"tradable": "90.003",
				"retired": "9.997",
				"escrowed": ""
			}
		],
```

x/ecocredit/core/genesis_test.go
```diff
@@ -0,0 +1,234 @@
...
+			},
+			{
+				Abbreviation: "BIO",
+				Name:         "biodiversity",
+				Unit:         "ton",
+				Precision:    6,
+			},
+		},
+	}
+	core.DefaultParams()
```
ryanchristo, line 56:
```suggestion
```

x/ecocredit/core/genesis_test.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/regen-network/regen-ledger/x/ecocredit/core"
+)
+
+func TestValidateGenesis(t *testing.T) {
+	x := `{
+		"regen.ecocredit.v1.BatchBalance":[
+			{"address":"gydQIvR2RUi0N1RJnmgOLVSkcd4=","batch_id":"1","tradable":"90.003","retired":"9.997","escrowed":""}
+		],
+		"regen.ecocredit.v1.BatchInfo":[
+			{"issuer":"WCBEyNFP/N5RoS4h43AqkjC6zA8=","project_id":"1","batch_denom":"BIO01-00000000-00000000-001","metadata":"batch metadata","start_date":null,"end_date":null,"issuance_date":"2022-04-08T10:40:10.774108141Z"},
```
ryanchristo, line 18: Any reason for using `null` for some of these values and not for others? Might be nice to only include the values necessary for validation or to include values for all properties. For example, `start_date` is `null` but `issuance_date` is not and neither are necessary for `GenesisValidate` to pass.

x/ecocredit/core/genesis_test.go
```diff
@@ -0,0 +1,234 @@
...
+		{
+			"valid: no credit batches",
+			func() json.RawMessage {
+				return json.RawMessage(`{
+					"regen.ecocredit.v1.ClassInfo": [{"id":1,"name":"1","admin":"0lxfU2Ca/sqly8hyRhD8/lNBrvM=","metadata":"meta-data","credit_type":"C"}]
+				}`)
+			},
+			defaultParams,
+			false,
+			"",
```
ryanchristo, line 80: Nice to see this here. Hopefully we will have credit batches but just in case we don't.

ryanchristo (commented):
This is looking great. Very nice work. A few questions and comments.

x/ecocredit/core/genesis_test.go
aleem1314: > start_date is null but issuance_date is not and neither are necessary for GenesisValidate to pass.

Yes. Those fields are not required for genesis validation. I'll remove them.

I'm also thinking about adding validation for each message type to check all fields are properly set.

x/ecocredit/core/genesis.go
```diff
@@ -0,0 +1,267 @@
...
+		if balance.Tradable != "" {
+			tBalance, err = math.NewNonNegativeFixedDecFromString(balance.Tradable, decimalPlaces[balance.BatchId])
+			if err != nil {
+				return err
+			}
+		}
+
+		if balance.Retired != "" {
+			rBalance, err = math.NewNonNegativeFixedDecFromString(balance.Retired, decimalPlaces[balance.BatchId])
+			if err != nil {
```
technicallyty, line 211: we should also add one for `Escrowed` amounts. these are differenciated on the batch balance level, but are part of the the `TradableSupply` amount. so:

supply tradable + supply retired = balances tradable + balances retired + balances escrowed

x/ecocredit/core/genesis_test.go
```diff
@@ -0,0 +1,414 @@
...
+)
+
+func TestValidateGenesis(t *testing.T) {
+	x := `{
+		"regen.ecocredit.v1.BatchBalance":[
+			{
+				"address":"gydQIvR2RUi0N1RJnmgOLVSkcd4=",
+				"batch_id":"1",
+				"tradable":"90.003",
+				"retired":"9.997"
```
technicallyty, line 19: we could add escrowed balance here as a test case

technicallyty (commented):
didnt catch these in my last review, but should be good to go after this

x/ecocredit/core/genesis_test.go
```diff
@@ -0,0 +1,414 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			err := core.ValidateGenesis(tc.gensisState(), tc.params)
+			if tc.expectErr {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), tc.errorMsg)
```
ryanchristo, line 408: Looks like you are using the full error for all test cases. You should be able to use `EqualError` here:

```suggestion
				require.EqualError(t, err, tc.errorMsg)
```

ryanchristo (commented):
Looks good to me. One minor suggestion and the requested changes to include escrowed.

x/ecocredit/core/genesis_test.go
ryanchristo: That sounds like the right direction to me.

x/ecocredit/core/genesis_test.go
aleem1314: Actual error is even longer `invalid creator address: decoding bech32 failed: invalid separator index -1: invalid address ...`

x/ecocredit/core/genesis.go
```diff
@@ -0,0 +1,415 @@
...
+	"google.golang.org/protobuf/reflect/protoreflect"
+
+	api "github.com/regen-network/regen-ledger/api/regen/ecocredit/v1"
+	"github.com/regen-network/regen-ledger/types/math"
+	"github.com/regen-network/regen-ledger/types/ormutil"
+	"github.com/regen-network/regen-ledger/x/ecocredit"
+)
+
+// ValidateGenesis performs basic validation for the following:
+// - params are valid param types with valid properties
```
ryanchristo, line 25:
```suggestion
// - params are valid param types with valid properties
// - proto messages are valid proto messages
```

x/ecocredit/core/genesis.go
ryanchristo: It looks like this might have been overlooked in most recent changes.

- `decimalPlaces` should be updated to `batchIdToPrecision`
- `calSupply` should be updated to `batchIdToCalSupply`

x/ecocredit/core/genesis_test.go
ryanchristo: Looks like this might have been overlooked. We should include a test case with a positive escrowed amount.

ryanchristo: It also looks like you are missing a `BatchBalance` validation.

ryanchristo: Also `CreditType` and `BatchSupply` validation.

ryanchristo (commented):
Updates look great. The proto message validation is a nice add. A few more comments.

x/ecocredit/core/genesis_test.go
aleem1314: > Also `CreditType` and `BatchSupply` validation.
 
Added validation for `CreditType`. Batch balance & supply validation is already covered by `ValidateGenesis`.

technicallyty (approved):
lgtm

Issue #896, opened 2022-03-15
Add data module acceptance tests


ryanchristo:
## Summary

Add acceptance tests for the data module using gherkin syntax.

See https://github.com/regen-network/gocuke for examples.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #817, opened 2022-02-28
Use go generate instead of make for mocks


robert-zaremba:
## Summary

`go generate` is the standard way to handle Go code dependencies. It integrate nicely with go modules and allow us to run [mockgen](https://github.com/golang/mock/) with module controlled mockgen version. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #994, opened 2022-04-07
test(x/data): add acceptance tests

ryanchristo:
## Description

Closes: #896

This pull request adds feature files for data module acceptance tests and implements those tests. Also fixes an issue with attestation timestamp previously using the anchor timestamp, which will not be the same if the data has already been anchored. Also some minor error improvements.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/data/server/features/anchor.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Anchor
+
+  Scenario: raw data is anchored
```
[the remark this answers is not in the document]
ryanchristo: Should we write separate features for message validation or add those here? The content hash is not validated in the rpc method implementation.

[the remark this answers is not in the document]
ryanchristo: Also, how would we represent a content hash as an example input?

Makefile
```diff
@@ -410,6 +410,9 @@ include sims.mk
 regen-mocks:
 	go install github.com/golang/mock/mockgen@latest
 
+	mkdir -p x/ecocredit/server/core/mocks
+	mockgen -source=x/data/expected_keepers.go -package mocks -destination x/data/mocks/expected_keepers.go
```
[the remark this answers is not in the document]
ryanchristo: Should we update all of these in a separate pull request #817 or make those changes here?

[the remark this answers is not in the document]
ryanchristo: Opened #998.

x/data/features/msg_anchor.feature
```diff
@@ -0,0 +1,21 @@
+Feature: MsgAnchor
+
+  Scenario Outline: sender must be valid
```
aaronc, line 3: we probably don't need this for every message, but it makes sense and is clear

aaronc, line 3: we probably don't need this for every message, but it makes sense and is clear

x/data/features/msg_anchor.feature
```diff
@@ -0,0 +1,21 @@
...
+    When the message is validated
+    Then an error of "<error>"
+
+    Examples:
+    | sender                                        | error                                                                   |
+    |                                               | empty address string is not allowed: invalid address                    |
+    | foo                                           | decoding bech32 failed: invalid bech32 string length 3: invalid address |
+    | cosmos1depk54cuajgkzea6zpgkq36tnjwdzv4afc3d27 |                                                                         |
+
+  Scenario: hash cannot be empty
```
aaronc, line 15: you could include this case in the scenario outline above by adding a content has column

aaronc, line 15: you could include this case in the scenario outline above by adding a content has column

x/data/features/msg_attest.feature
```diff
@@ -0,0 +1,21 @@
+Feature: MsgAttest
+
+  Scenario Outline: attestor must be valid
+    Given an attestor of "<attestor>"
+    And a valid content hash
+    When the message is validated
+    Then an error of "<error>"
+
+    Examples:
+    | attestor                                      | error                                                                   |
```
aaronc, line 10: one way we could do this to cover all cases msg validation would be something like this
```suggestion
    | message                                      | error                                                                   |
    | {"attestor":"foo"}.                      | decoding bech32 failed                                |
```

where message is proto json

aaronc, line 10: one way we could do this to cover all cases msg validation would be something like this
```suggestion
    | message                           | error                                              |
    | {"attestor":"foo"}.           | decoding bech32 failed            |
```

where message is proto json

x/data/msg_define_resolver_test.go
```diff
@@ -0,0 +1,51 @@
...
+	gocuke.NewRunner(t, &msgDefineResolverSuite{}).Path("./features/msg_define_resolver.feature").Run()
+}
+
+func (s *msgDefineResolverSuite) Before(t gocuke.TestingT) {
+	s.t = t
+	s.msg = &MsgDefineResolver{}
+}
+
+func (s *msgDefineResolverSuite) AValidManager() {
+	s.msg.Manager = "cosmos1depk54cuajgkzea6zpgkq36tnjwdzv4afc3d27"
```
aaronc, line 26: this can go in the .feature file I think

aaronc, line 26: this can go in the .feature file I think

x/data/server/features/anchor.feature
```diff
@@ -0,0 +1,18 @@
+Feature: Anchor
+
+  Scenario: data is anchored when the data has not been anchored
+    Given a valid content hash
```
aaronc, line 4: can we have an example, maybe in proto json in a gherkin doc string

aaronc, line 4: can we have an example, maybe in proto json in a gherkin doc string

x/data/server/features/anchor.feature
```diff
@@ -0,0 +1,18 @@
...
+    When alice attempts to anchor the data at block time "2020-01-01"
+    Then an error of ""
```
aaronc, line 6:
```suggestion
```

we can just skip checking for an error because we don't expect one

aaronc, line 6:
```suggestion
```

we can just skip checking for an error because we don't expect one

x/data/server/features/anchor.feature
```diff
@@ -0,0 +1,18 @@
...
+    And the data id entry exists
+    And the data anchor entry exists and the timestamp is equal to "2020-01-01"
+
+  Scenario: data anchor entry is not updated when the data has already been anchored
+    Given a valid content hash
+    And alice has anchored the data at block time "2020-01-01"
+    When alice attempts to anchor the data at block time "2020-01-02"
+    Then an error of ""
+    And the data id entry exists
```
aaronc, line 15:
```suggestion
```

not needed

x/data/server/features/attest.feature
```diff
@@ -0,0 +1,29 @@
+Feature: Attest
+
+  Scenario: data is attested to when the data has been anchored
+    Given a valid content hash
+    And alice has anchored the data at block time "2020-01-01"
+    When alice attempts to attest to the data at block time "2020-01-02"
+    Then an error of ""
+    And the data id entry exists
+    And the data anchor entry exists and the timestamp is equal to "2020-01-01"
+    And the data attestor entry exists and the timestamp is equal to "2020-01-02"
```
aaronc, line 10: we should be testing usually one thing per scenario, unless it's a "journey" scenario. this seems like too many ands

x/data/server/features/register_resolver.feature
```diff
@@ -0,0 +1,30 @@
+Feature: Register Resolver
+
+  Scenario: data is registered to the resolver when the data has been anchored
+    Given a valid content hash
```
aaronc, line 4: this can be in the `Background` section and let's use a real example

x/data/server/features/register_resolver.feature
```diff
@@ -0,0 +1,30 @@
...
+    Then an error of ""
+    And the data resolver entry exists
+
+  Scenario: data is anchored and registered to the resolver when the data has not been anchored
+    Given a valid content hash
+    And alice has defined a resolver
+    When alice attempts to register the data to the resolver
+    Then an error of ""
+    And the data anchor entry exists
+    And the data resolver entry exists
```
aaronc, line 17: any way we can test one thing per scenario and have fewer ands?

aaronc (commented):
This generally looks like it's going in the right direction. Left a few comments on how we can improve.

These are principles that the BDD book mentions that I think are good to keep in mind:
image

Some of these scenarios, in addition to using real data, could be more "essential" and "focused"

This is more .feature files then I think we would normally try to review in one go, but I do so far feel better reading these when doing the review than normal test files - it's easier to evaluate the intention. I also like how brief and straightforward the test implementation actually is with this approach

x/data/features/msg_attest.feature
ryanchristo: Any attempt to use json as the value produces the following error:

```
internal error: no matches found when matching ^a\s+content\s+hash\s+of\s+"([^"]*)"$
```

https://github.com/regen-network/gocuke/blob/main/run_step.go#L39-L42

aaronc: Ah... well you just have to manually register the step with a regex then

x/data/features/msg_anchor.feature
ryanchristo: duplicate

x/data/features/msg_anchor.feature
ryanchristo: duplicate

x/data/features/msg_attest.feature
ryanchristo: duplicate

x/data/msg_define_resolver_test.go
ryanchristo: Moved to the feature file and within a message defined with json.

x/data/server/features/anchor.feature
ryanchristo: Added an example using a gherkin doc string and added to background so that it could be reused.

ryanchristo: duplicate

x/data/server/features/anchor.feature
ryanchristo: Removed the error check whenever we verify the entry exists.

ryanchristo: duplicate

x/data/server/features/anchor.feature
ryanchristo: Removed the data id entry exists and error check.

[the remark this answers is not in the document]
ryanchristo: duplicate

x/data/server/features/attest.feature
ryanchristo: Updated to test one thing per scenario (i.e. separate tests for anchor entry and attestor entry)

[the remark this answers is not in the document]
ryanchristo: duplicate

x/data/server/features/register_resolver.feature
ryanchristo: Moved to background and updated to use real example with doc string.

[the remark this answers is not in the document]
ryanchristo: duplicate

x/data/server/features/register_resolver.feature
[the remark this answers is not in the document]
ryanchristo: Updated to test one thing per scenario (i.e. separate tests for anchor entry and resolver entry)

ryanchristo: duplicate

x/data/features/msg_anchor.feature
```diff
@@ -0,0 +1,47 @@
+Feature: MsgAnchor
+
+  Scenario: an error is returned if sender is empty
+    Given a message of
+    """
+    {}
+    """
+    When the message is validated
+    Then an error of "empty address string is not allowed: invalid address"
```
aaronc, line 9:
```suggestion
    Then expect the error "empty address string is not allowed: invalid address"
```

x/data/features/msg_anchor.feature
```diff
@@ -0,0 +1,47 @@
+Feature: MsgAnchor
+
+  Scenario: an error is returned if sender is empty
+    Given a message of
```
aaronc, line 4:
```suggestion
    Given the message
```

aaronc (commented):
I'm thinking it makes sense to use the Rule keyword to explain what is being tested better. What do you think @ryanchristo?

Generally looks like the right direction, but I think we should spend a bit more time refining this so it serves as a good example for other cases

x/data/features/msg_anchor.feature
```diff
@@ -0,0 +1,47 @@
...
+    Then an error of "hash cannot be empty: invalid request"
+
+  Scenario: no error is returned if sender and hash are valid
+    Given a message of
+    """
+    {
+      "sender": "cosmos1depk54cuajgkzea6zpgkq36tnjwdzv4afc3d27",
+      "hash": {
+        "raw": {
+          "hash": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
aaronc, line 38: we definitely want base64 or hex encoding here. let's use proto json

x/data/msg_anchor_test.go
```diff
@@ -0,0 +1,43 @@
...
+	runner.Run()
+}
+
+func (s *msgAnchorSuite) Before(t gocuke.TestingT) {
+	s.t = t
+	s.msg = &MsgAnchor{}
+}
+
+func (s *msgAnchorSuite) AMessageOf(a gocuke.DocString) {
+	err := json.Unmarshal([]byte(a.Content), &s.msg)
```
aaronc, line 29: we should be using https://pkg.go.dev/github.com/gogo/protobuf/jsonpb

x/data/features/msg_anchor.feature
ryanchristo: updated

x/data/features/msg_anchor.feature
ryanchristo: updated

x/data/features/msg_anchor.feature
ryanchristo: now using protopb and base64

ryanchristo: now using jsonpb and base64

x/data/msg_anchor_test.go
ryanchristo: now using jsonpb

ryanchristo:
> I'm thinking it makes sense to use the Rule keyword to explain what is being tested better.

I'm not sure how granular to make the rules, whether each message feature has one rule, e.g. "an error is returned if message is invalid", or each field has a rule, e.g. "an error is returned if sender is invalid", "an error is returned if content hash is invalid", and "no error is returned if sender and content hash are valid".

The more granular the rules the more redundant they become with the scenarios that follow. The scenarios are pretty descriptive as is but maybe those should be less descriptive and rules should be added. How do you see rules being used? Can you provide an example, maybe one for a message and one for server method?

x/data/features/msg_define_resolver.feature
```diff
@@ -0,0 +1,23 @@
+Feature: MsgDefineResolver
```
robert-zaremba, line 1: can rename the directory to `acceptance-tests`?

x/data/msg_define_resolver_test.go
```diff
@@ -0,0 +1,51 @@
...
+	s.err = s.msg.ValidateBasic()
+}
+
+func (s *msgDefineResolverSuite) AnErrorOf(a string) {
+	if a == "" {
+		require.NoError(s.t, s.err)
+	} else {
+		require.EqualError(s.t, s.err, a)
+	}
+}
```
robert-zaremba, line 43: I wonder if we should think about some base structure for other acceptance tests resolver.

robert-zaremba (approved):
overall looks good, however when expanding the implementation we should think how we will reuse the code.

x/data/features/msg_define_resolver.feature
ryanchristo: Using features for the directory name is actually a standard practice when writing features. Check out the examples in gocuke or the cucumber documentation if you're interested in learning more:

- https://github.com/regen-network/gocuke
- https://cucumber.io/docs/gherkin/reference/

x/data/msg_define_resolver_test.go
ryanchristo: Might be interesting to further explore.

x/data/features/msg_define_resolver.feature
robert-zaremba: I wouldn't call it it a _standard practice_. The file is _feature_ file of an acceptance test. Usually we define test in spec or test directories.

ryanchristo: I think it's safe to call it a standard practice and have not seen an example that does not use the `features` as the name of the directory. Seems like a silly thing to debate over.

ryanchristo: Seems like a silly thing to debate over. I think it's safe to call it a standard practice and have not seen an example that does not use `features` as the name of the directory.

ryanchristo: I would recommend checking out some of the resources. I think standard practice is pretty accurate. We define golang tests in a test directory but features are typically stored in a features directory.

x/data/features/msg_anchor.feature
```diff
@@ -0,0 +1,54 @@
+Feature: MsgAnchor
+
+  Rule: only a valid sender is accepted
```
aaronc, line 3: If rule is to be useful it should describe what valid means, sender must be a non-empty bech32 address

x/data/features/msg_anchor.feature
```diff
@@ -0,0 +1,54 @@
...
+
+    Scenario: an error is returned if sender is empty
```
aaronc, line 5: As is this is more informative than the rule above

aaronc (commented):
Whatever construct we use, rule or scenario part of the idea is that this also serves as human readable documentation. So in documentation we define what is valid, we can't assume anyone knows that

aaronc (commented):
Whatever construct we use, rule or scenario part of the idea is that this also serves as human readable documentation. So in documentation we define what is valid, we can't assume anyone knows that

x/data/features/msg_anchor.feature
ryanchristo: I can update to be more specific. The BDD book uses "only valid address is accepted" as an example.

aaronc: Interesting

x/data/features/msg_anchor.feature
ryanchristo: If the rule is "an error is returned if sender is empty" then the scenarios would be "sender is empty", "sender is not empty", and "sender is not empty" could still return an error if not a valid address.

ryanchristo: This would also cause a lot of duplication if we were testing the valid case for each rule. "sender is not empty" (or sender is a valid address) would then be duplicated with a different scenario name for each rule.

aaronc: So maybe we shouldn't use Rule?

ryanchristo: Hmmm... I'm not sure at this point. I think Rule could be useful in some cases. It might make sense in the context of the server methods but not necessarily for message and content hash validation. For example, when we are testing whether or not the data has been anchored when we are actually testing the `Attest` or `RegisterResolver` methods.

ryanchristo: I think Rule could be useful in some cases. It might make sense in the context of the server methods but not necessarily for message and content hash validation. For example, when we are testing whether or not the data has been anchored when we are actually testing the `Attest` or `RegisterResolver` methods ([see here](https://github.com/regen-network/regen-ledger/pull/994/files#diff-ee6cd2c2c155f81315035619ed96293656b72688fc7597e3399ca43a8cfd8df5R15-R41)).

x/data/features/msg_anchor.feature
```diff
@@ -0,0 +1,47 @@
...
+    Given the message
+    """
+    {
+      "sender": "cosmos1depk54cuajgkzea6zpgkq36tnjwdzv4afc3d27"
+    }
+    """
+    When the message is validated
+    Then expect the error "content hash cannot be empty: invalid request"
+
+  Scenario: no error is returned if all message fields pass validation
```
aaronc, line 33:
```suggestion
  Scenario: a valid message
```

Also let's make this the first scenario in the file

x/data/features/msg_anchor.feature
```diff
@@ -0,0 +1,47 @@
...
+    Given the message
+    """
+    {
+      "sender": "foo"
+    }
+    """
+    When the message is validated
+    Then expect the error "decoding bech32 failed: invalid bech32 string length 3: invalid address"
+
+  Scenario: an error is returned if content hash is empty
```
aaronc, line 23: This points to an issue with how we do validation that @robert-zaremba has raised, specifically that error reporting depends on ordering. What if we validate content hash first? Then these tests will change which just feels wrong. We should actually return all errors and bundle them together with something like https://github.com/hashicorp/go-multierror. Not for this PR obviously but something we should track. This would ideally be integrated with the SDK's error package.

x/data/features/msg_define_resolver.feature
aaronc: I prefer features/ personally

x/data/server/features/attest.feature
```diff
@@ -0,0 +1,40 @@
+Feature: Attest
+
+  Background: the message has been validated
```
aaronc, line 3:
```suggestion
  Background:
```
This doesn't seem needed

aaronc (approved):
A few tweaks I would suggest but otherwise looks good. Good work @ryanchristo 👏

x/data/features/msg_anchor.feature
ryanchristo: Looks like this was started in #941.

PR #998, opened 2022-04-08
build: use go generate for mocks

ryanchristo:
## Description

Closes: #817

Use `go generate` for generating mock code and add data module. Also removes unused mocks.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


technicallyty (approved):
tACK

aaronc (commented):
So how is this used?

ryanchristo:
> So how is this used? Do we need some docs?

You just need to run `go generate` instead of `make regen-mocks`.

We didn't have any prior documentation to update for `make regen-mocks`. Where would be the best place to add some? It looks like contributing could use an overhaul.

ryanchristo:
Opened #1003

generate.go
```diff
@@ -0,0 +1,8 @@
+package regen_ledger
+
+//go:generate mockgen -source=x/data/expected_keepers.go -package mocks -destination x/data/mocks/expected_keepers.go
+//go:generate mockgen -source=x/ecocredit/server/core/keeper.go -package mocks -destination x/ecocredit/server/core/mocks/keeper.go
+//go:generate mockgen -source=x/ecocredit/expected_keepers.go -package mocks -destination x/ecocredit/mocks/expected_keepers.go
+//go:generate mockgen -source=x/ecocredit/expected_keepers.go -package mocks -destination x/ecocredit/mocks/expected_keepers.go
+//go:generate mockgen -source=x/ecocredit/server/basket/keeper.go -package mocks -destination x/ecocredit/server/basket/mocks/keeper.go
+//go:generate mockgen -source=x/ecocredit/expected_keepers.go -package mocks -destination x/ecocredit/mocks/expected_keepers.go
```
robert-zaremba, line 8: normally we put the directive next to the related types. So the line above should be placed in x/ecocredit/expected_keepers.go file.

robert-zaremba (commented):
let's move the directives next to the related types and add `go generate ./...` to Makefile.

robert-zaremba:
@aaronc , the advantage over make is:
1. go tools will be part of the go module (as a versioned dependency) - so we can control which version we are using
2. we put the related directives next to the related types (interfaces)

ryanchristo:
> let's move the directives next to the related types and add go generate ./... to Makefile.

Where should `go generate` be added to the makefile. I thought the idea was to avoid using `make`? The issue is titled, "Use go generate instead of make for mocks". Maybe I misunderstood?

generate.go
ryanchristo: Can do. I found this as an example but looks like best practice is on a per file basis.

ryanchristo:
So `go generate ./...` does not work from the root directory when placed within each file. You have to run:

```
go generate x/data/expected_keepers.go
go generate x/ecocredit/expected_keepers.go
go generate x/ecocredit/server/basket/keeper.go
go generate x/ecocredit/server/core/keeper.go
```

Or you have to navigate to the each directory and run `go generate ./...`

generate.go
ryanchristo: I see why this was an example. You have to run `go generate ./...` from within the directory and this was being used as a convenient way to generate all mocks with one command.

ryanchristo:
Would be happy to make changes but not sure how to best proceed? @robert-zaremba

technicallyty:
> > let's move the directives next to the related types and add go generate ./... to Makefile.
> 
> Where should `go generate` be added to the Makefile? I thought the idea was to avoid using `make`? The issue is titled, "Use go generate instead of make for mocks". Maybe I misunderstood?

the issue was probably `go generate` vs `mockgen`, so i think adding the `go generate` lines to the makefile is good 👍🏻

robert-zaremba:
> the issue was probably go generate vs mockgen

The issue is with multiple modules. `./...` will only take current module into the account. We have the same problem with tests. In fact it's not guaranteed that every tools will work with a root module when there are multiple submodules.

> Where should go generate be added to the Makefile? 
> Would be happy to make changes but not sure how to best proceed?

Simple `generate:` job would be perfect. I think the following will work:

```
generate: 
    find -name go.mod -execdir go generate ./... \;
```

Makefile
```diff
@@ -404,17 +404,11 @@ localnet-stop:
...
+###############################################################################
+###                                Generate                                 ###
+###############################################################################
 
-	mkdir -p x/ecocredit/server/basket/mocks
-	mockgen -source=x/ecocredit/server/basket/keeper.go -package mocks -destination x/ecocredit/server/basket/mocks/keeper.go
-	mockgen -source=x/ecocredit/expected_keepers.go -package mocks -destination x/ecocredit/mocks/expected_keepers.go
-.PHONY: regen-mocks
+generate:
+	find . -name 'go.mod' -type f -execdir sh -c 'go generate ./...' \;
```
ryanchristo, line 414: This will only work for go generate commands within the same directory as `go.mod`. We had mockgen set up for `server/core` and `server/basket` but those were empty and unused.

robert-zaremba: I don't think you need to wrap the go generate command in `sh`

robert-zaremba (approved):

Issue #904, opened 2022-03-17
Add support for pending batch issuance


ryanchristo:
## Summary

Issuing credit batches is currently permissioned, i.e. only approved issuers of a credit class can issue credits from that credit class. In other issuance designs, a user can submit a request for issuing a credit batch that is then represented as a pending batch issuance. A pending batch issuance can then be approved, which triggers the issuance of the credit batch.

## Problem Definition

[ work in progress ]

## Proposal

[ work in progress ]

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ryanchristo:
Converting this issue to an epic as it should be properly scoped and require an RFC before implementation.

PR #1004, opened 2022-04-09
feat: Add support for pending batch issuance

clevinson:
## Description

ref: #904 

This PR demonstrates a proposed API for supporting "pending batch issuances". 

- [x] tx.proto updates
- [ ] query.proto updates
- [ ] state.proto updates

An example use case we're trying to cover is as follows:
1. Alice wants to bridge 500 credits from Verra to Regen Ledger, for a vintage, Kasigao 2015, that does not currently have a corresponding batch on Regen Ledger
2. Alice submits a `MsgMintPendingBatchCredits` call, with both `batch_info` and `project_info` filled out to provide sufficient on-chain metadata for these credits. She sets the `issuer` field in her request to Bob's regen address, as Bob is the bridge operator who will ultimately approve her batch. This request does not include any actual "issuances" yet, as she does not have a retirement serial number yet on Verra to submit as the `TxOrigin` for those issuances.
3. Alice then receives a `batch_denom` in the corresponding `MsgMintPendingBatchCreditsResponse`
4. Alice retires the credits on Verra, adding the following information to her retirement info:
    - Beneficiary: [Alice's regen address]
    - Retirement Details: [the `batch_denom` she received after submitting her pending batch request]
5. Alice submits another `MsgMintPendingBatchCredits` call. This one with `batch_denom` filled out, instead of `batch_info`, as the batch already exists in a pending state on-chain.  In this call, she sets in the `issuance` field the amount of credits that she wishes to receive, and designates herself as the recipient. Since she only listed herself as the Beneficiary on Verra's retirement form, she cannot add multiple issuances.  In the `tx_origin` field, she adds the Verra retirement serial number from her Verra retirement.
6. Bob can query regen ledger on-chain for pending batches & credit issuances, and see Alice's pending batch & issuance. Bob can then approve the credit issuance once he verifies that the corresponding Verra retirement information corresponds correctly with the on-chain batch & issuance request data.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


clevinson:
Interested in early feedback from anyone on this. In particular @aaronc @S4mmyb @ryanchristo

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -35,11 +35,20 @@ service Msg {
...
 
-  // MsgSealBatch sets the `batch.open` attribute to false. Sealed batch
+  // SealBatch sets the `batch.open` attribute to false. Sealed batch
   // can't issue more credits. Once batch is sealed it can't be toggled any
   // more. Only batch creator can seal a batch.
   rpc SealBatch(MsgSealBatch) returns (MsgSealBatchResponse);
 
+  // MintPendingBatchCredits creates a pending batch issuance of credits for an issuer to approve
+  // within a specified amount of time.
+  rpc MintPendingBatchCredits(MsgMintPendingBatchCredits) returns (MsgMintPendingBatchCreditsResponse);
```
robert-zaremba, line 45: if this creates a new batch, then maybe we should rename to `CreatePendingBatch`?

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -35,11 +35,20 @@ service Msg {
...
+
+  // ApproveBatchCredits approves a pending batch of credits, delivering the credit issuances
+  // to the users specified in the initial MintPendingBatchCredits RPC call. Only the designated
+  // issuer for a given set of pending batch credits can approve them.
+  rpc ApproveBatchCredits(MsgApproveBatchCredits) returns (MsgApproveBatchCreditsResponse);
```
robert-zaremba, line 50: `ApprovePendingBatch`?

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -215,19 +224,111 @@ message MsgMintBatchCredits {
...
-
 // MsgSealBatchResponse is the Msg/SealBatch response type.
 message MsgSealBatchResponse {}
 
+// MsgMintPendingBatchCredits is the Msg/MintPendingBatchCredits request type.
+message MsgMintPendingBatchCredits {
+
+  // issuer is the address of the credit class issuer who is being requested
+  // to approve the batch issuance.
+  string issuer = 1;
```
robert-zaremba, line 245: shouldn't this be Alice?

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -215,19 +224,111 @@ message MsgMintBatchCredits {
...
+    google.protobuf.Timestamp start_date = 4 [ (gogoproto.stdtime) = true ];
+
+    // end_date is the end of the period during which this credit batch was
+    // quantified and verified.
+    google.protobuf.Timestamp end_date = 5 [ (gogoproto.stdtime) = true ];
+
+    // If open is true we will enable future minting.
+    // Otherwise we will seal the batch and disable the future minting.
+    bool open = 6;
+  }
```
robert-zaremba, line 290: should we add a bond in Regen to avoid spam?

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -215,19 +224,111 @@ message MsgMintBatchCredits {
...
+
+  message PendingProjectInfo {
```
robert-zaremba, line 292: To be honest, I don't find the `Info` suffix useful. Why not simply calling it `PendingProject`?

proto/regen/ecocredit/v1/tx.proto
S4mmyb: No this would be Bob - Alice is the Verra credit holder requesting for a pending batch to be created. Bob is the issuer who will review and approve the pending batch

proto/regen/ecocredit/v1/tx.proto
S4mmyb: I like robert's suggestion on the name changes

proto/regen/ecocredit/v1/tx.proto
S4mmyb: Who can lock and unlock the batch? Is it just the admin? Or issuers as well?

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -215,19 +224,111 @@ message MsgMintBatchCredits {
...
+// MsgMintPendingBatchCreditsResponse is the Msg/MintPendingBatchCredits response type.
+message MsgMintPendingBatchCreditsResponse {
+  // batch_denom is the unique denomination ID of the newly created batch.
+  string batch_denom = 1;
+}
+
+// MsgApproveBatchCredits is the Msg/ApproveBatchCredits request type.
+message MsgApproveBatchCredits {
+  string batch_denom = 1;
+  repeated uint64 pending_issuance_id = 2;
```
S4mmyb, line 326: What is the `pending_issuance_id`? Did we add in an `issuance_id` field with dynamic minting of batches?

proto/regen/ecocredit/v1/tx.proto
clevinson: It might create a new batch, but not necessarily (e.g. if a user sets `batch_denom` instead of a `PendingBatchInfo`.  Alternatively, I was thinking about splitting this out into separate RPC calls:
- one for creating a new batch (and optionally adding issuances to it)
- one for minting new credits to an existing batch

Do you think that API is cleaner?

S4mmyb:
Thanks for putting this all together @clevinson! So far I think it looks good so far and as I think about it more I think this functionality could lend itself nicely to our native credits if you think about Project Developers putting in 'issuance requests' after a monitoring round is complete.

proto/regen/ecocredit/v1/tx.proto
robert-zaremba: the bond would be returned once batch is created. I think the credit class issuers can only do that.

proto/regen/ecocredit/v1/tx.proto
robert-zaremba: yes, in this case 2 RPC is better.

proto/regen/ecocredit/v1/tx.proto
clevinson: We call it `ProjectInfo` and `BatchInfo` in state.proto, so I was just trying to stay consistent with that.

proto/regen/ecocredit/v1/tx.proto
clevinson: Yes, its only the issuer of the specific batch who can lock a batch (assuming it was open initially). It can't be unlocked once it has been locked.

clevinson: @robert-zaremba how would you propose the bond be structured? What's the reason why gas isn't sufficient for spam prevention?

proto/regen/ecocredit/v1/tx.proto
blushi: I'd tend to agree, although that's already being used for classes, batches and projects: `ClassInfo`, etc.
so we should either keep it or remove it everywhere

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -215,19 +240,145 @@ message MsgMintBatchCredits {
...
+    // alphanumeric characters. country-code is required, while sub-national-code
+    // and postal-code can be added for increasing precision.
+    string project_location = 3;
+
+    // project_id is an optional user-specified project ID which can be used
+    // instead of an auto-generated ID. If project_id is provided, it must be
+    // unique within the credit class and match the regex [A-Za-z0-9]{2,16}
+    // or else the operation will fail. If project_id is omitted an ID will
+    // automatically be generated.
+    string project_id = 4;
```
blushi, line 326: what about just using this as in `ProjectInfo`
```suggestion
    string id = 4;
```

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -215,19 +240,145 @@ message MsgMintBatchCredits {
...
+    // end_date is the end of the period during which this credit batch was
+    // quantified and verified.
+    google.protobuf.Timestamp end_date = 5 [(gogoproto.stdtime) = true];
+
+    // If open is true we will enable future minting.
+    // Otherwise we will seal the batch and disable the future minting.
+    bool open = 6;
+  }
+
+  message PendingProjectInfo {
```
blushi, line 305: maybe this should be added to state.proto instead?

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -84,6 +102,13 @@ message MsgCreateClass {
   // acceptable fees for creating a credit class can be found in the governance
   // parameters for the ecocredit module.
   cosmos.base.v1beta1.Coin fee = 5;
+
+  // allow_pending_issuances enables permissionless creation of pending
+  // batches & batch issuances. For use in credit classes that have
+  // bridged assets, and also credit classes that want to allow for
+  // users to propose credit issaunces for a specific project (e.g.
+  // from the project developer).
+  boolean allow_pending_issuances = 6;
```
blushi, line 111: how do we plan to update this field value for existing credit classes (eg C01) as part of future upgrades if we want it to be true?
do we want to do it manually or add some msg for updating this?
Allowing for updates of such a field might get tricky.

proto/regen/ecocredit/v1/tx.proto
robert-zaremba: TBH, I don't see this as a strong argument. If we agree that the style is not good, then let's use a better one than continue going into wrong direction. Especially for naming.

proto/regen/ecocredit/v1/tx.proto
robert-zaremba: Gas is cheap. You can easily create tens of thousands of pending batches.
The bond will be refunded when the batch is "finalized" / created.

I think @aaronc suggested similar thing yesterday.

proto/regen/ecocredit/v1/tx.proto
clevinson: @blushi Ya I think that makes sense, would you suggest the same for `PendingBatchInfo` ?

proto/regen/ecocredit/v1/tx.proto
clevinson: I think making it updatable actually might make sense.  @S4mmyb what do you think? Is there a use case where someone may instantiate a credit class, and at a later date decide that they want to allow pending issuances?

I think from a technical perspective it shouldn't be too tricky:
1.  If a credit class starts out not allowing pending issuances, then allowing pending issuances later shouldn't require any special logic
2. if a credit class starts out allowing pending issuances, then disallowing pending issuances later should just immediately disapprove any pending issuances
I can't think of a use case for a user doing (2) in practice, but I also don't see why the technical solution described in (2) would be bad. If an admin wants to turn off pending issuance support, then it seems reasonable to wipe out any currently pending batches/issuances.

blushi: yeah my main question was about (2), whether we want to disallow existing pending issuances or only future ones. But seems like disallowing all seems more accurate indeed.

S4mmyb: I agree with @blushi on this one. I can't really think of too many use cases for this to be honest, but I agree including that functionality isn't a bad idea.

ryanchristo:
> Putting this on hold for now. We should revisit and maybe wrap up an MVP of this once we have #1313 completed, as I assume this API will depend a bit on how we do independent projects.

We should also write an RFC for this. I converted the issue to an epic.

Issue #982, opened 2022-04-05
update ecocredit invariants to use v1 (currently v1alpha2)


aleem1314:
## Summary

the ecocredit module `tradable-supply` `retired-supply` invariants are currently using v1alpha2. We should update them to use `v1` package.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1017, opened 2022-04-13
refactor(ecocredit): migrate core invariants to use v1 types

aleem1314:
## Description

Closes: #982

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/server/basket/util.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		amount, err := math.NewDecFromString(bb.Balance)
+		if err != nil {
+			return nil, err
+		}
+
+		var batchID uint64
+		if _, ok := batchDenomToId[basket.BasketDenom]; !ok {
+			bInfo, err := k.coreStore.BatchInfoTable().GetByBatchDenom(ctx, basket.BasketDenom)
```
ryanchristo, line 37: Same here. A batch should never exist with a basket denom as the batch denom. I think you want to loop through the batches within a basket and get the credit balance of each batch to check against the total supply. We don't need to check the balance of the basket denom against anything in core but we do need to check the balance of the credits within baskets against the total supply.

x/ecocredit/server/basket/util.go
```diff
@@ -0,0 +1,59 @@
...
+
+	batchDenomToId := make(map[string]uint64)     // map of a batch denom to batch id
+	batchIdToBalance := make(map[uint64]math.Dec) // map of a basket batch_id to balance
+	for itr.Next() {
+		basket, err := itr.Value()
+		if err != nil {
+			return nil, err
+		}
+
+		bb, err := k.stateStore.BasketBalanceTable().Get(ctx, basket.Id, basket.BasketDenom)
```
ryanchristo, line 25: I'm a bit confused why we're getting the balance of the basket tokens here and not the balance of the credits within the basket. A basket denom is the bank denom that takes the form of `eco.dC.NCT` whereas a batch denom is the ecocredit denom that takes the form of `C01-00000000-00000000-001`. We should be looping through the batches within a basket to get the credit balance of each batch within the basket.

x/ecocredit/server/basket/util.go
aleem1314: > Same here. A batch should never exist with a basket denom as the batch denom. I think you want to loop through the batches within a basket and get the credit balance of each batch to check against the total supply. We don't need to check the balance of the basket denom against anything in core but we do need to check the balance of the credits within baskets against the total supply.

updated

x/ecocredit/server/invariants_test.go
```diff
@@ -1,371 +1,268 @@
...
-			ecocredit.SetDecimal(store, key, d)
-		}
+
+		require.NoError(t, ss.BatchBalanceTable().Save(ctx, &api.BatchBalance{
+			Address:  b.Address,
+			BatchId:  b.BatchId,
+			Tradable: b.Tradable,
+			Retired:  b.Retired,
+			Escrowed: b.Escrowed,
+		}))
```
ryanchristo, line 254: I think we want to use `Insert` here rather than `Save`. `Save` overwrites an existing entry if it already exists but `Insert` will error if there is an existing entry, which we are not expecting.

x/ecocredit/server/invariants_test.go
```diff
@@ -1,371 +1,268 @@
...
-		require.NoError(t, err)
-		key := ecocredit.TradableSupplyKey(denomT)
-		ecocredit.AddAndSetDecimal(store, key, d)
-		d, err = math.NewNonNegativeDecFromString(s.RetiredSupply)
+		err := ss.BatchSupplyTable().Save(ctx, &api.BatchSupply{
+			BatchId:         s.BatchId,
+			TradableAmount:  s.TradableAmount,
+			RetiredAmount:   s.RetiredAmount,
+			CancelledAmount: s.CancelledAmount,
+		})
```
ryanchristo, line 265: Same here `Insert` rather than `Save`.

x/ecocredit/server/invariants_test.go
```diff
@@ -1,371 +1,268 @@
...
-					TradableBalance: "210",
-				},
-				{
-					Address:        acc2.String(),
-					BatchDenom:     "1/2",
-					RetiredBalance: "210",
+					Address:  acc1,
+					BatchId:  1,
+					Tradable: "210",
+					Retired:  "110",
```
ryanchristo, line 88: I think we should include two accounts with a balance for the same batch. It looks like we only test two balances for two separate batches. We should either add here or add a new test case.

ryanchristo (commented):
Looking good. A few comments and suggestions.

technicallyty (commented):
looks good, a few suggestions

ryanchristo (approved):
Looks good to me. Variable names using `id` should be using `key`. Pre-approving.

technicallyty (approved):
looks good! 👍🏻

Issue #1012, opened 2022-04-12
Use "jurisdiction" for projects and retirement


ryanchristo:
## Summary

We currently use "location" in `project_location` and `retirement_location`. This might cause confusion as these locations are not intended to represent a geographic location but rather a "jurisdiction".

## Problem Definition

The use of "location" in `project_location` and `retirement_location` might cause confusion as these locations are not intended to represent a geographic location but rather a "jurisdiction".

## Proposal

Update `project_location` and `retirement_location` to `project_jurisdiction` and `retirement_jurisdiction`.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1018, opened 2022-04-13
test(x/ecocredit): fix intermittent race condition

ryanchristo:
## Description

This pull request fixes intermittent race condition caused by random string generator in tests:

```
  github.com/cosmos/cosmos-sdk/types/simulation.RandStringOfLength()
      /home/runner/go/pkg/mod/github.com/regen-network/cosmos-sdk@v0.45.0-regen-1/types/simulation/rand_util.go:29 +0x114
  github.com/regen-network/regen-ledger/x/ecocredit/core.TestMsgCreateClass()
      /home/runner/work/regen-ledger/regen-ledger/x/ecocredit/core/msg_create_class_test.go:92 +0xc0d
```

```
  github.com/cosmos/cosmos-sdk/types/simulation.RandStringOfLength()
      /home/runner/go/pkg/mod/github.com/regen-network/cosmos-sdk@v0.45.0-regen-1/types/simulation/rand_util.go:27 +0x7c
  github.com/regen-network/regen-ledger/x/ecocredit/core.TestMsgCreateBatch()
      /home/runner/work/regen-ledger/regen-ledger/x/ecocredit/core/msg_create_batch_test.go:235 +0x1f6d
```

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] followed the guidelines for [building modules](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules)
- [ ] included the necessary unit and integration [tests](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#testing)
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

[...]

PR #1020, opened 2022-04-13
refactor(x/ecocredit): update location to jurisdiction

technicallyty:
## Description

- changes usages of location to jurisdiction

Closes: #1012 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/msgs.go
```diff
@@ -125,7 +125,7 @@ func (m *MsgCreateBatch) ValidateBasic() error {
 			}
 
 			if !retiredAmount.IsZero() {
-				if err = ValidateLocation(iss.RetirementLocation); err != nil {
+				if err = ValidateJurisdiction(iss.RetirementLocation); err != nil {
```
ryanchristo, line 128: Looks like issuance still needs to be updated for `MsgCreateBatch`.

x/ecocredit/msgs.go
```diff
@@ -182,7 +182,7 @@ func (m *MsgSend) ValidateBasic() error {
 		}
 
 		if !retiredAmount.IsZero() {
-			if err = ValidateLocation(credit.RetirementLocation); err != nil {
+			if err = ValidateJurisdiction(credit.RetirementLocation); err != nil {
```
ryanchristo, line 185: Also here for credits within `MsgSend`.

x/ecocredit/msgs.go
```diff
@@ -227,7 +227,7 @@ func (m *MsgRetire) ValidateBasic() error {
 		}
 	}
 
-	if err := ValidateLocation(m.Location); err != nil {
+	if err := ValidateJurisdiction(m.Location); err != nil {
```
ryanchristo, line 230: Also within `MsgRetire`.

x/ecocredit/msgs.go
```diff
@@ -506,7 +506,7 @@ func (m *MsgBuy) ValidateBasic() error {
 		}
 
 		if order.RetirementLocation != "" {
-			if err := ValidateLocation(order.RetirementLocation); err != nil {
+			if err := ValidateJurisdiction(order.RetirementLocation); err != nil {
```
ryanchristo, line 509: Also within `MsgBuy`.

x/ecocredit/msgs.go
```diff
@@ -578,7 +578,7 @@ func (m *MsgCreateProject) ValidateBasic() error {
 		return ErrMaxLimit.Wrap("create project metadata")
 	}
 
-	if err := ValidateLocation(m.ProjectLocation); err != nil {
+	if err := ValidateJurisdiction(m.ProjectLocation); err != nil {
```
ryanchristo, line 581: And here so for `MsgCreateProject`.

x/ecocredit/server/core/send.go
```diff
@@ -151,7 +151,7 @@ func (k Keeper) sendEcocredits(ctx context.Context, credit *core.MsgSend_SendCre
 			Retirer:    to.String(),
 			BatchDenom: credit.BatchDenom,
 			Amount:     sendAmtRetired.String(),
-			Location:   credit.RetirementLocation,
+			Location:   credit.RetirementJurisdiction,
```
ryanchristo, line 154: And here for `EventRetire`.

ryanchristo (commented):
Looks like `Location` is still showing up in a few places.

x/ecocredit/msgs.go
technicallyty: i dont think its really necessary to update these, it's pretty much just waiting to be deleted at this point. all the messages have already been migrated to `core/` 🤷🏻

x/ecocredit/server/core/send.go
technicallyty: ah, looks looks like this was using an outdated event too! good catch 👍🏻

x/ecocredit/msgs.go
ryanchristo: Yup, that makes sense.

x/ecocredit/server/core/send.go
ryanchristo: Nice. At least one of my comments was relevant.

ryanchristo:
Also looks like another case of random string used in message validation tests causing a race condition:

https://github.com/regen-network/regen-ledger/runs/6032177203?check_suite_focus=true 

Resolved with #1018 when you have a chance to review.

ryanchristo:
I think some of these should just be `jurisdiction` rather than `project_jurisdiction` or `retirement_jurisdiction`.

I think it would be safe to change the following:

- `ProjectInfo` should  use `jurisdiction` rather than `project_jurisdiction`.
- `MsgCreateProject` should use `jurisdiction` rather than `project_jurisdiction`.

I added #1036 to track other changes that we may want to consider but the two changes listed might be nice to include within this pull request.

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -114,13 +114,13 @@ message MsgCreateProject {
...
   string metadata = 3;
 
-  // project_location is the location of the project backing the credits in this
-  // batch. It is a string of the form
+  // project_jurisdiction is the jurisdiction of the project backing the credits
+  // in this batch. It is a string of the form
   // <country-code>[-<sub-national-code>[ <postal-code>]], with the first two
   // fields conforming to ISO 3166-2, and postal-code being up to 64
   // alphanumeric characters. country-code is required, while sub-national-code
   // and postal-code can be added for increasing precision.
```
aaronc, line 122:
```suggestion
  // and postal-code are optional and can be added for increased precision.
```

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -273,13 +272,12 @@ message MsgSend {
...
-    // is a string of the form
-    // <country-code>[-<sub-national-code>[ <postal-code>]], with the first two
-    // fields conforming to ISO 3166-2, and postal-code being up to 64
-    // alphanumeric characters.
-    string retirement_location = 4;
+    // retirement_jurisdiction is the jurisdiction of the beneficiary or buyer
+    // of the retired credits. This must be provided if retired_amount is
+    // positive. It is a string of the form <country-code>[-<sub-national-code>[
+    // <postal-code>]], with the first two fields conforming to ISO 3166-2, and
+    // postal-code being up to 64 alphanumeric characters.
```
aaronc, line 279: It's maybe not a good idea to duplicate all these docs

aaronc (approved):
Reviewed API changes primarily and those LGTM

ryanchristo:
> I think some of these should just be jurisdiction rather than project_jurisdiction or retirement_jurisdiction. I think it would be safe to change the following:
>
> ProjectInfo should use jurisdiction rather than project_jurisdiction.
> MsgCreateProject should use jurisdiction rather than project_jurisdiction.
>
> I added https://github.com/regen-network/regen-ledger/issues/1036 to track other changes that we may want to consider but the two changes listed above might be nice to include within this pull request.

It looks like the above changes are the only changes necessary needed for #1036 and would have been nice to include here. Also would have been nice to address the proto documentation suggestion aaron made. Let's make sure are are addressing suggestions before merging in the future.

Issue #992, opened 2022-04-07
Consistent use of id, name, and abbreviation


ryanchristo:
## Summary

In the ecocredit module, there are inconsistencies with how we use `id`, `name`, and `abbrev`.

When migrating to ORM tables, we added an `id` field to state objects, a `uint64` that represents the primary key for `ClassInfo`, `ProjectInfo`, and `BatchInfo`, but we also have `class_id` and `project_id`, a `string` that represents the id/name/abbreviation of the object and holds semantic value (i.e. `C01` is the first credit class of the `C` credit type, and `C0101` is the first project of the `C01` credit class if a `project_id` is not specified upon creation).

With `CreditType`, we use `abbreviation` and reference it as `credit_type_abbrev` when creating a new credit class but we use `credit_type` to reference the same abbreviation in `ClassInfo` and in `ClassSequence`.

In `ClassSequence`, we use a `uint64` for `next_class_id` that represents the next sequence number (i.e. the next number to be used when generating a `class_id` that is different than the `id` and the `class_id` itself). We also do this in `BatchSequance` for `next_batch_id` that is different than the `id` and `batch_denom`).

In `ClassSequence`, we use `class_id` as the primary key, which is the `id` from `ClassInfo` (the `uint64` value), but in `BatchSequence` we use `project_id` as the primary key, which is the `name` from `ProjectInfo` (the `string` value).

## Problem Definition

In the ecocredit module, there are inconsistencies with how we use `id`, `name`, and `abbrev`. This could cause confusion to users and is already causing confusion among our developers.

## Proposal

[ work in progress ]

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1021, opened 2022-04-14
refactor(x/ecocredit): consistent use of id, key, and abbreviation

ryanchristo:
## Description

Closes: #992

This pull request updates `id` in state proto to `key` and applies consistent use of id, key, and abbreviation.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1/state.proto
```diff
@@ -12,40 +12,44 @@ option go_package = "github.com/regen-network/regen-ledger/x/ecocredit/core";
...
+    primary_key : {fields : "row_id", auto_increment : true}
     index : {id : 1 fields : "name", unique : true}
     index : {id : 2 fields : "admin"}
     index : {id : 3 fields : "credit_type"}
   };
 
-  // id is the unique ID of credit class.
-  uint64 id = 1;
+  // row_id is the numeric table row identifier of the credit class.
+  uint64 row_id = 1;
```
aaronc, line 48: How about something more cryptic like dbid, rowid, or dbrow? To make it less human readable

ryanchristo: I think `rowid` makes the most sense out of those options. The type would be `Rowid` then rather than `RowId` though.

aaronc: How about db_key?

ryanchristo: When I see `db_id` or `db_key` I think the id for the database or the key for the database, not an item in the database. I still think `row_id` makes the most sense to me but I don't have a strong preference.

aaronc: I'm just trying to find a way we can avoid `id` which is used in other places like `projet_id`

ryanchristo: That makes sense. We can think on it some more. I do like the idea of `key` but `row_key` seems a bit odd.

ryanchristo: What about `pk` for primary key?

ryanchristo: That would probably get confusing though with `BatchBalance` that used the `pk` of the credit batch but the primary key for the `BatchBalance` table is actually `address,row_id`.

proto/regen/ecocredit/v1/events.proto
```diff
@@ -9,7 +9,7 @@ option go_package = "github.com/regen-network/regen-ledger/x/ecocredit/core";
 // EventCreateClass is an event emitted when a credit class is created.
 message EventCreateClass {
 
-  // class_id is the unique ID of credit class.
+  // class_id is the unique string identifier of the credit class.
```
robert-zaremba, line 12: do we need to repeat `string` here?

proto/regen/ecocredit/v1/state.proto
robert-zaremba: why you want to make it cryptic? If it's an identifier then I think `id` is the right name. Since here we have other `id` field, maybe we can call this `seq` or `key`?

`pk` might be confusing.

proto/regen/ecocredit/v1/state.proto
```diff
@@ -108,25 +113,27 @@ message ProjectInfo {
...
     index : {id : 1 fields : "batch_denom", unique : true}
     index : {id : 2 fields : "project_id"}
     index : {id : 3 fields : "start_date"}
     index : {id : 4 fields : "issuer"}
   };
 
-  // id is an auto-incrementing integer to succinctly identify the batch
-  uint64 id = 1;
+  // row_id is the numeric table row identifier of the credit batch.
+  uint64 row_id = 1;
```
robert-zaremba, line 124: here we don't have "other" `id`, so why renaming? I think ID should be the canonical attribute used when we reference the objects.

proto/regen/ecocredit/v1/events.proto
ryanchristo: Might not be necessary.

proto/regen/ecocredit/v1/state.proto
ryanchristo: I think we should be consistent. It might be confusing if some tables use `id` and others use `key` or something else.

proto/regen/ecocredit/v1/state.proto
ryanchristo: I think cryptic because the user should never be exposed to this. This field is only meant to be used internally.

I think we should be consistent with whatever we choose, i.e. not use `id` for some tables and `key` or something else for others. I would prefer `key` over `seq` because we use `next_sequence` in `ClassSequence`, `ProjectSequence` and `BatchSequence` and each of those would also need a numeric identifier.

ryanchristo: This field is only meant to be used internally. I think cryptic because the user should never be exposed to this identifier so we would make that more explicit by making it cryptic.

I think we should be consistent with whatever we choose, i.e. not use `id` when we can and `key` or something else when we can't. I would prefer `key` over `seq` because we use `next_sequence` in `ClassSequence`, `ProjectSequence` and `BatchSequence` and each of those would also need a numeric identifier.

ryanchristo: I've updated to `row_key` for now and will start implementing but happy to change if there is another preference.

proto/regen/ecocredit/v1/state.proto
```diff
@@ -108,25 +111,28 @@ message ProjectInfo {
...
+  // project_key is the table row identifier of the credit class used internally
+  // for efficient lookups. This links a credit batch to a project.
+  uint64 project_key = 3;
 
-  // batch_denom is the unique string identifier of the credit batch formed
-  // from the project name, batch sequence number and dates.
+  // batch_denom is the unique identifier of the credit batch formed from the
+  // project id, the batch sequence number, and the start and end date of the
+  // credit batch.
   string batch_denom = 4;
```
ryanchristo, line 136: This could also simply be `denom` rather than `batch_denom`. Opened #1031.

ryanchristo:
Any ideas why this is producing a group module error in the experimental simulation tests? This pull request does not make any changes to the group module. Maybe @aleem1314 you have an idea? 

```
panic: group accounts: unique constraint violation
```

https://github.com/regen-network/regen-ledger/runs/6067693726?check_suite_focus=true

aaronc (approved):

technicallyty (approved):
LGTM! highly in favor of the `batch_denom` -> `denom` change!!

PR #1034, opened 2022-04-18
test(x/data): data query acceptance tests

ryanchristo:
## Description

This pull request adds data query acceptance tests.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/data/server/features/query_attestors.feature
```diff
@@ -0,0 +1,206 @@
...
+  Background:
+    Given the content entry
+    """
+    {
+      "iri": "regen:13toVfvC2YxrrfSXWB5h2BGHiXZURsKxWUz72uDRDSPMCrYPguGUXSC.rdf",
+      "timestamp": "2020-01-01T00:00:00.000000000Z"
+    }
+    """
+
+  Rule: one attestor is returned if one attestor entry exists
```
ryanchristo, line 12: I don't think this is an accurate use of the Rule keyword but it would be nice to group cases together that use the same Background. Should this just be a scenario with sub-scenarios?

ryanchristo, line 12: I don't think this is an accurate use of the Rule keyword but it would be nice to group cases together that use the same Background. Should this just be a scenario with sub-scenarios?

aaronc: why can't this all be in the main background? or does it make sense to have separate files? i agree this isn't really a "rule"

aaronc (commented):
I personally don't think it adds as much value to have this style of acceptance tests for queries as for msg's. Most of this stuff just feels self evident based on how msg's and state work and doesn't really describe any new behavior. Whereas with msg's we are really describing the _behavior of the system_. What do you think @ryanchristo?

x/data/server/features/query_attestors.feature
ryanchristo: > why can't this all be in the main background?

The attestor entries were not in the main background because there is a scenario for each query in which no attestor entries exist, but it might make more sense to create a separate content entry in the main background.

We also might want to consider avoiding the Background keyword. The authors of the BDD books have requested the Background keyword be removed from Gherkin and they say it may be removed in the future (with options to support backwards compatibility). "We believe that the Background keyword should not be used, because it has a negative effect on the readability of the scenarios."

> or does it make sense to have separate files?

I was mimicking the current file structure and attempting to limit duplication, but I think these are technically separate features so separate files would probably make more sense. Also, the BDD Formulation book recommends simplicity and readability over duplication and redundancy when writing test code.

ryanchristo:
> I personally don't think it adds as much value to have this style of acceptance tests for queries as for msg's. Most of this stuff just feels self evident based on how msg's and state work and doesn't really describe any new behavior. Whereas with msg's we are really describing the behavior of the system. What do you think @ryanchristo?

Hmm... I think we've already introduced two styles of using Gherkin, one style that is more aligned with the examples in the BDD Formulation book that illustrates user stories (i.e. the acceptance tests for the message server implementation) and another style that is more focused on documenting functionality (i.e. the acceptance tests for message and content hash validation).

The BDD Formulation book does not include examples for the latter given the focus of the book is behavior-driven development but it does point to [an article on using Gherkin for writing APIs](https://specflow.org/challenges/test-apis-with-gherkin/), which is more focused on documenting functionality. The authors also mention, "Cucumber and Gherkin were created to support a behaviour-driven approach to software specification and delivery. However, many teams that use Cucumber use Gherkin scenarios to test functionality that has already been implemented, rather than to drive its development. While Cucumber can be used in this way, it has nothing to do with BDD."

I do think we should prioritize feature files that are more aligned with behavior-driven development, but the question is then, do we want to use Gherkin and Cucumber to test and document other functionality? I think the message and content hash validation was a nice addition and queries might be as well (although maybe less so). We recently removed message validation criteria from the ecocredit documentation because it was outdated and now we only point to the proto files on buf schema registry. The proto files only provide so much information and no specific examples. If we need to write tests for queries, it might be nice to write our tests and documentation together so that the documentation would always be up-to-date.

x/data/server/features/query_attestors.feature
ryanchristo: > We also might want to consider avoiding the Background keyword.

This might be more relevant when illustrating user stories and less so in cases like this.

clevinson:
@ryanchristo Is this PR still relevant? Looks like we already have feature files for data module.

ryanchristo:
> @ryanchristo Is this PR still relevant? Looks like we already have feature files for data module.

We don't have feature files for queries. If we close this, we should create a new issue and link the issue to this pull request. Whether we should write feature files for query methods is not high priority but still worth considering.

Issue #1035, opened 2022-04-19
Remove unnecessary fields from events


ryanchristo:
## Summary

Events should return only the necessary information that can then be used to lookup relevant details (e.g. creating a project should emit an event that includes the project id so that the project information can be looked up using the id rather than the event including all the project information).

## Problem Definition

Events are not intended to be used in place of queries and should only include the necessary information to look up more information with queries.

## Proposal

Remove unnecessary fields from events.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

aaronc:
👍

PR #1044, opened 2022-04-20
refactor(ecocredit): remove unnecessary fields in events

technicallyty:
## Description

Ref: #1035 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/basket/v1/events.proto
```diff
@@ -27,11 +25,8 @@ message EventPut {
   // basket_denom is the basket bank denom that the credits were added to.
   string basket_denom = 2;
 
-  // credits are the credits that were added to the basket.
-  repeated BasketCredit credits = 3;
```
ryanchristo, line 31: We are not bumping the basket proto version in this release so we may need to deprecate and add rather than updating the existing messages and fields:

https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-044-protobuf-updates-guidelines.md

We started using `Since Revision 1` for fields we are adding rather than what would be `Since: v4.0` based on the guidelines linked above. Let's discuss this on our call tomorrow.

ryanchristo (commented):
Looking good although we might need to make adjustments for the basket changes because we are currently not bumping the basket proto version in this next release.

proto/regen/ecocredit/basket/v1/events.proto
ryanchristo: Based on our discussion, we should keep the field and mark it as deprecated and remove in the next version.

technicallyty: reverted https://github.com/regen-network/regen-ledger/pull/1044/commits/4e607ee07f8cf91e96cbc564bf5dca0df6f94cbd

proto/regen/ecocredit/basket/v1/events.proto
```diff
@@ -44,9 +39,6 @@ message EventTake {
   // basket_denom is the basket bank denom that credits were taken from.
   string basket_denom = 2;
 
-  // credits are the credits that were taken from the basket.
-  repeated BasketCredit credits = 3;
```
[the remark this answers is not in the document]
technicallyty: https://github.com/regen-network/regen-ledger/pull/1044/commits/268bea4810a59cb1d000e91cf2747d5b65413d55

proto/regen/ecocredit/basket/v1/events.proto
ryanchristo: > and mark it as deprecated

Can we market it as deprecated? :pray:

proto/regen/ecocredit/basket/v1/events.proto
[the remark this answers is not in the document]
ryanchristo: > and mark it as deprecated

Can we market it as deprecated? :pray:

proto/regen/ecocredit/basket/v1/events.proto
technicallyty: woops, added: https://github.com/regen-network/regen-ledger/pull/1044/commits/99bf0286a9492869793ef77ee2540424a1686996

proto/regen/ecocredit/basket/v1/events.proto
```diff
@@ -6,7 +6,7 @@ import "regen/ecocredit/basket/v1/types.proto";
 
 option go_package = "github.com/regen-network/regen-ledger/x/ecocredit/basket";
 
-// EventCreate is an event emitted when a basket is created.
+// DEPRECATED: EventCreate is an event emitted when a basket is created.
```
ryanchristo, line 9: This event is not deprecated and does not include any fields that should be deprecated.

```suggestion
// EventCreate is an event emitted when a basket is created.
```

proto/regen/ecocredit/basket/v1/events.proto
```diff
@@ -17,8 +17,8 @@ message EventCreate {
   string curator = 2;
 }
 
-// EventPut is an event emitted when credits are put into a basket in return
-// for basket tokens.
+// DEPRECATED: EventPut is an event emitted when credits are put into a basket
+// in return for basket tokens.
```
ryanchristo, line 21: This event is not deprecated and only the `credits` field should be marked as deprecated.

```suggestion
// EventPut is an event emitted when credits are put into a basket
// in return for basket tokens.
```

ryanchristo: I'm unable to make a suggestion for the `credits` field, which is the field we want to deprecate, but I think we should follow the same format as the sdk.

```
  // credits are the credits that were added to the basket.
  // Deprecated: this field has been deprecated and will always return empty.
  repeated BasketCredit credits = 3;
```

proto/regen/ecocredit/basket/v1/events.proto
```diff
@@ -34,8 +34,8 @@ message EventPut {
   string amount = 4;
 }
 
-// EventTake is an event emitted when credits are taken from a basket starting
-// from the oldest credits first.
+// DEPRECATED: EventTake is an event emitted when credits are taken from a
+// basket starting from the oldest credits first.
```
ryanchristo, line 38: This event is not deprecated and only the `credits` field should be marked as deprecated.

```suggestion
// EventTake is an event emitted when credits are taken from a
// basket starting from the oldest credits first.
```

ryanchristo: I'm unable to make a suggestion for the `credits` field, which is the field we want to deprecate, but I think we should follow the same format as the sdk.

```
  // credits are the credits that were taken from the basket.
  // Deprecated: this field has been deprecated and will always return empty.
  repeated BasketCredit credits = 3;
```

ryanchristo (commented):
We shouldn't be marking the event messages as deprecated, only the `credits` field within the events that have it as deprecated. We should also avoid `DEPRECATED` in all caps and follow the same format as the sdk. We don't need to return the credits in the events if they are deprecated and we can specify that in the deprecated comment and revert https://github.com/regen-network/regen-ledger/pull/1044/commits/4e607ee07f8cf91e96cbc564bf5dca0df6f94cbd.

proto/regen/ecocredit/basket/v1/events.proto
ryanchristo: We can also say `this field will be removed in the next version` and return the credits in the event. I think either way would be ok but we may want to provide backwards compatibility. What do you think?

proto/regen/ecocredit/basket/v1/events.proto
ryanchristo: Same as above. We can also say `this field will be removed in the next version` and return the credits in the event. I think either way would be ok but we may want to provide backwards compatibility. What do you think?

proto/regen/ecocredit/basket/v1/events.proto
technicallyty: added & put the depreacted proto option as i saw fields do in the SDK

x/ecocredit/server/marketplace/update_sell_orders.go
```diff
@@ -45,10 +45,13 @@ func (k Keeper) UpdateSellOrders(ctx context.Context, req *marketplace.MsgUpdate
...
 	var creditType *ecoApi.CreditType
-	event := marketplace.EventUpdateSellOrder{}
+	event := marketplace.EventUpdateSellOrder{
+		Seller:            sdk.AccAddress(order.Seller).String(),
+		SellOrderId:       order.Id,
+		DisableAutoRetire: update.DisableAutoRetire,
+	}
 
 	order.DisableAutoRetire = update.DisableAutoRetire
-	event.DisableAutoRetire = update.DisableAutoRetire
```
ryanchristo, line 51: I think we should remove this from the event but in the current definition of the event, this is one of the updated fields, so this should not have been removed given the current event definition.

x/ecocredit/server/marketplace/update_sell_orders.go
```diff
@@ -119,8 +121,8 @@ func (k Keeper) applySellOrderUpdates(ctx context.Context, order *api.SellOrder,
 				return err
 			}
 			order.Quantity = update.NewQuantity
-			event.NewQuantity = update.NewQuantity
 		}
+		event.NewQuantity = update.NewQuantity
```
ryanchristo, line 125: As mentioned above, I think we should remove this from the event and only include the seller address and the sell order id when emitting an event for updating the sell order.

ryanchristo (commented):
This is a step in the right direction but the current changes only partially solve the issue. We are still returning a lot of unnecessary fields for events and we should further refine. I would be happy to do a pair-programming session on this to provide clarity on the proposed changes.

I also think it might be helpful to take a look at the group module and how minimal those events are, although I do believe including the address such as the admin, issuer, etc. is something we should include and which is not included in those events.

https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/group/v1/events.proto

x/ecocredit/server/marketplace/update_sell_orders.go
technicallyty: yeah i had removed it originally in this PR, but during a standup it was said we should keep fields that are updated in events so brought (most) of the fields back. 

ill just axe it.

proto/regen/ecocredit/marketplace/v1/events.proto
```diff
@@ -2,44 +2,27 @@ syntax = "proto3";
...
+  uint64 order_id = 2;
 }
 
 // EventUpdateSellOrder is an event emitted when a sell order is updated.
 message EventUpdateSellOrder {
 
-  // owner is the owner of the sell orders.
-  string owner = 1;
+  // seller is the owner of the sell order.
+  string seller = 1;
```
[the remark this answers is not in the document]
technicallyty: switched to `owner`

proto/regen/ecocredit/marketplace/v1/tx.proto
```diff
@@ -158,23 +154,3 @@ message MsgBuyDirect {
 
 // MsgBuyDirectResponse is the Msg/BuyDirect response type.
 message MsgBuyDirectResponse {}
-
-// MsgAllowAskDenom is the Msg/AllowAskDenom request type.
-message MsgAllowAskDenom {
-  // root_address is the address of the governance account which can authorize
-  // ask denoms
```
technicallyty, line 165: removed this as we did with `Buy` as it will **not** be used in v4.0

aleem1314 (approved):
lgtm

proto/regen/ecocredit/basket/v1/events.proto
ryanchristo: > ...but we may want to provide backwards compatibility. What do you think?

I'm thinking it might be better to provide backwards compatibility in the same way we do for basket queries and we may want to do the same for `amount` as well. See https://github.com/regen-network/regen-ledger/pull/1056#discussion_r861065178.

proto/regen/ecocredit/marketplace/v1/events.proto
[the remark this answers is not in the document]
ryanchristo: > I think seller makes sense but we should probably be consistent. I would support those changes here or we can open a separate issue to do a more thorough audit of all proto fields and comments for marketplace.

Apologies again. My comment was not very clear. I think we should be using `seller` rather than `owner` for consistency but this can be handled in a followup when auditing the proto files.

ryanchristo (commented):
> I also think it might be helpful to take a look at the group module and how minimal those events are. I do think it would be nice to include the address such as the admin, issuer, etc, which is not included in the group events, but not necessary because that information can be queried using the main identifier included in the event.
> 
> https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/group/v1/events.proto

For the update events in core...

- `EventClassAdminUpdated`
- `EventClassIssuersUpdated`
- `EventClassMetadataUpdated`
- `EventProjectAdminUpdated`
- `EventProjectMetadataUpdated`

...we could reduce the fields to just the identifier that was updated, which is how it is done in the group module. To avoid dragging this out, I can approve and we can get these changes merged and addressed in a followup. I think this is a step in the right direction and would be happy to do pair-programming on the proposed changes if you would prefer to handle them in this pull request. Otherwise I think we should keep the issue open upon merging.

technicallyty:
> > I also think it might be helpful to take a look at the group module and how minimal those events are. I do think it would be nice to include the address such as the admin, issuer, etc, which is not included in the group events, but not necessary because that information can be queried using the main identifier included in the event.
> > https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/group/v1/events.proto
> 
> For the update events in core...
> 
> * `EventClassAdminUpdated`
> * `EventClassIssuersUpdated`
> * `EventClassMetadataUpdated`
> * `EventProjectAdminUpdated`
> * `EventProjectMetadataUpdated`
> 
> ...we could reduce the fields to just the identifier that was updated, which is how it is done in the group module. To avoid dragging this out, I can approve and we can get these changes merged and then address my comments in a followup. I think this is a step in the right direction and would be happy to do pair-programming on the proposed changes if you would prefer to handle them in this pull request. Otherwise I think we should keep the issue open after merging these changes.

i'm ok with either merging or you can just take over this PR 👍🏻

ryanchristo (approved):
This is a step in the right direction. Approving and merging but keeping the issue open and will add a followup pull request. I started making adjustments but stumbled into a bit of a rabbit hole with how we've been writing events and how I think we can improve and I think a separate pull request would make more sense with the updates I would like to make.

Issue #1053, opened 2022-04-25
Update basket queries and curator address in state


ryanchristo:
## Summary

As a temporary solution to #930, we should add a separate query proto message called `BasketInfo` that returns basket information without the `id` field (only used internally for efficient lookups) and with `curator` as a string. Also, the `curator` field in the state proto message for `Basket` is currently a string but this should be updated to bytes.

The same should be done for `BasketBalance`. We should create a query proto message called `BasketBalanceInfo` that does not include `basket_id` (only used internally for efficient lookups) but rather `basket_denom`.

## Problem Definition

We should be using bytes for addresses in state for efficient storage and then returning human-readable information in query responses and we should not be including fields that are only used internally within those responses.

## Proposal

- Update `curator` field in the `Basket` state proto message to bytes
- Add `BasketInfo` query proto message that includes `curator` as a string and excludes `id`
- Update `QueryBasketResponse` and `QueryBasketsResponse` to use `BasketInfo`
- Add `BasketBalanceInfo` query proto message that includes `basket_denom` and excludes `basket_id`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1056, opened 2022-04-26
refactor(ecocredit): Update basket queries and curator address in state

aleem1314:
## Description

Closes: #1053 

- Update `curator` field in the `Basket` state proto message to bytes
- Add `BasketInfo` query proto message that includes `curator` as a string and excludes `id`
- Update `QueryBasketResponse` and `QueryBasketsResponse` to use `BasketInfo`
- Add `BasketBalanceInfo` query proto message that includes `basket_denom` and excludes `basket_id`

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/basket/v1/query.proto
```diff
@@ -46,7 +47,7 @@ message QueryBasketRequest {
 // QueryBasketResponse is the Query/Basket response type.
 message QueryBasketResponse {
   // basket is the queried basket.
-  Basket basket = 1;
+  BasketInfo basket = 1;
```
ryanchristo, line 50: We are not bumping the basket proto version in this release so we may need to deprecate and add rather than updating the existing messages and fields:

https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-044-protobuf-updates-guidelines.md

We started using `Since Revision 1` for fields we are adding rather than what would be `Since: v4.0` based on the guidelines linked above. Let's discuss this on our call tomorrow.

proto/regen/ecocredit/basket/v1/state.proto
```diff
@@ -48,7 +48,7 @@ message Basket {
   // basket settings.
   //
   // Since Revision 1
-  string curator = 8;
+  bytes curator = 8;
```
ryanchristo, line 51: Here's an example of a revision comment since we added this field since the original v1 proto version released as part of Regen Ledger v3.0.

ryanchristo (commented):
Looking good although we might need to make adjustments for the basket changes because we are currently not bumping the basket proto version in this next release.

x/ecocredit/server/basket/create.go
```diff
@@ -40,7 +40,7 @@ func (k Keeper) Create(ctx context.Context, msg *basket.MsgCreate) (*basket.MsgC
 	}
 
 	id, err := k.stateStore.BasketTable().InsertReturningID(ctx, &api.Basket{
-		Curator:           msg.Curator,
+		Curator:           sender,
```
blushi, line 43: why not using just `curator` as variable name here?

proto/regen/ecocredit/basket/v1/query.proto
amaurym: I'd also go with the former approach. If possible, the deprecated field should be populated too, to offer maximum backwards compatibiliy.

ryanchristo (approved):
This looks good to me.

ryanchristo:
@technicallyty can you review when you get a chance? :pray:

proto/regen/ecocredit/basket/v1/query.proto
```diff
@@ -98,4 +115,48 @@ message QueryBasketBalanceRequest {
...
+  // balance is the amount of ecocredits held in the basket
+  string balance = 2;
+
+  // basket_denom is the basket bank denom.
+  string basket_denom = 3;
+
+  // batch_start_date is the start date of the batch. This field is used
+  // to create an index which is used to remove the oldest credits first.
+  google.protobuf.Timestamp batch_start_date = 4;
+}
```
technicallyty, line 162: should we leave some of these fields out? `batch_start_date` was only in the table so it could be used as an index, and the `basket_denom` seems a bit redundant since its in the request already.

Issue #1084, opened 2022-05-03
Additional tasks for supporting Polygon bridge work


clevinson:
There's two additional features that came up in today's Ledger Product call, which we'd like to prioritize adding in for Llangorse Update (#818):

- [x] Support for deduplication on tx_origin #1112
- [x] add custom bridging support for cancelling credits #1101 

### Support for deduplication on tx_origion

Any call to `CreateBatch` or `MintBatchCredits` should check to ensure that no credits in the given credit class have been issued with the same `tx_origin`. This should be done by adding an index / table for storing `tx_origin`'s in state, so lookups can be performed to ensure the `tx_origin` has not already been used.

### Add custom bridging support for cancelling credits @aleem1314 

To further support bidirectional bridging of credits (e.g. assets moving off Regen Ledger to Polygon) we've decided to add custom bridge support for Canceling credits.

Namely:
- Add a "reason" string field to MsgCancel (and add this field to the corresponding event)
- Add a new `MsgBridge` that wraps a `MsgCancel`
  - has the following additional fields:
    - bridge_target string
    - bridge_recipient string
    - bridge_contract string
  - has a separate "bridge" event which can be used by a bridge service to consume specific events

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1101, opened 2022-05-12
feat(ecocredit): custom bridging support for cancelling credits

aleem1314:
## Description

ref: #1084 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/core/msg_bridge.go
```diff
@@ -0,0 +1,49 @@
...
+
+	if err := m.MsgCancel.ValidateBasic(); err != nil {
+		return err
+	}
+
+	if len(m.BridgeContract) == 0 {
+		return sdkerrors.ErrInvalidRequest.Wrap("bridge contract should not be empty")
+	}
+
+	if len(m.BridgeRecipient) == 0 {
```
aleem1314, line 34: Do we need additional validations for these fields?

proto/regen/ecocredit/v1/events.proto
```diff
@@ -106,6 +106,9 @@ message EventCancel {
 
   // amount is the decimal number of credits that have been cancelled.
   string amount = 3;
+
+  // reason is any arbitary encoded data.
+  string reason = 4;
```
ryanchristo, line 111: The issue and notes are a bit unclear on this but because we decided to add a specific event for bridging credits from regen ledger to another chain (`EventBridge`) and because this added event includes the necessary data we need for the bridge operation, this field will actually be used to provide a more generic reason.

We still want this field here because we are not storing a reason in state and this would be valuable information to have but the reason will not include encoded data such as the target chain, recipient, and contract. Also, because this is an event field, we don't need to include details about how the string needs to be formatted.

```suggestion
  // reason is the reason the credits were cancelled.
  string reason = 4;
```

proto/regen/ecocredit/v1/events.proto
```diff
@@ -159,3 +162,15 @@ message EventAddCreditType {
...
+message EventBridge {
+  // bridge_target is the target chain.
+  string bridge_target = 1;
+
+  // bridge_recipient is the recipient address.
+  string bridge_recipient = 2;
+
+  //  bridge_contract is the contract address.
+  string bridge_contract = 3;
+}
```
ryanchristo, line 176: Although included in the issue, I think we should shorten these names and exclude the `bridge_` prefix.

cc @clevinson is that ok with you?

```suggestion
// EventBridge is emitted emitted when credits are bridged to another chain.
message EventBridge {
  // target is the target chain.
  string target = 1;

  // recipient is the recipient address.
  string recipient = 2;

  // contract is the contract address.
  string contract = 3;
}
```

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -321,6 +324,9 @@ message MsgCancel {
   // credits are the credits being cancelled.
   repeated CancelCredits credits = 2;
 
+  // reason is any arbitary encoded data, will be passed to an event.
+  string reason = 3;
```
ryanchristo, line 328: As mentioned above, this will be more generic given we decided to add a bridge specific event.

```suggestion
  // reason is any arbitrary string that specifies the reason for cancelling credits.
  string reason = 3;
```

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -421,3 +427,21 @@ message MsgUpdateProjectMetadata {
...
+  MsgCancel msg_cancel = 1;
+
+  // bridge_target is the target chain.
+  string bridge_target = 2;
+
+  // bridge_recipient is the recipient address.
+  string bridge_recipient = 3;
+
+  // bridge_contract is the contract address.
+  string bridge_contract = 4;
```
ryanchristo, line 443: Same as above, I think we should exclude the `bridge_` prefix on these. cc @clevinson 

```suggestion
  // target is the target chain.
  string target = 2;

  // recipient is the recipient address.
  string recipient = 3;

  // contract is the contract address.
  string contract = 4;
```

x/ecocredit/client/tx.go
```diff
@@ -370,14 +370,16 @@ Parameters:
...
+		Use:   "cancel [credits] [reason]",
 		Short: "Cancels a specified amount of credits from the account of the transaction author (--from)",
 		Long: `Cancels a specified amount of credits from the account of the transaction author (--from)
 
 Parameters:
   credits:  comma-separated list of credits in the form [<amount> <batch-denom>]
-            eg: '10 C01-20200101-20210101-001, 0.1 C01-20200101-20210101-001'`,
-		Args: cobra.ExactArgs(1),
+            eg: '10 C01-20200101-20210101-001, 0.1 C01-20200101-20210101-001'
+  reason:   reason is any arbitary encoded data.
```
ryanchristo, line 380:
```suggestion
  reason:   reason is any arbitrary string that specifies the reason for cancelling credits.
```

x/ecocredit/core/msg_bridge.go
```diff
@@ -0,0 +1,49 @@
...
+
+// ValidateBasic does a sanity check on the provided data.
+func (m *MsgBridge) ValidateBasic() error {
+
+	if err := m.MsgCancel.ValidateBasic(); err != nil {
+		return err
+	}
+
+	if len(m.BridgeContract) == 0 {
+		return sdkerrors.ErrInvalidRequest.Wrap("bridge contract should not be empty")
```
ryanchristo, line 31: Given this will error, we should probably say "cannot be empty" or "required". If we use "should not", it means it is not recommended but still possible.

```suggestion
		return sdkerrors.ErrInvalidRequest.Wrap("bridge contract cannot be empty")
```

x/ecocredit/core/msg_bridge.go
ryanchristo: Good question. We are only planning to support bridging to Polygon in the short-term but we are planning to bridge to other chains in the future. I think we should validate for an Ethereum address.

Updating the validation check could be done in a patch release if we need to change it but better to be more cautious given we are cancelling credits and this action cannot be undone.

The following is an example from the [Go Ethereum Documentation](https://goethereumbook.org/en/address-check/):

```
re := regexp.MustCompile("^0x[0-9a-fA-F]{40}$")

fmt.Printf("is valid: %v\n", re.MatchString("0x323b5d4c32345ced77393b3530b1eed0f346429d")) // is valid: true
fmt.Printf("is valid: %v\n", re.MatchString("0xZYXb5d4c32345ced77393b3530b1eed0f346429d")) // is valid: false
```

x/ecocredit/core/msg_bridge.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	if len(m.BridgeRecipient) == 0 {
+		return sdkerrors.ErrInvalidRequest.Wrap("bridge recipient address should not be empty")
+	}
+
+	if len(m.BridgeTarget) == 0 {
+		return sdkerrors.ErrInvalidRequest.Wrap("bridge target should not be empty")
+	}
```
ryanchristo, line 40: We might want to consider defining supported targets and validate against those. We could have enumerated values stored in the proto files (similar to how we do digest algorithm, media type, etc. in the date module). We would only support Polygon in this initial version. We should be as cautious as we can given cancelling credits cannot be undone. Ok to leave without for now but something to discuss.

ryanchristo (commented):
Looking good. Mostly nits and I think we should validate contract and recipient address. I'm also wondering if we should define supported bridge targets and validate target as well.

x/ecocredit/core/msg_bridge.go
aaronc: maybe not a bad idea to prevent unexpected fund transfers. should be gov gated in the future IMHO so a chain upgrade isn't always needed but maybe okay for now to just allow `polygon-tco2`

clevinson: if we're doing validation here, I prefer `polygon` as opposed to `polygon-tco2`, as we don't want a situation where there could be another way of interpreting the same polygon txhash for creating a different kind of ecocredit / bridge operation.

x/ecocredit/client/tx.go
```diff
@@ -370,14 +370,16 @@ Parameters:
...
-		Use:   "cancel [credits]",
+		Use:   "cancel [credits] [reason]",
 		Short: "Cancels a specified amount of credits from the account of the transaction author (--from)",
 		Long: `Cancels a specified amount of credits from the account of the transaction author (--from)
 
 Parameters:
   credits:  comma-separated list of credits in the form [<amount> <batch-denom>]
-            eg: '10 C01-20200101-20210101-001, 0.1 C01-20200101-20210101-001'`,
-		Args: cobra.ExactArgs(1),
+            eg: '10 C01-20200101-20210101-001, 0.1 C01-20200101-20210101-001'
```
ryanchristo, line 379:
```suggestion
            eg: '10 C01-001-20200101-20210101-001, 0.1 C01-001-20200101-20210101-001'
```

x/ecocredit/core/msg_bridge.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"regexp"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/cosmos/cosmos-sdk/x/auth/legacy/legacytx"
+	"github.com/regen-network/regen-ledger/x/ecocredit"
+)
+
+// TODO: remove after we open governance changes for bridge target
```
ryanchristo, line 12:
```suggestion
// TODO: remove after we open governance changes for bridge target
// https://github.com/regen-network/regen-ledger/issues/1119
```

x/ecocredit/server/testsuite/suite.go
```diff
@@ -609,6 +609,7 @@ func (s *IntegrationTestSuite) TestScenario() {
 						Amount:     tc.toCancel,
 					},
 				},
+				Reason: "insufficient funds",
```
ryanchristo, line 612:
```suggestion
				Reason: "bridging assets to another chain",
```

ryanchristo (approved):
Looks good to me. Nice work.

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -421,3 +427,21 @@ message MsgUpdateProjectMetadata {
 // MsgUpdateProjectMetadataResponse is the Msg/UpdateProjectMetadataResponse
 // response type.
 message MsgUpdateProjectMetadataResponse {}
+
+// MsgBridge is the Msg/Bridge request type.
+message MsgBridge {
+  // MsgCancel is the MsgCancel message type.
+  MsgCancel msg_cancel = 1;
```
clevinson, line 434: Looking at this again in more detail, I think it's preferable for us to not nest proto messages that are used as GRPC request messages like this. Ideally when defining msg server API endpoints, each can evolve independently (e.g. `Bridge()` and `Cancel()` should be able to evolve independently). So similar to how we always have explicitly separate `MsgBridgeRepsonse` and `MsgCancelResponse`, we should probably avoid embedding a `MsgCancel` field inside the `MsgBridge` message.

In terms of which fields from `MsgCancel` we need when a user is calling `MsgBridge`, I the following is sufficient:
- holder
- `repeated credits` (could use a similar sub-message to `CancelCredits` called `BridgeCredits`?)

My assumption is that we don't actually need an explicit "reason" in the bridge call, as most of the information we would want to pass in the cancel reason would be describing that it's a bridge action, and possibly what chain its coming from. But all of this info we can just get from the `target` field, and knowing its a `Bridge()` RPC call.

So for filling the "reason" in the cancel event, can we just hardcode it in the message server to set `reason = "bridge"` or `reason = "bridge-${target}"` ?

clevinson (changes requested):
Nice work! One thing I'm worried abt though is nesting `MsgCancel` within `MsgBridge`... feels like its not 100% necessary as I don't think we should expect `Bridge()` calls to explicitly provide a reason when they are already providing the `target` chain explicitly, and the RPC call is called "Bridge" which is essentially the reason for cancelling credits.

proto/regen/ecocredit/v1/tx.proto
ryanchristo: Just to note, this was implemented as specified in the issue. That being said, I do agree with the changes proposed. When filling the reason for `MsgCancel` in the message server implementation, I would be in favor of `"bridge-${target}"`.

clevinson: sorry abt the confusion!

x/ecocredit/core/msg_bridge.go
```diff
@@ -0,0 +1,79 @@
...
+			return err
+		}
+	}
+
+	if m.Target != BRIDGE_TARGET {
+		return sdkerrors.ErrInvalidRequest.Wrapf("expected %s got %s", BRIDGE_TARGET, m.Target)
+	}
+
+	if !isValidEthereumAddress(m.Recipient) {
+		panic("OOOPS")
```
ryanchristo, line 59:
```suggestion
```

clevinson (approved):
Thanks @aleem1314 !

Approving. One thing that came up in discussion with @ryanchristo today is that we may actually want to revise our thinking around "contract" being an explicit field in the `Bridge()` API call...

This was my oversight when writing out #1084... From a security perspective, we can't only rely on an end user's desired "contract" for tokens to get minted into on the destination chain. As this may not actually be the corresponding TCO2 contract for the given batch denom.  Since the bridge service will anyway need to manage their own mapping between batch denoms & TCO2 contracts on the destination chain, there probably isn't much added value in the user providing this upfront...

Issue #1047, opened 2022-04-21
Implement CLI tests for commands without them


ryanchristo:
## Summary

Following the proposal of #1041, a discussion around whether we should migrate our client tests to unit tests now or wait for updates upstream that will make writing CLI commands and tests obsolete.

It was proposed that for now we treat CLI tests as a "nice to have" in the backlog while we explore our options. This issue tracks CLI commands that do not have tests.

- `buy-direct` https://github.com/regen-network/regen-ledger/pull/1078
- `buy-direct-batch` https://github.com/regen-network/regen-ledger/pull/1103
- ~~credit-type-proposal~~ cannot be tested
- `update-project-admin` https://github.com/regen-network/regen-ledger/pull/1102
- `update-project-metadata` https://github.com/regen-network/regen-ledger/pull/1105
- `allowed-denoms` https://github.com/regen-network/regen-ledger/pull/1106

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

technicallyty:
looking like proposals cannot be tested, or at least not included in our tests. I tested that a proposal was created, however, leaving this test in breaks the app because:

the gov module wraps proposal commands in tx flags. because of this, adding any txFlags (such a --from, and --yes) causes the app to panic during build. and since i cannot add these flags, the test *always* fails, as the `--yes` flag missing causes the transaction to not execute. 

luckily, through this process i was able to catch some errors and fix both proposal handlers, however, it looks like these will have to be tested manually on a testnet, or localnet.

PR #1105, opened 2022-05-16
test(ecocredit): update project metadata integration test

technicallyty:
## Description

- adds integration test for updating project metadata

ref: #1047

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/client/testsuite/tx.go
```diff
@@ -1426,6 +1428,81 @@ func (s *IntegrationTestSuite) TestTxBuyDirect() {
 	}
 }
 
+func (s *IntegrationTestSuite) TestUpdateProjectMetadata() {
+	val0 := s.network.Validators[0]
+	valAddrStr := val0.Address.String()
+	clientCtx := val0.ClientCtx
+	cmd := coreclient.TxUpdateProjectMetadataCmd()
+	_, projectId, _ := s.createClassProjectBatch(clientCtx, valAddrStr)
```
ryanchristo, line 1436: With all these integration tests, I'm not sure why we are creating a new credit class, project, and batch for each test scenario. This is what we should be doing in a unit test, not necessarily an integration test. This also slows down our CI and tests. It's also not necessary to create a batch for this scenario.

x/ecocredit/client/testsuite/tx.go
```diff
@@ -1426,6 +1428,81 @@ func (s *IntegrationTestSuite) TestTxBuyDirect() {
...
+
+	unauthAcc, _, err := val0.ClientCtx.Keyring.NewMnemonic("unauthMetadata", keyring.English, sdk.FullFundraiserPath, keyring.DefaultBIP39Passphrase, hd.Secp256k1)
+	s.Require().NoError(err)
+	unauthAddr := unauthAcc.GetAddress()
```
ryanchristo, line 1440: We can leverage the testsuite setup here and use `s.addr` rather than creating a new account.

```suggestion
	unauthAddr := s.addr
```

ryanchristo (approved):
I'm a bit concerned about these tests further slowing down our CI. For example, we are creating a new class, project, and batch in each of these scenarios and for this particular scenario we don't need a new batch. Would be nice if we could tighten things up here and throughout but these are also not high priority and will be removed in the future. Pre-approving.

x/ecocredit/client/testsuite/tx.go
technicallyty: i could refactor to have some safety classes/projects that wont get trampled by other tests. the intention of the pattern was to avoid cases where other tests corrupt data needed by another test. 

> This is what we should be doing in a unit test, not necessarily an integration test

I don't think its necessarily detrimental thing to attempt to isolate individual integration test - its definitely a bit harder to achieve in a performant way with testing big systems like this though 🤷🏻 

> not necessary to create a batch for this scenario.

[chore: use new helper method, rename variables](https://github.com/regen-network/regen-ledger/pull/1105/commits/f15412e6d8561518a285494f1b1e12722a8f86fd), removes the batch creation part 👍🏻

x/ecocredit/client/testsuite/tx.go
technicallyty: fixed in https://github.com/regen-network/regen-ledger/pull/1105/commits/f15412e6d8561518a285494f1b1e12722a8f86fd

x/ecocredit/client/testsuite/tx.go
ryanchristo: Removing batch creation helps. Not sure if we should spend too much time on this and attempt a larger refactor given we are planning to move away from client integration tests. Isolate and integration seem contradictory to me but that would be assuming integration tests are testing the integration of our features, not the integration with the sdk.

Issue #1114, opened 2022-05-19
Prune Orders deletes nil value expirations


technicallyty:
## Summary of Bug

PruneOrders should delete all sell orders with expirations that fall within 0-<block_height>. Sell orders with `nil` expirations should not be deleted. 

The ORM treats `nil` values as Go's default 0 values, so orders with `nil` expirations are being deleted.

## Version

`main`

## Steps to Reproduce

1. Create sell orders with nil expiration
2. prune orders
3. see that the orders are now deleted

## Solution

1. create a uint64 slice for sell order id's
2. check in the iterator loop that expiration != nil, inserting into the slice if so
3. loop over the slice and delete the orders in the uint64 slice, instead of using `DeleteRange`

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1115, opened 2022-05-19
fix(ecocredit/marketplace): nil expirations no longer pruned

technicallyty:
## Description

Fixes a bug where sell orders with nil expirations were being pruned

Closes: #1114 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


ryanchristo (approved):
Looks good to me. Nice work.

x/ecocredit/server/marketplace/prune.go
```diff
@@ -24,16 +24,26 @@ func (k Keeper) PruneSellOrders(ctx context.Context) error {
...
-		if err = k.unescrowCredits(ctx, sellOrder.Seller, sellOrder.BatchId, sellOrder.Quantity); err != nil {
+		if sellOrder.Expiration != nil {
+			if err = k.unescrowCredits(ctx, sellOrder.Seller, sellOrder.BatchId, sellOrder.Quantity); err != nil {
+				return err
+			}
+			expiredOrderIds = append(expiredOrderIds, sellOrder.Id)
+		}
+	}
+	for _, id := range expiredOrderIds {
+		if err := k.stateStore.SellOrderTable().Delete(ctx, &marketplaceapi.SellOrder{Id: id}); err != nil {
```
ryanchristo, line 42: Actually, it looks like we can call this within the `it.Next()` loop rather than appending to `expiredOrderIds`. No need to create `expiredOrderIds` and append the expired orders and then loop through a second time.

technicallyty: per https://github.com/cosmos/cosmos-sdk/blob/main/orm/model/ormtable/iterator.go#L15-L19 we should not mutate the table during iteration. 

however, i came back to this and actually might have a better solution 😈 . so after talking with aaron, nil timestamps are actually treated as zero values, or `timestamppb.Timestamp{Seconds:0, Nanos: 0}`. so if we want the performance boost of `DeleteRange`, we could just delete everything starting at `timestamppb.Timestamp{Seconds:0, Nanos: 1}`. 

the tests still hold up with this [change](https://github.com/regen-network/regen-ledger/pull/1115/commits/f60f734d5843e3f7224ba960390eefdebc55229a) 😄 . lmk if that sounds sane or not

aaronc:
Seeing this I really think this should be fixed in the orm because otherwise it's unexpected behavior

x/ecocredit/server/marketplace/prune.go
```diff
@@ -17,13 +17,15 @@ import (
 func (k Keeper) PruneSellOrders(ctx context.Context) error {
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
 
-	min, blockTime := timestamppb.New(time.Unix(0, 0)), timestamppb.New(sdkCtx.BlockTime())
+	// we set the min to 1 ns because nil expirations are encoded as the 0 value timestamp,
+	// and we DO NOT want those to be deleted/unescrowed.
```
ryanchristo, line 21:
```suggestion
	// we set the min to 1 ns because nil expirations are encoded as the 0 value timestamp,
	// and we DO NOT want those to be deleted/unescrowed.
	// https://github.com/cosmos/cosmos-sdk/issues/11980
```

x/ecocredit/server/marketplace/prune_test.go
```diff
@@ -66,3 +66,41 @@ func TestSell_Prune(t *testing.T) {
...
+// TestPrune_NilExpiration tests that sell orders with nil expirations are not deleted when PruneOrders is called.
+func TestPrune_NilExpiration(t *testing.T) {
+	t.Parallel()
+	s := setupBase(t)
+	s.testSellSetup(batchDenom, ask.Denom, ask.Denom[1:], "C01", start, end, creditType)
+
+	blockTime, err := time.Parse("2006-01-02", "2020-01-01")
+	assert.NilError(t, err)
+	expired, err := time.Parse("2006-01-02", "2010-01-01")
+	assert.NilError(t, err)
```
ryanchristo, line 79: We recently added `types.ParseDate` and should probably make use of it here. No need to do in the above test as it looks like it will get updated in https://github.com/regen-network/regen-ledger/pull/1118.

```suggestion
	blockTime, err := types.ParseDate("block time", "2020-01-01")
	assert.NilError(t, err)
	expired, err := time.Parse("expiration", "2010-01-01")
	assert.NilError(t, err)
```

ryanchristo (approved):
Updates look good to me. Just a couple nits.

aleem1314 (approved):

technicallyty:
going to convert to draft here pending https://github.com/regen-network/regen-ledger/pull/1115#issuecomment-1133512456, we might not need this

Issue #980, opened 2022-04-04
Upgrade handlers and manual state operations


ryanchristo:
## Summary

In order to support migrations for both mainnet and redwood, we will need two upgrade handlers, one for each chain. Both upgrade handlers will run the same state migrations. Following the state migrations, we will need to run manual operations to patch items specific to each chain. Within each upgrade handler, we will also need to add any new parameters.

### Manual Operations

**Mainnet**

- update community member accounts ([governance forum discussion](https://forum.regen.network/discussion/3540-in-need-of-communities-assistance-tokens-lost-in-wallet-conversion))
- add reference id to existing projects

**Mainnet and Redwood**

- update curator address for baskets
- update issuance date for credit batches

### Note

Following the upgrade to Regen Ledger v4.0, we may need to perform the operations for projects, baskets, and credit batches again for items created between the time of the release and the upgrade.

## Updated Summary

A single upgrade handler has been implemented (#1121) and checks are in place for `regen-1` and `regen-redwood-1` whereby we recover funds when chain id is `regen-1` and run state migrations with manual operations specific to each chain performed at the end of the state migrations when chain id is either `regen-1` or `regen-redwood-1`.

### Followups

The current implementation is a working implementation and will be included in the first beta release to undergo thorough testing with temporary test networks that make use of existing state. We may or may not need to address the following items:

**Mainnet**

- [ ] ~Update curators for baskets if any baskets are created between now and the official release.~
- [x] Update issuance dates for batches if any new batches are created between now and the official release.
- [ ] ~Update reference ids for projects if any new batches are created between now and the official release and/or if the reference ids provided need to be modified in any way (see https://github.com/regen-network/regen-ledger/pull/1121#discussion_r888410762)~
- [ ] ~It would be easier to recover funds with a new address that is not associated with an existing account (i.e. an account that does not have a token balance and does not yet exist on chain). This would allow creating a new vesting account that matches the previous account as opposed to getting the existing balance of a basic account, deleting the account, creating a new vesting account, and adding the funds to the new vesting account (see https://github.com/regen-network/regen-ledger/pull/1121#discussion_r888786845)~

**Redwood**
- [x] Update curators for baskets if any new baskets are created between now and the official release.
- [x] Update issuance dates for batches if new batches are created between now and the official release.
- [ ] ~Update reference ids for projects if any new batches are created between now and the official release (we currently do not set any reference ids for projects created on redwood and whether we want to set any is still an open question)~

____

#### For Admin Use

[...]

PR #1121, opened 2022-05-23
feat(app): add upgrade handlers

aleem1314:
## Description

Ref: #980 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/migrations/v3/core.go
```diff
@@ -315,6 +318,96 @@ func MigrateState(sdkCtx sdk.Context, storeKey storetypes.StoreKey,
 		return err
 	}
 
+	// - update curator address for baskets
+	// we don't have baskets on mainnet
+
```
aleem1314, line 323: This PR only contains mainnet migration. We cannot add both mainnet & Redwood migrations here.

ryanchristo: So the idea was that the manual operations specific to each chain (mainnet and redwood) would be run in the upgrade handler and the state migrations would work for both. In the same way that `recoverFunds` is included in the upgrade handler and run separately from the state migrations, the manual operations specific to each chain would also be included in the upgrade handler and no hard-coded values specific to mainnet would be included in the state migrations.

We would then have two upgrade handlers that each perform manual operations following state migrations, one would have the upgrade name `v4.0.0` (mainnet) and one would have the upgrade name `v4.0.0-redwood`. We can hold off on implementing the upgrade handler for redwood in this pull request if you prefer but we should make sure state migrations will run for both mainnet and redwood and the manual operations for mainnet are separate from the state migrations.

ryanchristo: So the idea was that the manual operations specific to each chain (mainnet and redwood) would be run in the upgrade handler and the state migrations would work for both. In the same way that `recoverFunds` is included in the upgrade handler and run separately from the state migrations, the manual operations specific to each chain would also be included in the upgrade handler and no hard-coded values specific to mainnet would be included in the state migrations.

We would then have two upgrade handlers that each perform manual operations following state migrations, one would have the upgrade name `v4.0.0` (mainnet) and one would have the upgrade name `v4.0.0-redwood`. We can hold off on implementing the upgrade handler for redwood in this pull request but we should make sure state migrations will run for both mainnet and redwood and the manual operations for mainnet are separate from the state migrations.

aleem1314: > In the same way that `recoverFunds` is included in the upgrade handler and run separately from the state migrations, the manual operations specific to each chain should be included in upgrade handlers specific to each chain and no hard-coded values specific to one chain and not the other would be included in the state migrations.
> 
I tried a similar approach, I don't think we can access ecocredit state outside. We are using new app wiring for regen modules.

ryanchristo: Oof. Yea, that sounds about right. I'll try to take a look. cc @clevinson @aaronc

ryanchristo: We could expose the keeper methods we need and make use of them in the upgrade handler in the same way that we expose `AddCreditType` and `AllowDenom` for custom governance proposals and `PruneOrders` for begin blocker...

ryanchristo: Not sure that is the right solution though. Let's discuss on our extended standup.

ryanchristo: We could create and expose a keeper method that would handle the manual operations in the same way that we expose `AddCreditType` and `AllowDenom` for custom governance proposals and `PruneOrders` for begin blocker...

Not sure that is the right solution though. Let's discuss on our extended standup.

ryanchristo: Ok, I think it makes the most sense to perform the manual operations within the state migrations. We can perform specific operations for each chain by pulling the chain id from sdk context.

It would be nice to separate out the manual operations from the current state migrations and to test them separately. We could call a single function at the end of  `MigrateState` that we define in a separate file that checks the chain id and calls a specific function for each chain id (`regen-1` and `regen-redwood-1`) that performs the manual operations.

ryanchristo: Ok, I think it makes the most sense to perform the manual operations within the state migrations. We can perform specific operations for each chain by pulling the chain id from sdk context.

It would be nice to separate out the manual operations from the current state migrations and to test them separately. We could call a single function at the end of  `MigrateState` that we define in a separate file that checks the chain id and then calls a specific function for each chain id (`regen-1` and `regen-redwood-1`).

Does that sound like a reasonable solution? I think this is actually an improvement from the original idea @clevinson and I were thinking with separate upgrade handlers. Not sure why we didn't think of it then.

ryanchristo: Maybe `patch.go` within `migrations/v3` and something like `patchMigration` that would then call either `patchMainnet` or `patchRedwood` based on chain id? If this seems unreasonable, we can further discuss on our extended standup.

ryanchristo: Maybe `patch.go` within `migrations/v3` and `patchMigration` that would call either `patchMainnet` or `patchRedwood` based on chain id? If this seems unreasonable, we can discuss on our extended standup.

ryanchristo: Maybe `patch.go` within `migrations/v3` and `patchMigration` that then calls `patchMainnet` or `patchRedwood` based on chain id? If this seems unreasonable, we can further discuss on our extended standup.

x/ecocredit/migrations/v3/patch.go
```diff
@@ -0,0 +1,188 @@
...
+	// project location -> reference-id
+	// FR              -> "" // TODO: add reference-id
+	// US              -> ""
+	// AU-NSW 2453     -> ""
+	// K               -> ""
+	// US-FL 9876      -> ""
+
+	locationToReferenceIdMap := make(map[string]string)
+	locationToReferenceIdMap["CD-MN"] = ""
+	locationToReferenceIdMap["KE"] = ""
```
aleem1314, line 40:
```suggestion
	locationToReferenceIdMap["FR"] = ""
	locationToReferenceIdMap["US"] = ""
	locationToReferenceIdMap["AU-NSW 2453"] = ""
	locationToReferenceIdMap["K"] = ""
	locationToReferenceIdMap["US-FL 9876"] = ""
```

x/ecocredit/migrations/v3/patch.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func patchMainnet(ctx context.Context, ss api.StateStore, oldBatchDenomToNewDenomMap map[string]string) error {
+	// project location -> reference-id
+	// CD-MN -> "" // TODO: add reference-id
+	// KE    -> ""
+
+	locationToReferenceIdMap := make(map[string]string)
+	locationToReferenceIdMap["CD-MN"] = ""
+	locationToReferenceIdMap["KE"] = ""
```
ryanchristo, line 122: Same as above, we should probably use `projectIdToReferenceId` here.

x/ecocredit/migrations/v3/patch.go
```diff
@@ -0,0 +1,191 @@
...
+		basket.Curator = sdk.AccAddress(curator)
+		if err := basketStore.BasketTable().Update(ctx, basket); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func patchMainnet(ctx context.Context, ss api.StateStore, oldBatchDenomToNewDenomMap map[string]string) error {
```
ryanchristo, line 115: Might be nice to move this between `pathMigrations` and `patchRedwood` so that the order is:

- `patchMigrations`
- `patchMainnet`
- `patchRedwwod`

And then then helper functions below.

ryanchristo (commented):
This is looking good. A few requested changes. I think separating the tests for manual operations would be nice so that we are not only testing mainnet and the tests in `core_test.go` would test a scenario with no matching chain id. Would be easier to add cases for manual operations that are not being tested at the moment and to do so specific to each chain.

x/ecocredit/migrations/v3/patch.go
```diff
@@ -0,0 +1,191 @@
...
+// - update curator address for baskets
+func patchMigrate(ctx context.Context, sdkCtx sdk.Context, ss api.StateStore,
+	basketStore basketapi.StateStore, oldBatchDenomToNewDenomMap map[string]string) error {
+	if sdkCtx.ChainID() == "regen-1" {
+		return patchMainnet(ctx, ss, oldBatchDenomToNewDenomMap)
+	} else if sdkCtx.ChainID() == "regen-redwood-1" {
+		return patchRedwood(ctx, ss, basketStore, oldBatchDenomToNewDenomMap)
+	} else {
+		return fmt.Errorf("invalid chain-id %s", sdkCtx.ChainID())
+	}
```
ryanchristo, line 26:
```suggestion
	}
```

anilCSE (commented):
why don't we take a complete new account for recovering funds? In that way, we can just replace old account address, pubkey with new one in the account state and bank state

app/stable_appconfig.go
```diff
@@ -59,7 +68,102 @@ func setCustomOrderEndBlocker() []string {
...
+
+		// run x/ecocredit state migrations
+		if err := app.smm.RunMigrations(ctx, app.AppCodec()); err != nil {
+			return nil, err
+		}
+		toVersion[ecocredit.ModuleName] = ecocreditmodule.Module{}.ConsensusVersion()
+
+		if err := recoverFunds(ctx, app.AccountKeeper, app.BankKeeper); err != nil {
+			return nil, err
+		}
```
ryanchristo, line 90: Separate upgrade handlers are no longer needed if we are performing manual operations in the state migrations based on chain id. We should check chain id here and run recover funds only for chain id `regen-1` and remove the separate upgrade handler for redwood so that this upgrade handler works for `regen-1`, `regen-redwood-1`, and for any other test networks with a separate chain id. Apologies for the confusion.

```suggestion
		// recover funds for community member (regen-1 governance proposal #11)
		if ctx.ChainID() == "regen-1" {
			if err := recoverFunds(ctx, app.AccountKeeper, app.BankKeeper); err != nil {
				return nil, err
			}
		}
```

app/stable_appconfig.go
```diff
@@ -59,7 +68,102 @@ func setCustomOrderEndBlocker() []string {
...
+
+		return toVersion, nil
+	})
+
+	// redwood upgrade handler
+	const redwoodUpgradeName = "v4.0.0-redwood"
+	app.UpgradeKeeper.SetUpgradeHandler(upgradeName, func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
```
ryanchristo, line 97: As mentioned above, we should move forward with a single upgrade handler for all networks regardless of chain id and only run certain operations when chain id is either `regen-1` or `regen-redwood-1`. This upgrade handler can be removed with the changes requested above. Apologies for the confusion.

app/stable_appconfig.go
```diff
@@ -59,7 +68,102 @@ func setCustomOrderEndBlocker() []string {
...
+
+		return toVersion, nil
+	})
+}
+
+func recoverFunds(ctx sdk.Context, ak authkeeper.AccountKeeper, bk bankkeeper.Keeper) error {
+	// address with funds inaccessible
+	const lostAddr = "regen1c3lpjaq0ytdtsrnjqzmtj3hceavl8fe2vtkj7f"
+	// address that the community member has access to
+	const newAddr = "regen14tpuqrwf95evu3ejm9z7dn20ttcyzqy3jjpfv4"
```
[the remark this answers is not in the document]
ryanchristo: cc @clevinson @dpdanpittman

x/ecocredit/migrations/v3/patch.go
```diff
@@ -0,0 +1,190 @@
...
+		return patchRedwood(ctx, ss, basketStore, oldBatchDenomToNewDenomMap)
+	}
+
+	return nil
+}
+
+func patchMainnet(ctx context.Context, ss api.StateStore, oldBatchDenomToNewDenomMap map[string]string) error {
+	// project location -> reference-id
+	// CD-MN -> "" // TODO: add reference-id
+	// KE    -> ""
```
ryanchristo, line 31: @clevinson and I discussed reference id with @S4mmyb for the projects that be created but no reference id was decided upon. We can add later when we are closer to a final release if this is still undecided.

x/ecocredit/migrations/v3/patch.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func patchRedwood(ctx context.Context, ss api.StateStore,
+	basketStore basketapi.StateStore, oldBatchDenomToNewDenomMap map[string]string) error {
+	// project location -> reference-id
+	// FR              -> "" // TODO: add reference-id
+	// US              -> ""
+	// AU-NSW 2453     -> ""
+	// K               -> ""
+	// US-FL 9876      -> ""
```
ryanchristo, line 70: Same as above. Will check with @S4mmyb on reference id for projects created on redwood.

ryanchristo (commented):
This looks great. I was think that the tests in `patch_test.go` wouldn't need to set up and run the state migrations (not including the manual operations) and then they would only include testing the manual operations but how you set them up works too.

Items remaining in this pull request:

- consolidate into one upgrade handler

Items remaining that we can address in a followup pull request:

- update community member address (we need an address that does not yet exist on chain)
- add reference ids and update reference ids and issuance dates for any new batches created

app/stable_appconfig.go
```diff
@@ -59,7 +68,105 @@ func setCustomOrderEndBlocker() []string {
...
+		}
+
+		// run x/ecocredit state migrations
+		if err := app.smm.RunMigrations(ctx, app.AppCodec()); err != nil {
+			return nil, err
+		}
+		toVersion[ecocredit.ModuleName] = ecocreditmodule.Module{}.ConsensusVersion()
+
+		return toVersion, nil
+	})
```
aleem1314, line 114:
```suggestion
```

app/stable_appconfig.go
```diff
@@ -59,7 +68,88 @@ func setCustomOrderEndBlocker() []string {
...
+		}
+
+		// run x/ecocredit state migrations
+		if err := app.smm.RunMigrations(ctx, app.AppCodec()); err != nil {
+			return nil, err
+		}
+		toVersion[ecocredit.ModuleName] = ecocreditmodule.Module{}.ConsensusVersion()
+
+		// recover funds for community member (regen-1 governance proposal #11)
+		if ctx.ChainID() == "regen-1" {
```
anilCSE, line 89: do we really need this check here? What if we want to run a testnet off-of mainnet candidate and test this

ryanchristo: Without the check, we would run recover funds on redwood. If you want to test on a testnet, you would need to update here and the check in `migrations/v3/patch.go`. We don't want to run recover funds if we don't need to.

ryanchristo (approved):
Successfully tested software upgrade *without* chain id `regen-1` or `regen-redwood-1`. Will need to do more thorough testing using existing mainnet state and redwood state but this looks good to me and the remaining items can be addressed in a followup.

Also, one issue came up but unrelated to the changes here, which is `issuer` was not populated in the new credit batch following migrations (opened #1138).

x/ecocredit/migrations/v3/patch_test.go
```diff
@@ -0,0 +1,846 @@
...
+}
+
+func assetBatchIssuanceDate(t *testing.T, ctx context.Context, ss api.StateStore, denom, exp string) {
+	res, err := ss.BatchTable().GetByDenom(ctx, denom)
+	require.NoError(t, err)
+	parsed, _ := time.Parse(time.RFC3339, exp)
+	require.Equal(t, res.IssuanceDate.AsTime(), parsed)
+}
+
+func assetProjectReferenceId(t *testing.T, ctx context.Context, ss api.StateStore, id, jurisdiction, referenceID string) {
```
clevinson, line 841:
```suggestion
func assertProjectReferenceId(t *testing.T, ctx context.Context, ss api.StateStore, id, jurisdiction, referenceID string) {
```

x/ecocredit/migrations/v3/patch_test.go
```diff
@@ -0,0 +1,846 @@
...
+	//  C01-20210909-20220101-003  ->  "2022-03-08T17:18:19Z"
+	assetBatchIssuanceDate(t, ctx, ss, "C01-001-20170101-20180101-001", "2022-03-30T07:46:01Z")
+	assetBatchIssuanceDate(t, ctx, ss, "C01-001-20170101-20180101-002", "2022-03-30T07:51:12Z")
+	assetBatchIssuanceDate(t, ctx, ss, "C01-001-20170101-20180101-003", "2022-03-30T15:04:30Z")
+	assetBatchIssuanceDate(t, ctx, ss, "C01-002-20180101-20200101-001", "2022-02-09T09:10:02Z")
+	assetBatchIssuanceDate(t, ctx, ss, "C01-003-20180909-20200101-001", "2022-03-08T17:25:23Z")
+	assetBatchIssuanceDate(t, ctx, ss, "C01-004-20190101-20191231-001", "2022-05-11T11:35:30Z")
+	assetBatchIssuanceDate(t, ctx, ss, "C01-001-20190101-20210101-004", "2022-02-14T09:07:25Z")
+	assetBatchIssuanceDate(t, ctx, ss, "C01-005-20190101-20210101-001", "2022-04-22T16:32:09Z")
+	assetBatchIssuanceDate(t, ctx, ss, "C01-001-20210909-20220101-005", "2022-03-08T17:18:19Z")
```
clevinson, line 808:
```suggestion
	assertProjectReferenceId(t, ctx, ss, "C01-001", "US", "")
	assertProjectReferenceId(t, ctx, ss, "C01-002", "AU-NSW 2453", "")
	assertProjectReferenceId(t, ctx, ss, "C01-003", "FR", "")
	assertProjectReferenceId(t, ctx, ss, "C01-004", "KE", "")
	assertProjectReferenceId(t, ctx, ss, "C01-005", "US-FL 98765", "")
	assertProjectReferenceId(t, ctx, ss, "C02-001", "FR", "")
	assertProjectReferenceId(t, ctx, ss, "C02-002", "US", "")
	assertProjectReferenceId(t, ctx, ss, "C04-001", "FR", "")
	assertProjectReferenceId(t, ctx, ss, "C04-002", "US", "")

	// batch issuance dates
	//  C01-20170101-20180101-005  ->  "2022-03-30T07:46:01Z"
	//  C01-20170101-20180101-006  ->  "2022-03-30T07:51:12Z"
	//  C01-20170101-20180101-007  ->  "2022-03-30T15:04:30Z"
	//  C01-20180101-20200101-001  ->  "2022-02-09T09:10:02Z"
	//  C01-20180909-20200101-004  ->  "2022-03-08T17:25:23Z"
	//  C01-20190101-20191231-009  ->  "2022-05-11T11:35:30Z"
	//  C01-20190101-20210101-002  ->  "2022-02-14T09:07:25Z"
	//  C01-20190101-20210101-008  ->  "2022-04-22T16:32:09Z"
	//  C01-20210909-20220101-003  ->  "2022-03-08T17:18:19Z"
	assertBatchIssuanceDate(t, ctx, ss, "C01-001-20170101-20180101-001", "2022-03-30T07:46:01Z")
	assertBatchIssuanceDate(t, ctx, ss, "C01-001-20170101-20180101-002", "2022-03-30T07:51:12Z")
	assertBatchIssuanceDate(t, ctx, ss, "C01-001-20170101-20180101-003", "2022-03-30T15:04:30Z")
	assertBatchIssuanceDate(t, ctx, ss, "C01-002-20180101-20200101-001", "2022-02-09T09:10:02Z")
	assertBatchIssuanceDate(t, ctx, ss, "C01-003-20180909-20200101-001", "2022-03-08T17:25:23Z")
	assertBatchIssuanceDate(t, ctx, ss, "C01-004-20190101-20191231-001", "2022-05-11T11:35:30Z")
	assertBatchIssuanceDate(t, ctx, ss, "C01-001-20190101-20210101-004", "2022-02-14T09:07:25Z")
	assertBatchIssuanceDate(t, ctx, ss, "C01-005-20190101-20210101-001", "2022-04-22T16:32:09Z")
	assertBatchIssuanceDate(t, ctx, ss, "C01-001-20210909-20220101-005", "2022-03-08T17:18:19Z")
```

clevinson (approved):

app/recover_funds_test.go
```diff
@@ -0,0 +1,97 @@
...
+	regenApp.BankKeeper.SendCoinsFromModuleToAccount(ctx, minttypes.ModuleName, newAddr, newAccountBalance)
+
+	ctx = ctx.WithBlockTime(time.Now())
+	err = recoverFunds(ctx, regenApp.AccountKeeper, regenApp.BankKeeper)
+	require.NoError(t, err)
+
+	require.True(t, regenApp.BankKeeper.GetBalance(ctx, lostAddr, "uregen").IsZero())
+
+	newbalance := regenApp.BankKeeper.GetAllBalances(ctx, newAddr)
+	require.Equal(t, newbalance, newAccountBalance.Add(oldAccountBalance...))
```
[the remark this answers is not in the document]
clevinson: OK I think I understand this now... 

So this is in the case where the user has a pre-existing account and we want to add their old balance.  I think actually this works as expected.

x/ecocredit/migrations/v3/patch.go
clevinson: My guess is that we can leave reference ID blank on redwood.

ryanchristo: Sam is checking with the interface team. They may want some of these projects to have a reference id but we can handle in a followup.

app/stable_appconfig.go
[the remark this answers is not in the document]
clevinson: Following up on this... @ryanchristo and I just spent some time reviewing this PR in more detail (in particular the `recoverFunds()` implementation), and I think this may work as desired even if the user has a pre-existing balance.

The error when setting a periodic vesting account on a pre-existing account only happens in the message server for [`CreateVestingAccount()`](https://github.com/cosmos/cosmos-sdk/blob/main/x/auth/vesting/msg_server.go#L52-L54).

So I think the way we are doing it (by ensuring that the vesting acc balance is added to the pre-existing new account balance), and then calling `accountKeeper.SetAccount()` should avoid an issues.

app/stable_appconfig.go
```diff
@@ -59,7 +68,88 @@ func setCustomOrderEndBlocker() []string {
...
+			}
+		} else if addr == newAddr {
+			newAccount = account
+			if lostAccount != nil {
+				return true
+			}
+		}
+
+		return false
+	})
```
anilCSE, line 122: cant we simply use 

```
lostAccout := regenApp.AccountKeeper. GetAccount(ctx, lostAddr)
newAccount := regenApp.AccountKeeper. GetAccount(ctx, lostAddr)
```

Also seems like this doesn't work if `newAccount` is nil

anilCSE, line 122: cant we simply use 

```
lostAccout := regenApp.AccountKeeper. GetAccount(ctx, lostAddr)
newAccount := regenApp.AccountKeeper. GetAccount(ctx, lostAddr)
```

Also seems like this doesn't work if `newAccount` is nil

app/stable_appconfig.go
```diff
@@ -59,7 +68,88 @@ func setCustomOrderEndBlocker() []string {
...
+	spendable := bk.SpendableCoins(ctx, lostAccount.GetAddress())
+	if err := bk.SendCoins(ctx, lostAccount.GetAddress(), newAccount.GetAddress(), spendable); err != nil {
+		return err
+	}
+
+	newPVA := vestingtypes.NewPeriodicVestingAccount(
+		authtypes.NewBaseAccount(newAccount.GetAddress(), newAccount.GetPubKey(), newAccount.GetAccountNumber(), newAccount.GetSequence()),
+		va.OriginalVesting, va.StartTime, vestingPeriods,
+	)
+	ak.SetAccount(ctx, newPVA)
```
anilCSE, line 149: This works all fine but looks complex for me. I was advising to consider a brand new account from user for simplicity. The solution can simply be
```
// read lost-account from auth store
acc := regenApp.AccountKeeper.GetAccount(ctx, lostAddr)
newAcc := acc;
newAcc.Address = newAddr
newAcc.Pubkey = nil

// delete old account
err := regenApp.AccountKeeper. RemoveAccount(ctx, lostAcc)

// set new account
err := regenApp.AccountKeeper.SetAccount(ctx, newAcc)
```

similarly for bank balances too. Not sure how this sounds for others but I am comfortable with this approach.

anilCSE (approved):
lgtm, couple of notes for your consideration.

x/ecocredit/migrations/v3/patch.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func patchMainnet(ctx context.Context, ss api.StateStore, oldBatchDenomToNewDenomMap map[string]string) error {
+	// project location -> reference-id
+	// CD-MN -> "" // TODO: add reference-id
+	// KE    -> ""
+
+	locationToReferenceIdMap := make(map[string]string)
+	locationToReferenceIdMap["CD-MN"] = ""
+	locationToReferenceIdMap["KE"] = ""
```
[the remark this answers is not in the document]
ryanchristo: Committing this change and merging this pull request so that we can tag a beta and test. If we need to make any revisions, we can do so at a later point in time. Updated the issue and added followups: https://github.com/regen-network/regen-ledger/issues/980

x/ecocredit/migrations/v3/patch.go
ryanchristo: See https://github.com/regen-network/regen-ledger/pull/1121#discussion_r888410762

app/stable_appconfig.go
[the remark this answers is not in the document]
ryanchristo: See https://github.com/regen-network/regen-ledger/pull/1121#discussion_r888786845

ryanchristo:
Very nice work @aleem1314 :rocket:

x/ecocredit/migrations/v3/patch.go
S4mmyb: We can move forward without including any sort of reference id for projects on redwood @ryanchristo

Issue #1111, opened 2022-05-18
Restrictions on defining a data resolver URL


ryanchristo:
## Summary

When defining a resolver in the data module, there are currently no restrictions on the provided URL. Anyone can define a resolver with any URL and essentially domain squat a URL that they do not administer.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ryanchristo:
cc @clevinson @aaronc

I think the simplest solution would be remove unique constraints on URL. We currently look up a resolver based on resolver id and we could allow multiple resolvers with the same URL.

It would be nice to verify a TXT record but not sure how we could do that deterministically. Other options would include a fee (does not solve the problem completely) or governance (adds a lot of overhead and additional work).

clevinson:
Makes sense to me. Also in favor of removing uniqueness constraint.

ryanchristo:
> We currently look up a resolver based on resolver id and we could allow multiple resolvers with the same URL.

We actually look up a resolver by url in the current implementation but we can switch to id. I was thinking of registering data to a resolver, which currently takes the resolver id and not the url and which doesn't need to change.

The id is returned when defining a resolver, so the manager will need to store it for use when registering data and for any future operations we add related to the resolver (e.g. updating a resolver manager).

When querying for resolvers by IRI or content hash, we return the resolver urls rather than the resolver ids. This still makes sense although we may want to consider adding a query for resolvers by url (multiple resolvers returned if multiple resolvers with the same url exist), which would allow a user to look up a resolver without the id. Might be nice to have.

Opened #1128 to remove the unique constraint and to update the query by resolver url.

clevinson:
I agree that allowing for users to query for resolvers by URL (and therefore find the data that is registered in that resolver) would be valuable.  In support of adding this as a nice to have.

PR #1128, opened 2022-05-25
refactor(x/data): remove unique constraint on resolver url

ryanchristo:
## Description

Closes: #1111

This pull request removes the unique constraint on url and updates query resolver to take the resolver id rather than url.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


aleem1314 (approved):
lgtm

x/data/client/testsuite/grpc.go
```diff
@@ -474,22 +472,22 @@ func (s *IntegrationTestSuite) TestQueryResolverInfo() {
...
 			"not found",
 			0,
 		},
 		{
 			"valid request",
-			fmt.Sprintf("%s/regen/data/v1/resolver?url=%s", val.APIAddress, url),
+			fmt.Sprintf("%s/regen/data/v1/resolver?id=%d", val.APIAddress, s.resolverID),
 			false,
 			"",
 			2,
```
clevinson, line 486: Doesn't look like `expItems` is ever checked. Should we remove it from the table?

Also - I don't think we're actually able to query resolver with pagination, so maybe we should drop this last test case entirely?

x/data/server/features/define_resolver.feature
```diff
@@ -1,15 +1,12 @@
 Feature: Define Resolver
 
-  Background:
-    Given the resolver url "https://foo.bar"
+  Rule: the resolver is defined with no unique constraints on the URL
 
-  Rule: the resolver is defined if the resolver url is unique
+    Scenario: the url is unique
```
clevinson, line 5:
```suggestion
    Scenario: a resolver for a new URL is registered
```

x/data/server/features/define_resolver.feature
```diff
@@ -1,15 +1,12 @@
...
 
-    Scenario: the resolver has not been defined
-      When alice attempts to define the resolver
-      Then the resolver exists and alice is the manager
-
-    Scenario: the resolver has already been defined
-      Given alice has defined the resolver
-      When alice attempts to define the resolver
-      Then expect the error "resolver URL already exists"
+    Scenario: the url is not unique
```
clevinson, line 9:
```suggestion
    Scenario: A resolver is registered with a URL that is already registered
```

clevinson (approved):
🙏 thank u @ryanchristo !

Just dropped at note in https://github.com/regen-network/regen-ledger/issues/1111#issuecomment-1137861872 echo'ing that I think querying for resolvers by URL with pagination (returning all resolvers with their ID & manager) would be useful.  Potentially in a follow-up.

x/data/server/features/define_resolver.feature
ryanchristo: In general, I was thinking the scenario would be short and to the point and the rule would be more descriptive. Given the rule is "the resolver is defined with no unique constraints on the URL", the two scenarios are "the url is unique" and "the url is not unique". This is also how the BDD books write scenarios. For example,

```
Rule: Any visitor to the website can place a customer-collection order

  Scenario: Unauthenticated customer chooses to collect order
    Given the customer is unauthenticated
    When they choose to collect their order
    Then they should be asked to supply contact details

  Scenario: Authenticated customer chooses to collect order
    Given the customer is authenticated
    When they choose to collect their order
    Then they should be asked to confirm contact details
```

ryanchristo: This also produces the test logs:

```
=== RUN   TestDefineResolver
=== RUN   TestDefineResolver/Define_Resolver
=== RUN   TestDefineResolver/Define_Resolver/a_resolver_for_a_new_url_is_registered
=== RUN   TestDefineResolver/Define_Resolver/a_resolver_is_registered_with_a_url_that_is_already_registered
```

Versus the current test logs:

```
=== RUN   TestDefineResolver
=== RUN   TestDefineResolver/Define_Resolver
=== RUN   TestDefineResolver/Define_Resolver/the_url_is_unique
=== RUN   TestDefineResolver/Define_Resolver/the_url_is_not_unique
```

ryanchristo: We are also "defining" a resolver here, not "registering" data to the resolver.

x/data/client/testsuite/grpc.go
ryanchristo: Good catch. Looks like some old copy pasta that went unnoticed.

ryanchristo: https://github.com/regen-network/regen-ledger/pull/1128/commits/56f1b975eeab03dc2338fe4a9b3de0e53fa24c0f

ryanchristo:
@clevinson Would be happy to further discuss best format for scenarios and update throughout but the current format is aligned with how I've been writing scenarios elsewhere and I think that if we include a descriptive rule, scenarios should be concise and provide only the necessary information, which makes reading through the test logs easier when debugging failing scenarios.

Issue #1080, opened 2022-05-03
Inconsistent address fields in ecocredit proto definitions


ryanchristo:
## Summary

Audit address field names and make sure we are using consistent names. For example:

- In `MsgSell`, `MsgUpdateSellOrders`, `EventUpdateSellOrders`, and `EventSell` we use `owner`
- In `MsgCancelSellOrder`, `SellOrder` we use `seller`

## Problem Definition

Using different field names to reference the same address is confusing.

## Proposal

- Update all instances of `owner` within the context of sell order to `seller` (or vice versa)
- Audit ecocredit proto definitions for address field names and make other changes if necessary

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1153, opened 2022-06-02
refactor(x/ecocredit): consistent address field names

ryanchristo:
## Description

Closes: #1080

This pull requests makes the following changes to field names:

- All previous references to `owner` in relation to sell orders have been renamed to `seller` (previously we were using `owner` and `seller` but we are using `buyer` for buy orders and `owner` for credits)
- All previous references to `account` for credit balances have been renamed to `address` (previously we were using `account` and `address` but `address` is more in line with the sdk bank module)
- All previous references of `holder` in relation to credits being cancelled, retired, and bridged have been renamed to `owner` (previously we were using `holder` and `owner`)

This pull request also updates `SellOrdersByAddress` to `SellOrdersBySeller`.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


clevinson:
Generally giving a concept ACK on this. I do think it might be better to try and stay consistent with bank module usage of address where appropriate.

So I think:
- QueryBalanceReuqest
- QueryBalancesRequest
- BatchBalanceInfo
- BatchBalance

Might be better to keep as "address" instead of "owner".

proto/regen/ecocredit/marketplace/v1/state.proto
```diff
@@ -20,7 +20,7 @@ message SellOrder {
   // id is the unique ID of sell order.
   uint64 id = 1;
 
-  // seller is the bytes address of the owner of the credits being sold.
+  // seller is the address of the account that is selling credits.
```
clevinson, line 23:
```suggestion
  // seller is the address (in bytes) of the account that is selling credits.
```
maybe it still makes sense to explicitly state that its in byte representation?

clevinson (approved):

technicallyty (approved):
👍🏻

proto/regen/ecocredit/marketplace/v1/state.proto
ryanchristo: I don't think it's necessary. I previously made a change to add something like "unique string identifier" in relation to credit class IDs and it was pointed out that this is excessive because the field clearly states the type. I would have agreed before but I don't think we need to state the type unless we are adding details such as "any arbitrary string".

PR #1160, opened 2022-06-07
chore(ecocredit/core): keeper method audit

technicallyty:
## Description

- refactors MsgCancel and MsgBridge by taking out the embedded credits, moving to it's own proto struct. This was done to avoid an unnecessary loop over credits just to change their type in the `Bridge` keeper method.
- updated errors where user provided values could return a `not found` error.
- simplifies `timestamppb` conversions in `CreateBatch`
- updates `assertClassIssuer` to use naming of `key` instead of `id`.
- removes instances where `GetCreditTypeFromBatchDenom` could just be `CreditTypeTable.Get`

ref: #715 

Closes: #1171 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/server/core/create_batch.go
```diff
@@ -88,8 +88,8 @@ func (k Keeper) CreateBatch(ctx context.Context, req *core.MsgCreateBatch) (*cor
 		recipient, _ := sdk.AccAddressFromBech32(issuance.Recipient)
 
 		// get the current batch balance of the recipient account
-		// Note: Get because batch balance may or may not already exist
-		// depending on the length of issuance and if recipient is the same
+		// Note: Issuance could be for an account with no prior balance,
+		// 		 so we must catch the not found case and apply a zero value balance.
```
ryanchristo, line 92:
```suggestion
		// Note: Issuance could be for an account with no prior balance,
		// so we must catch the not found case and apply a zero value balance.
```

ryanchristo (approved):
Looks good to me.

I was not sure at first on whether or not we should define `Credits` separately or define within each message but if for some reason we need to change `Credits` specifically for one message, we could theoretically redefine `Credits` as needed specifically for that message and avoid breaking changes to the other uses of `Credits`.

I also don't think this would be a likely scenario and if we were to make any changes to `Credits`, we would likely need to make the same changes to the other uses of `Credits`. I think we could maybe do something similar for `BatchIssuance`, which I can do in #1161.

ryanchristo:
CI is breaking. Looks like unexpected error messages.

technicallyty:
looks like there was an odd edge case in `NewDecFromString` caught by the CI.. if you pass any form of "NaN" ("NAn", "nAn", etc), it actually doesn't error out 😮

ryanchristo (approved):

ryanchristo:
@aleem1314 can you review when you get a chance? :pray:

ryanchristo:
> looks like there was an odd edge case in NewDecFromString caught by the CI.. if you pass any form of "NaN" ("NAn", "nAn", etc), it actually doesn't error, but the test expects one open_mouth

Opened #1181

aleem1314 (approved):

ryanchristo:
cc @technicallyty can you resolve conflicts? :pray:

x/ecocredit/server/core/create_batch.go
```diff
@@ -51,7 +51,7 @@ func (k Keeper) CreateBatch(ctx context.Context, req *core.MsgCreateBatch) (*cor
 		return nil, err
 	}
 
-	startDate, endDate := timestamppb.New(req.StartDate.UTC()), timestamppb.New(req.EndDate.UTC())
+	startDate, endDate := timestamppb.New(*req.StartDate), timestamppb.New(*req.EndDate)
```
ryanchristo, line 54: I missed this the first time around... I don't think we want to store local times if the user provides a local time. I think we always want to store the UTC time.

ryanchristo: cc @clevinson

technicallyty: whoops, nice catch - fixed here https://github.com/regen-network/regen-ledger/pull/1160/commits/524e2f5cf0b7f930f05e299575a778ff7a1769d5

PR #1174, opened 2022-06-08
feat(ecocredit): BridgeReceive

technicallyty:
## Description

- adds a new msg & method `BridgeReceive`
- this method facilitates the bridging of credits by minting into existing bridged batches if they exist, or creating a new project and issuing a batch to it.

Closes: n/a

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/server/core/bridge_receive.go
```diff
@@ -0,0 +1,75 @@
...
+		})
+		if err != nil {
+			return nil, err
+		}
+		return &core.MsgBridgeReceiveResponse{BatchDenom: batchRes.BatchDenom}, nil
+	}
+	// otherwise, we can simply mint into the batch
+	_, err = k.MintBatchCredits(ctx, &core.MsgMintBatchCredits{
+		Issuer:     req.ServiceAddress,
+		BatchDenom: "", // TODO(Tyler): does this come from the bridge? do we form it somehow?
```
dave, line 64: I think we're missing need logic to determine if there is a existing batch or not...

e.g. 

```
select project

if no project {
    MsgCreateProject
}

select batch

if no batch {
    MsgCreateBatch
} else {
    MsgMintBatchCredits
}

```

proto/regen/ecocredit/v1/state.proto
```diff
@@ -83,6 +83,7 @@ message Project {
     index : {id : 2, fields : "class_key,id", unique : true}
     index : {id : 3, fields : "admin"}
     index : {id : 4, fields : "reference_id"}
+    index : {id : 5, fields : "admin,reference_id"}
```
technicallyty, line 86: added this index to narrow down how many projects we have to sift through.

x/ecocredit/server/core/bridge_receive.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+		return &core.MsgBridgeReceiveResponse{BatchDenom: batchRes.BatchDenom, ProjectId: projectRes.ProjectId}, nil
+	}
+
+	// TODO(Tyler): not sure if this is the best way to match a batch.
+	// TODO(Tyler): multiple batches could have same start/end in a project..
+	project := projects[0]
+	// idx will be of form C01-001-20210107-" catching all batches with that project Id and start date.
+	batchIdx := fmt.Sprintf("%s-%s-", project.Id, req.StartDate.Format("20060102"))
+	bIt, err := k.stateStore.BatchTable().List(ctx, api.BatchDenomIndexKey{}.WithDenom(batchIdx))
```
technicallyty, line 79: this doesn't seem too ideal..need a better way to get the batch the bridged credit belongs to

x/ecocredit/server/core/bridge_receive.go
technicallyty: updated work in https://github.com/regen-network/regen-ledger/pull/1174/commits/f75d1403b91e4e024a8c28b9dcba1bd09978497a

x/ecocredit/server/core/bridge_receive.go
technicallyty: during Cory and I's 1;1 sync he suggested in the case we have multiple batches (i think a rare occurence?) we could have this method select which batch to mint into by oldest issuance date.

thoughts? @dave @ryanchristo @aaronc

x/ecocredit/server/core/bridge_receive.go
```diff
@@ -89,7 +89,8 @@ func (k Keeper) BridgeReceive(ctx context.Context, req *core.MsgBridgeReceive) (
 		if err != nil {
 			return nil, err
 		}
-		if batch.Metadata == req.BatchMetadata {
+		// the timestamp stored in the batch is more granular than the date in the denom representation, so we match here.
+		if batch.StartDate.AsTime().Equal(*req.StartDate) && batch.EndDate.AsTime().Equal(*req.EndDate) && batch.Metadata == req.BatchMetadata {
```
dave, line 93: @technicallyty The current metadata converter rounds to the nearest day, so if the YYYYMMDD version matches then this will match. 

Why do we need to compare the metadata? If we're already selecting the batch based on project ID and date range, then I don't believe we need to match on metadata?

technicallyty: hmm I did this with the idea that there might be multiple batches with same start/end dates within a project but perhaps that is not possible 🧐

ryanchristo: I think it's unlikely but possible and the extra check would be good to have. :+1:

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -435,5 +439,61 @@ message MsgBridge {
...
+  string project_metadata = 9;
+
+  // batch_metadata is the metadata for the batch.
+  string batch_metadata = 10;
+
+  // note is an arbitrary note to include in the origin tx.
+  string note = 11;
+
+  // class_id is the id of the class the credits being bridged belong to.
+  string class_id = 12;
```
ryanchristo, line 487: Should we create an embedded `Batch` and `Project` message to hold these values rather than prefixing each with either `project_` or `batch_` and creating one long list of mixed values? We could then including `origin_tx`, `note`, `start_date`, and `end_date` in `Batch` and `class_id` and `reference_id` in `Project`? It would also be nice to use the same field names so `ref_id` would be `reference_id`.

x/ecocredit/server/core/bridge_receive.go
ryanchristo: Oof. Based on conversations in slack, it looks like there might not be any batch metadata to fill...

x/ecocredit/server/core/bridge_receive.go
ryanchristo: Any reason for not including the end data?

ryanchristo: Any reason for not including the end date?

x/ecocredit/server/core/bridge_receive.go
```diff
@@ -0,0 +1,135 @@
...
+	for it.Next() {
+		project, err := it.Value()
+		if err != nil {
+			return nil, err
+		}
+		projects = append(projects, project)
+	}
+
+	if len(projects) > 1 {
+		return nil, sdkerrors.ErrInvalidRequest.Wrapf("fatal error: bridge service %s has %d projects registered "+
```
aaronc, line 39: Why would this happen? Shouldn't that be an error?

proto/regen/ecocredit/v1/state.proto
technicallyty: that would definitely simplify things up a bit. however, what should we do for projects that don't specify a ref id? should we just set it to the project id?

x/ecocredit/server/core/bridge_receive.go
dave: Remember we're selecting all projects from all classes that have a matching reference ID... The reference ID will just be the Toucan ProjectId (which is the Verra project code) e.g. `VCS-1418`. If we're not 100% confident that this will uniquely identify a single project, then perhaps we should have some other method of identifying a project record?

In the code I wrote yesterday, I added a [bit of json](https://github.com/rnd-inc/toucan-bridge-service/blob/6f80c39dde705bfdf301865de9d9a8b2c162b05a/metadata/metadata_temporary.go#L86-L92) to the project metadata, which [was matched](https://github.com/rnd-inc/toucan-bridge-service/blob/6f80c39dde705bfdf301865de9d9a8b2c162b05a/metadata/metadata_temporary.go#L65-L73) [when selecting products](https://github.com/rnd-inc/toucan-bridge-service/blob/6f80c39dde705bfdf301865de9d9a8b2c162b05a/events.go#L205-L207). Do we need something like that? 

Alternatively perhaps we can add a prefix to the Reference ID to ensure it's unique?

dave: > Remember we're selecting all projects from all classes that have a matching reference ID. 

Or maybe not... I missed that `WithAdminReferenceId` only selects projects with a given admin address... I guess this is OK then?

proto/regen/ecocredit/v1/state.proto
ryanchristo: `WithAdmin(bridgeServiceAddr).WithReferenceId(req.ProjectRefId)` without the added index?

ryanchristo: I don't think we can make this unique given reference id has no unique constraints. We don't provide unique constraints because we want to avoid a situation where someone could potentially name squat a reference id that is not actually the reference id for the project.

proto/regen/ecocredit/v1/tx.proto
technicallyty: embedded here https://github.com/regen-network/regen-ledger/pull/1174/commits/0510f6b0aa80e915d1a49a2e69baf7cc0bf35040

proto/regen/ecocredit/v1/state.proto
ryanchristo: I don't think we can make this unique given reference id has no unique constraints. We added reference id without unique constraints because we want to avoid a situation where someone could potentially name squat a reference id.

technicallyty: that's not possible unfortunately, the index is needed

ryanchristo: I don't think we can make this index unique given reference id has no unique constraints and reference id is only optional. We don't set unique constraints on reference id because we want to avoid a situation where someone could name squat a reference id and we don't require it because some projects may not have one.

technicallyty: > could name squat a reference id
you wouldnt be able to name squat on a unique constraint on an admin+ref_id index key

>  because some projects may not have one

yeah, thats why i suggested we just set it to the generated project ID if they dont supply one, but not really sure how ideal that is

x/ecocredit/server/core/bridge_receive.go
ryanchristo: I think we should still check and leave this as is for now. The larger issue is how we populate the batch metadata that is currently blank in vintages stored in the toucan contracts.

x/ecocredit/server/core/bridge_receive.go
ryanchristo: I think checking metadata is a good idea.

ryanchristo: It looks like `polygon` is only hardcoded for `target` (the reverse direction) at the moment but we should make sure these are aligned. The original idea was source would represent the chain or registry (e.g. polygon, ethereum, verra). I think this still makes sense but not opposed to `toucan` for consideration.

technicallyty: prob best to have it as "polygon"; if toucan deployed on multiple chains, "toucan" source would be too ambiguous

proto/regen/ecocredit/v1/state.proto
ryanchristo: > you wouldnt be able to name squat on a unique constraint on an admin+ref_id index key

True. Only an admin could name squat their own projects... which would be silly.

ryanchristo: > the generated project ID if they dont supply one

This could work. I'm guessing reference id would be associated with the `ClassProject` not the `Project` when we migrate away from projects being strictly defined within a credit class. Seems a bit strange to have the same generated project id stored twice for projects that do not provide a reference id. Curious if @clevinson has any input here.

x/ecocredit/server/core/create_project.go
```diff
@@ -35,6 +35,10 @@ func (k Keeper) CreateProject(ctx context.Context, req *core.MsgCreateProject) (
 		return nil, err
 	}
 
+	if req.ReferenceId == "" {
+		req.ReferenceId = projectID
+	}
```
ryanchristo, line 40: cc @clevinson

proto/regen/ecocredit/v1/state.proto
clevinson: Let's not use the generated project ID for reference ID. It's intended to be an externally set identifier and optional, so I think its better if in state we reflect what the user's intent was (which is to not store a ref ID if they don't set it)

technicallyty: ok ill revert the unique constraint changes. i think it might be safe to remove the constraint and error in the rare case we end up with two projects (i think this literally only happens if our keys are compromised and someone makes another project with the same ref id?). can flesh out further/other solutions on tomorrows call too 🙂

x/ecocredit/server/core/bridge_receive.go
clevinson: re- projects returning more than one result per ref-id: my brain is getting a bit foggy but I think we should discuss this in our bridge sync tomorrow.

Stepping back, I'm thinking of a few things...
- Projects will eventually be disambiguated from credit classes entirely... (with permission less projects)
- Projects (independent from credit classes) likely won't need a reference ID, and reference IDs could feasibly be only assignable to projects once they are adopted into a credit class
- If we choose this approach, we could design it such that issuers are the only ones able to set reference IDs, and reference IDs are scoped to credit classes
- Its probably safe to assume that issuers will act in good faith and in cooperation with each other, and we don't have to worry about issuers "name squatting" on each other
- So without the name squatting concern, we can probably enforce uniqueness of reference ID at the credit class level
- if we do this, then we don't even need the `admin+reference_id` index


So to make this all work, the only change in our assumptions is that "reference IDs are set by credit class issuers as a way to externally identify and index a project in the context of a credit class". And embedded in that is hte assumption that issuers will coordinate and act in good faith with each other to do so in an appropriate way. Since project metadata and project admin can always be updated, I think these are safe assumptions to make.

x/ecocredit/server/core/bridge_receive.go
```diff
@@ -0,0 +1,135 @@
...
+	// we only want the first project that matches the reference ID, so we do not loop here.
+	var project *api.Project
+	if it.Next() {
+		var err error
+		project, err = it.Value()
+		if err != nil {
+			return nil, err
+		}
+	}
+	it.Close()
```
technicallyty, line 42: talked with @clevinson and decided instead of erroring in the expectedly rare event that _multiple_ projects with the same Reference_Id exist within a credit class, we just pick the first project out of the iterator, if any.

This __does__ however leave one vulnerability, which i presume to be incredibly rare, but still worth mentioning. The scenario where things can go wrong:


- user creates Project with `VCS-001` as reference ID in `C01`
- user creates a batch from this project, using all TCO2 data that our bridge would use
- bridge service gets processes an event that __should__ bridge into `VCS-001` in that batch bob created, triggering a `MintBatchCredits` call
- this call would **fail** as `MintBatchCredits` is currently gated by the issuer address, which would be the user, not the bridge service address.

again, i presume this is not likely, but just want to be sure we are all aware and if its something we should be considering and if so if we should be preparing some sort of procedure if that scenario does occur.

technicallyty: one remedy to eliminate this situation entirely: only match batches that have the same issuer address as the bridge service's address

x/ecocredit/core/msg_bridge_receive.go
```diff
@@ -0,0 +1,88 @@
...
+	if len(batch.Metadata) > MaxMetadataLength {
+		return sdkerrors.ErrInvalidRequest.Wrapf("batch metadata length (%d) exceeds max metadata length: %d", len(batch.Metadata), MaxMetadataLength)
+	}
+	if len(batch.Note) > MaxMetadataLength { // do we need to do this?
+		return sdkerrors.ErrInvalidRequest.Wrapf("note length (%d) exceeds max length: %d", len(batch.Note), MaxMetadataLength)
+	}
+
+	// project validation
+	if m.Project == nil {
+		return sdkerrors.ErrInvalidRequest.Wrapf("project cannot be empty")
```
ryanchristo, line 66:
```suggestion
		return sdkerrors.ErrInvalidRequest.Wrap("project cannot be empty")
```

ryanchristo (commented):
Proto and message validation review comments. Still working my way through.

proto/regen/ecocredit/v1/state.proto
```diff
@@ -83,6 +83,7 @@ message Project {
     index : {id : 2, fields : "class_key,id", unique : true}
     index : {id : 3, fields : "admin"}
     index : {id : 4, fields : "reference_id"}
+    index : {id : 5, fields : "class_key,reference_id"}
```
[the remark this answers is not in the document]
technicallyty: that brings us back to here https://github.com/regen-network/regen-ledger/pull/1174#discussion_r897059786, what to do when no ref id is provided?

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -469,5 +473,73 @@ message MsgBridge {
...
+
+// MsgBridgeReceive is the Msg/BridgeReceive request type.
+message MsgBridgeReceive {
+
+  // service_address is the account address of the service bridging the credits.
+  string service_address = 1;
+
+  Batch batch = 2;
+
+  Project project = 3;
```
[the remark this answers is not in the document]
technicallyty: seems like i would just be copying the comment in the embedded message. should we just copy the comment or just duplicate them on both the field and the definition?

proto/regen/ecocredit/v1/state.proto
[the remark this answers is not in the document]
ryanchristo: Right... I'm seeing the comment above now about multiple projects with the same reference id and deciding to pick the first project rather than error.

proto/regen/ecocredit/v1/tx.proto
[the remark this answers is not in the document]
ryanchristo: In the `MsgSend` example, we have a short one liner and then a more detailed comment for the message. I think the same would apply here. Maybe just the first line `project defines the project information for the bridged credits`.

x/ecocredit/core/msg_bridge_receive.go
```diff
@@ -0,0 +1,88 @@
...
+	if batch.EndDate == nil {
+		return sdkerrors.ErrInvalidRequest.Wrap("end_date is required")
+	}
+	if batch.StartDate.After(*batch.EndDate) {
+		return sdkerrors.ErrInvalidRequest.Wrap("start_date must be a time before end_date")
+	}
+	if len(batch.Metadata) > MaxMetadataLength {
+		return sdkerrors.ErrInvalidRequest.Wrapf("batch metadata length (%d) exceeds max metadata length: %d", len(batch.Metadata), MaxMetadataLength)
+	}
+	if len(batch.Note) > MaxMetadataLength { // do we need to do this?
```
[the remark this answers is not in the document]
technicallyty: made a constant for this here https://github.com/regen-network/regen-ledger/pull/1174/commits/ed73285c4dc02776ce432a32bc7ce75202bb2086

proto/regen/ecocredit/v1/tx.proto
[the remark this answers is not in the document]
technicallyty: https://github.com/regen-network/regen-ledger/pull/1174/commits/ee75f785fadbeafb4f3fbcc9322631758095d4b8

technicallyty:
Current Flow:

- projects are no longer iterated over, we simply pick the first project that matches the <Class_id>+<Refrence_Id> index.
- batches are matched based on denom, start+end+metadata+issuer. Not so sure about matching on metadata though. could be an issue if a batch has already been issued/bridged on the ledger, and then the external registry changes it's metadata... causing it to not match and then mint a fresh batch

x/ecocredit/core/msg_bridge_receive.go
```diff
@@ -0,0 +1,88 @@
...
+
+// GetSignBytes implements the LegacyMsg interface.
+func (m MsgBridgeReceive) GetSignBytes() []byte {
+	return sdk.MustSortJSON(ecocredit.ModuleCdc.MustMarshalJSON(&m))
+}
+
+// ValidateBasic does a sanity check on the provided data.
+func (m *MsgBridgeReceive) ValidateBasic() error {
+	if _, err := sdk.AccAddressFromBech32(m.Issuer); err != nil {
+		return sdkerrors.ErrInvalidAddress.Wrap("address")
```
ryanchristo, line 28:
```suggestion
		return sdkerrors.ErrInvalidAddress.Wrap("issuer")
```

x/ecocredit/server/core/bridge_receive.go
```diff
@@ -0,0 +1,138 @@
...
+
+	batches := make([]*api.Batch, 0)
+	for bIt.Next() {
+		batch, err := bIt.Value()
+		if err != nil {
+			return nil, err
+		}
+		// the timestamp stored in the batch is more granular than the date in the denom representation, so we match here.
+		if batch.StartDate.AsTime().Equal(*req.Batch.StartDate) &&
+			batch.EndDate.AsTime().Equal(*req.Batch.EndDate) &&
```
ryanchristo, line 92: Thinking back to the recent audit pull request, and noticing local time is being used in the tests above, I think we may have an issue here if the request is local time and the time stored in state is UTC...

ryanchristo: We should probable use `req.Batch.StartDate.UTC()` and `req.Batch.EndDate.UTC()` and we may want to test both local and UTC times in the request and always UTC time in state.

ryanchristo: We should probably also double check that we want to always store UTC time and we may want to open a follow up issue to make sure we have dates stored as expected and maybe update proto comments to note that the time will be stored in UTC... cc @clevinson

x/ecocredit/server/core/bridge_receive.go
```diff
@@ -0,0 +1,138 @@
...
+			batches = append(batches, batch)
+		}
+	}
+	it.Close()
+
+	// TODO(Tyler): potentially select a batch by oldest issuance date?
+	amtBatches := len(batches)
+	if amtBatches > 1 {
+		return nil, sdkerrors.ErrInvalidRequest.Wrapf("fatal error: bridge service %s has %d batches issued "+
+			"with start %v and end %v dates in project %s", bridgeServiceAddr.String(), len(batches), req.Batch.StartDate, req.Batch.EndDate, project.Id)
```
ryanchristo, line 104: I think oldest issuance date would be safer than an error here.

x/ecocredit/server/core/bridge_receive.go
```diff
@@ -0,0 +1,138 @@
...
+			EndDate:   req.Batch.EndDate,
+			Open:      true,
+			OriginTx:  req.Batch.OriginTx,
+			Note:      req.Batch.Note,
+		})
+		if err != nil {
+			return nil, err
+		}
+		return &core.MsgBridgeReceiveResponse{BatchDenom: batchRes.BatchDenom, ProjectId: projectRes.ProjectId}, nil
+	}
```
ryanchristo, line 73: Should we be using `else` and returning the response once at the end of `BridgeReceive`. I'm noticing we don't have an event for `BridgeReceive` and we may want to add one and we would need to add to each condition.

If each condition was handled without the returned response, we could define the event and response before running through the conditions, set within each condition, and then emit and return at the end.

x/ecocredit/server/core/bridge_receive.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// batches are matched on their denom, iterating over all batches within the <ProjectId>-<StartDate>-<EndDate> range.
+	// any batches in that iterator that have matching metadata, are added to the slice.
+	// idx will be of form C01-001-20210107-20210125-" catching all batches with that project Id and in the date range.
```
ryanchristo, line 77: It might be nice to break out the matching logic into its own private helper method. That would essentially return a bool whether batch exists or not and then we could have an `if/else` for `batchExists` that either calls create or mint. I think this would make things a bit more readable and we could test separately.

ryanchristo (commented):
Looking good. Dug a little deeper this time.

x/ecocredit/server/core/bridge_receive.go
ryanchristo: Ideally this would never happen but better to probably better to bridge the credits if matching batches exist rather than to error and have the bridge operation fail. We should do whatever we can to prevent two matching batches earlier in the bridge process so that this would never happen.

x/ecocredit/server/core/bridge_receive.go
ryanchristo: We can add an event in a follow up and consider a single return response when/if we do. Ok to leave as is but was having a hard time following each condition, which might also be resolved by pulling out the matching logic.

technicallyty: hows this https://github.com/regen-network/regen-ledger/pull/1174/commits/794e070947ce135fa8739adeb7d82b84abd24f74 ?

x/ecocredit/server/core/bridge_receive.go
technicallyty: does this help? https://github.com/regen-network/regen-ledger/pull/1174/commits/794e070947ce135fa8739adeb7d82b84abd24f74

ryanchristo: Yea, this helps. :+1:

x/ecocredit/server/core/bridge_receive.go
technicallyty: switched everything to convert to UTC, also in this commit https://github.com/regen-network/regen-ledger/pull/1174/commits/794e070947ce135fa8739adeb7d82b84abd24f74

x/ecocredit/server/core/bridge_receive.go
technicallyty: ok using the oldest issuance date first logic in https://github.com/regen-network/regen-ledger/pull/1174/commits/794e070947ce135fa8739adeb7d82b84abd24f74

x/ecocredit/server/core/bridge_receive.go
ryanchristo: This helps :+1:

x/ecocredit/server/core/bridge_receive.go
```diff
@@ -0,0 +1,170 @@
...
+	// TODO: emit event?
+	return response, nil
+}
+
+// getBatchFromBridgeReq attempts to retrieve a batch from state given the request.
+// In the event that multiple batches are matched, the batch with the oldest issuance date is selected.
+// When no batches are found, nil is returned for both return values.
+func (k Keeper) getBatchFromBridgeReq(ctx context.Context, req *core.MsgBridgeReceive_Batch, projectId string, bridgeAddr sdk.AccAddress) (*api.Batch, error) {
+	// batches are matched on their denom, iterating over all batches within the <ProjectId>-<StartDate>-<EndDate> range.
+	// any batches in that iterator that have matching metadata, are added to the slice.
```
clevinson, line 106:
```suggestion
	// any batches in that iterator that were created by the same issuer and have matching metadata, are added to the slice.
```

x/ecocredit/server/core/bridge_receive_test.go
```diff
@@ -0,0 +1,316 @@
...
+		Project: &core.MsgBridgeReceive_Project{
+			ReferenceId:  "VCS-001",
+			Jurisdiction: "US-KY",
+			Metadata:     "foo",
+			ClassId:      "C01",
+		},
+	}
+	res, err := s.k.BridgeReceive(s.ctx, &msg)
+	assert.NilError(t, err)
+
```
clevinson, line 134: Just to be safe, could we explicitly assert that the project bridged to is not `VCS-002` ?

x/ecocredit/server/core/bridge_receive_test.go
```diff
@@ -0,0 +1,316 @@
...
+			ClassId:      "C01",
+		},
+	}
+	res, err := s.k.BridgeReceive(s.ctx, &msg)
+	assert.NilError(t, err)
+	// check to make sure the first project is selected
+	assert.Equal(t, res.ProjectId, project.Id)
+}
+
+func TestBridgeReceive_ChoosesLatestBatch(t *testing.T) {
```
clevinson, line 247:
```suggestion
func TestBridgeReceive_ChoosesOldestBatch(t *testing.T) {
```

x/ecocredit/server/core/bridge_receive_test.go
```diff
@@ -0,0 +1,316 @@
...
+	assert.NilError(t, err)
+	bal, err := s.k.Balance(s.ctx, &core.QueryBalanceRequest{
+		Address:    recipient,
+		BatchDenom: batch.Denom,
+	})
+	assert.NilError(t, err)
+	assert.Equal(t, bal.Balance.TradableAmount, msg.Batch.Amount)
+}
+
+func TestBridgeReceive_TooManyBatches(t *testing.T) {
```
clevinson, line 145: Not sure I'm understanding this test scenario. Isn't it the same as `TestBridgeReceive_ChoosesLatestBatch()` below? Aren't we not erroring any more on multiple matching batches?

x/ecocredit/server/core/bridge_receive_test.go
```diff
@@ -0,0 +1,316 @@
...
+			ReferenceId:  refId,
+			Jurisdiction: "US-KY",
+			Metadata:     "",
+			ClassId:      "C01",
+		},
+	}
+	res, err := s.k.BridgeReceive(s.ctx, &msg)
+	assert.NilError(t, err)
+	// check to make sure the first project is selected
+	assert.Equal(t, res.ProjectId, project.Id)
```
clevinson, line 244: I know this is an edge case, but I'm curious -- do we actually have guaranteed ordering here?

x/ecocredit/server/core/bridge_receive.go
clevinson: Hmm I'm not sure I fully understand the nuance here. But we should be making sure that if a user is setting "Jan 1, 2020" that its not doing some weird local timestamp, that when converted to UTC actually reads it as "Jan 2" or "Dec 31"...  Is this something we can really ensure against?

If our intention is to only allow dates, maybe `ValidateBasic()` could ensure that the when evaluated in UTC, the timestamp % day == 00:00:00, aka that the timestamp is rounded to the nearest day @ midnight UTC?

clevinson (approved):
Nice work! 

Pre-approving with some high level notes & suggested follow-ups:
- the proto API generally looks good, though I left some suggestions about restructuring some fields
- implementation generally looks good! I think this matches our intended approach
- One test (for multiple batches) seemed to be based off an older expected behavior (failing on too many batches). I think we should delete this test in favor of `ChoosesLatestBatch()` test

Follow-ups that come to mind:
- translate tests to gherkin & add more cases?
- I don't think we're enforcing uniqueness yet on project reference_id. We should add a follow-up / issue for adding that uniqueness constraint at the msg server layer (`CreateProject()` handler). We likely cannot do it at the ORM layer since reference_id is sometimes null.

x/ecocredit/server/core/bridge_receive_test.go
technicallyty: done in https://github.com/regen-network/regen-ledger/pull/1174/commits/8842f9b99f1aaa5347a30946ec929a48e7d84d1d

x/ecocredit/server/core/bridge_receive_test.go
technicallyty: ah yeah this test should be stale now, removed it in https://github.com/regen-network/regen-ledger/pull/1174/commits/47f604c0b5a1d3b7efde98adf016ce462672fc7b

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -469,5 +473,79 @@ message MsgBridge {
...
+
+    // jurisdiction is the project jurisdiction.
+    string jurisdiction = 6;
+
+    // metadata is the metadata for the project.
+    string metadata = 9;
+
+    // id is the unique identifier of the credit class within which the project
+    // and credit batch already exist or will be created.
+    string class_id = 12;
```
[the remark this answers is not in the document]
technicallyty: refactored here https://github.com/regen-network/regen-ledger/pull/1174/commits/914e9a9c67b25b79681a4d6a2223310b9e5332d1

x/ecocredit/server/core/bridge_receive_test.go
technicallyty: yeah both the [string codec](https://github.com/cosmos/cosmos-sdk/blob/main/orm/encoding/ormfield/string.go#L22-L24) and [uint64 codec](https://github.com/cosmos/cosmos-sdk/blob/main/orm/encoding/ormfield/uint64.go#L22-L24) are ordered fields. although i'm not sure if combining them as index keys makes any difference... maybe @aaronc  can confirm?

x/ecocredit/server/core/bridge_receive.go
technicallyty: hmm i feel like the rounding might be okay, as long as the entire chain speaks in UTC terms? i think it would make the most sense to either drop support for explicit time values (only dates), or just speak UTC throughout the chain.

dave:
> Current Flow:
> Not so sure about matching on metadata though. could be an issue if a batch has already been issued/bridged on the ledger, and then the external registry changes it's metadata... causing it to not match and then create a fresh batch

Yeah matching the whole metadata string seems wrong to me, considering we don't know exactly what will be in that field... This check would put a restriction on what we can actually store there...

x/ecocredit/server/core/bridge_receive.go
dave: @clevinson @technicallyty the Toucan data suffers from this problem. All the date/times are stored as UNIX date/times - e.g. without timezone data. They've obviously been entered as an exact date, but the local timezone where they've been inserted is a few hours off from UTC, so all the dates/times in the Toucan data are a few hours before the exact date. ([See raw Toucan data here on the Vintage tab](https://docs.google.com/spreadsheets/d/1NhhGPrSZoc3_ptcHpUjWn_XR71VFjuTQhu7Gf21oyk8/edit#gid=561399407)). 

We [discussed this in the meta data conversion ticket here](https://github.com/rnd-inc/toucan-bridge-service/issues/48), and decided to round the Toucan date/time data to the nearest whole day. [The conversion code is here](https://github.com/rnd-inc/toucan-bridge-service/blob/main/metadata/metadata.go#L66-L69), [here](https://github.com/rnd-inc/toucan-bridge-service/blob/main/metadata/metadata.go#L46-L47), and [here](https://github.com/rnd-inc/toucan-bridge-service/blob/main/metadata/metadata_test.go#L9-L30).

dave: @clevinson as per slack comment, I've just realised that the Toucan data was actually /not/ off by one hour, and that issue and all the discussion about it was nonsense. However, this highlights the importance of using `.UTC()` everywhere in Go, or you'll get times converted to whatever the local timezone of the server is!

dave: ... however, the regen ledger code that generates the batch denom [isn't using .UTC()](https://github.com/regen-network/regen-ledger/blob/852fb8d82ecef8823f322894c8f0ce7610e31a0a/x/ecocredit/core/utils.go#L87-L91), and I assume the reason that this works fine is that the timezone of the server (and probably all servers) is set to UTC... Whereas when I ran the script to read the live data from the Toucan blockchain, I ran it on my Mac with a local  timezone set. 🤔

technicallyty:
> Opened #1193 for followups. Is this ready for review again?

yup good to review now

x/ecocredit/server/core/bridge_receive.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// getBatchFromBridgeReq attempts to retrieve a batch from state given the request.
+// In the event that multiple batches are matched, the batch with the oldest issuance date is selected.
+// When no batches are found, nil is returned for both return values.
+func (k Keeper) getBatchFromBridgeReq(ctx context.Context, req *core.MsgBridgeReceive_Batch, projectId string, bridgeAddr sdk.AccAddress) (*api.Batch, error) {
+	// batches are matched on their denom, iterating over all batches within the <ProjectId>-<StartDate>-<EndDate> range.
+	// any batches in that iterator that were created by the same issuer and have matching metadata, are added to the slice.
+	// idx will be of form C01-001-20210107-20210125-" catching all batches with that project Id and in the date range.
+	batchIdx := fmt.Sprintf("%s-%s-%s-", projectId, req.StartDate.Format("20060102"), req.EndDate.Format("20060102"))
```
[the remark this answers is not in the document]
ryanchristo: I think `UTC()` for now to be consistent with the current implementation but we can reevaluate as a followup.

```suggestion
	batchIdx := fmt.Sprintf("%s-%s-%s-", projectId, req.StartDate.UTC().Format("20060102"), req.EndDate.UTC().Format("20060102"))
```

ryanchristo (approved):
Nice work. This looks good to me.

A couple small suggestions and anything else we can address in followups. The only concerns I have at the moment are about matching timestamps with regard to local/UTC and metadata and I think adding/transition to gherkin/gocuke tests would help clarify intended behavior.

I opened an issue to address followups so please add any other open questions to the issue and might be best to move some of the discussion around those items to the issue. #1193

x/ecocredit/core/origin_tx.go
```diff
@@ -0,0 +1,15 @@
...
+
+func (m OriginTx) Validate() error {
+	if len(m.Id) == 0 {
+		return fmt.Errorf("invalid OriginTx: no id")
+	}
+	if len(m.Source) == 0 {
+		return fmt.Errorf("invalid OriginTx: no source")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
technicallyty: ah got it- fixed here https://github.com/regen-network/regen-ledger/pull/1174/commits/892514da474fa4ad50538a47d5b36b834eb5282a

x/ecocredit/core/msg_bridge_receive_test.go
```diff
@@ -0,0 +1,210 @@
...
+			getMsg: func(validMsg MsgBridgeReceive) MsgBridgeReceive {
+				validMsg.ClassId = "Foobar12345"
+				return validMsg
+			},
+			errMsg: "class ID didn't match the format",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
```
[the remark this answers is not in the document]
technicallyty: yes, `Run` blocks until it returns, or you can call `t.Parallel` within `Run` to make all test cases in the table test run concurrently. The outer `t.Parallel` just signals the test can be run in parallel with other tests.

x/ecocredit/core/origin_tx.go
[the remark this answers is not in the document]
technicallyty: err, i just removed my code and used the existing `validateOriginTx` - is that ok too? or should we move to a receiver method like i added?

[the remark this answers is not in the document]
ryanchristo: Either is ok for now but I think a receiver method would be nice.

x/ecocredit/server/core/bridge_receive.go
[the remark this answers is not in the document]
ryanchristo: Seems like we should be using `UTC` when formatting the denom if we are storing `UTC` for start and end date. If not, the denom might not actually match the start and end state stored.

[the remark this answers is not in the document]
technicallyty: okay i dug a little deeper into our `time` usage, and we are pretty much fine save for a few cases (sell order expiration,  updating sell order expiration, and formatting batch denom). everything else is already using timestamppb, or gogo timestamp, which converts to `unix/UTC` under the hood.

opened https://github.com/regen-network/regen-ledger/pull/1194 to solidify UTC usage where its still needed and will merge the suggestion above

Issue #1127, opened 2022-05-24
Reassess need for `contract` as a field in MsgBridge() & MsgBridgeReceive()


clevinson:
> One thing that came up in discussion with @ryanchristo today is that we may actually want to revise our thinking around "contract" being an explicit field in the `Bridge()` API call...
> 
> This was my oversight when writing out #1084... From a security perspective, we can't only rely on an end user's desired "contract" for tokens to get minted into on the destination chain. As this may not actually be the corresponding TCO2 contract for the given batch denom.  Since the bridge service will anyway need to manage their own mapping between batch denoms & TCO2 contracts on the destination chain, there probably isn't much added value in the user providing this upfront...

_Originally posted by @clevinson in https://github.com/regen-network/regen-ledger/pull/1101#pullrequestreview-983975547_

@aaronc do you have thoughts on this as you're pretty close to the bridge work?

aaronc:
I don't have a strong opinion. You guys might know more than me. Maybe we should also ask Toucan?

clevinson:
Following up here with some comments I had previously made in #1206.

From conversations we've had both at regen + with Toucan folks, it seems that we've reached alignment on adding a `contract` field into the `MsgBridgeReceive()` call so that regen ledger will store on-chain a mapping between ecocredit batches and Toucan contract addresses on polygon.

This indeed does mean that we can remove the `contract` field from the `MsgBridge()` call, as there's no need for an end user on the Regen side to state the destination contract address on Polygon, as this will be inferred from the on-chain mapping in regen ledger.

Some outstanding questions still remain though...

### Inconsistent contract addresses for a given project vintage

While I understand it's intended that project vintages on Polygon correspond 1-1 with TCO2 contract addresses, we still need to decide on how we handle the unlikely edge case of a `MsgBridgeRecieve()` with the same project vintage metadata, but a different contract address than has been previously used.

Example would look like:
- Alice bridges 500 credits to regen ledger from Polygon. The credits come from contract address `0xA` which represents project Foo, 2015 vintage on Verra.
- This creates a new ecocredit batch on regen ledger corresponding to project Foo, 2015 vintage, and it stores the Polygon contract address for this vintage as `0xA`.
- Bob later bridges 200 credits to regen ledger from Polygon. These credits come from a contract address `0xB` which has metadata corresponding to the same project Foo and 2015 vintage.

2 options for how we handle Bob's `MsgBridgeReceive()`:
[...]

Issue #1193, opened 2022-06-16
MsgBridgeReceive implementation followups


ryanchristo:
## Summary

In #1174 we added `MsgBridgeReceive` to reduce the likelihood of race conditions and to simplify requirements. The following items still need to be addressed:

- [ ] Should we error in the event that there are two projects? https://github.com/regen-network/regen-ledger/pull/1174#discussion_r897477885
- [ ] Should we check for matching metadata when matching a batch? https://github.com/regen-network/regen-ledger/pull/1174#issuecomment-1157591742
- [ ] Double check use of `UTC` in batch creation and ensure matching is sound https://github.com/regen-network/regen-ledger/pull/1174#discussion_r898587281
- [ ] Review and potentially migrate tests to use gherkin/gocuke https://github.com/regen-network/regen-ledger/pull/1174#pullrequestreview-1008308927 
- [ ] Unique constraint on reference_id at msg server level https://github.com/regen-network/regen-ledger/pull/1174#pullrequestreview-1008308927 

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1206, opened 2022-06-22
refactor(x/ecocredit): update origin tx and add contract mapping

ryanchristo:
## Description

Closes: #1127 / Ref: #1193

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1/events.proto
```diff
@@ -193,4 +194,19 @@ message EventBridge {
 
   // contract is the contract address.
   string contract = 3;
-}
\ No newline at end of file
+
+  // amount is the amount of credits.
+  string amount = 4;
```
ryanchristo, line 199: Added amount so that the bridge service receives all information needed in the event.

proto/regen/ecocredit/v1/state.proto
```diff
@@ -285,9 +288,12 @@ message BatchOriginTx {
   // the mint process (e.g. polygon, ethereum, verra).
   string source = 2;
 
-  // note is a reference note for accounting that is passed to an event.
-  string note = 3;
```
ryanchristo, line 289: Note was not intended to be something we stored in state, only passed to events including origin tx.

ryanchristo: Also questioning whether we need note and unclear on the original intention for how it should be used.

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -221,11 +226,6 @@ message MsgCreateBatch {
   // issuing credits and should only be set when bridging assets from another
   // chain or registry as a result of a bridge operation.
   OriginTx origin_tx = 8;
-
-  // note is a reference note for accounting that should only be set when
-  // bridging assets from another chain or registry as a result of a bridge
-  // operation and will be stored in state as a field of BatchOriginTx.
-  string note = 9;
```
ryanchristo, line 228: Added to origin_tx so removed here.

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -253,11 +253,6 @@ message MsgMintBatchCredits {
   // origin_tx is the transaction from another chain or registry that triggered
   // the minting of credits.
   OriginTx origin_tx = 4;
-
-  // note is a reference note for accounting that should only be set when
-  // bridging assets from another chain or registry as a result of a bridge
-  // operation and will be stored in state as a field of BatchOriginTx.
-  string note = 5;
```
ryanchristo, line 260: Same as above. Added to origin tx so removed here.

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -482,22 +474,34 @@ message MsgBridgeReceive {
   // issuer is the account address of the service bridging the credits.
   string issuer = 1;
 
-  // batch defines the credit batch information for the bridged credits.
-  Batch batch = 2;
+  // class_id is the unique identifier of the credit class within which the
+  // project and credit batch already exist or will be created.
+  string class_id = 2;
```
ryanchristo, line 479: Just reordering here to class, project, batch, which seems more natural.

x/ecocredit/core/msg_mint_batch_credits.go
```diff
@@ -99,5 +96,11 @@ func validateOriginTx(o *OriginTx, required bool) error {
 	if !reOriginTxSource.MatchString(o.Source) {
 		return errBadReq.Wrap("origin_tx.source must be at most 32 characters long, valid characters: alpha-numberic, space, '-' or '_'")
 	}
+	if len(o.Contract) > 0 && !isValidEthereumAddress(o.Contract) {
+		return sdkerrors.ErrInvalidAddress.Wrapf("%s is not a valid ethereum address", o.Contract)
+	}
```
ryanchristo, line 101: Was not sure whether we should always require a valid ethereum address when validating origin tx and required is set to true in the validation function. Dynamic minting could have other use cases.

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -466,11 +461,8 @@ message MsgBridge {
   // recipient is the address of the account receiving the bridged credits.
   string recipient = 3;
 
-  // recipient is the address of the contract handling the bridged credits.
-  string contract = 4;
```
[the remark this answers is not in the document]
ryanchristo: #1127

technicallyty (dismissed):
lgtm

clevinson (commented):
Thanks for taking this on @ryanchristo 

I think generally having contract stored on the ledger side makes sense, and agree that it makes bridge service implementation a bit cleaner.

The expectation for contract field to be 1-1 to batch denom without enforcing it on-chain still makes me a bit uncomfortable though. Can we at least add some enforcement in the MsgBridgeReceive handler so we error out if the bridge service tries to bridge to an existing batch with a new contract address?

x/ecocredit/server/core/bridge.go
```diff
@@ -19,13 +21,41 @@ func (k Keeper) Bridge(ctx context.Context, req *core.MsgBridge) (*core.MsgBridg
...
+		// we only want the first origin tx that matches the batch denom
+		var originTx *api.BatchOriginTx
+		if it.Next() {
+			var err error
+			originTx, err = it.Value()
+			if err != nil {
+				return nil, err
+			}
+		}
+		it.Close()
```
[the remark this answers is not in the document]
clevinson: Thinking about this more. If it was considered less desirable to error out, I could see us considering the following alternative:

- Create a new batch denom when a new contract field is provided (for a given project + batch start/end/metadata set)
- Change lookups of batch denom for a given `MsgBridgeReceiveRequest` to not just lookup based on `project + batch_start_date + batch_end_date + batch_metadata`, but instead `project + batch_start_date + batch_end_date + batch_metadata + contract`

I think the question we should be asking not whether Toucan ever expects to "change TCO2 contract address", as this is really not possible given their usage of proxy contracts. Instead, the question is if whether or not Toucan will ever have two TCO2 contracts for the same set of "project vintage data". If this is possible, then we probably want to handle gracefully (following the alternative strategy illustrated here above).  If this is not possible, then it's probably sufficient to error out if a project vintage ever is attempted to be bridged to Regen on two occasions with two different source contract addresses.

@dave do you have any thoughts on this?

ryanchristo:
Breaking this out into smaller pull requests. Also need to discuss expectations with 1-to-1 relationship between a contract and a credit batch. If we make contract unique and always have a 1-to-1 relationship, we can simplify the matching for exiting credit batches and we wouldn't have to use batch with oldest issuance date. Something to discuss on our call tomorrow.

Issue #1023, opened 2022-04-14
Simple Send Credit CLI Command


technicallyty:
## Summary

The current CLI cmd for Sending credits works, but is a bit cumbersome for users who may just want to execute a simpler transaction (sending tradable credits from 1 account to another). 

## Proposal

We should make a CLI comand for simple credit transfers. We could call this something like `send-single` or `simple-send` or call this `send` and rename the existing one to `send-batch`

it could work like so:

### Send tradable
```
regen tx ecocredit send-simple 45.2351 C01-20200101-20220101-001 regen1zz22dfpvw3zqpeyhvhmx944a588fgcal3u7ss3
```
### Send retired
```
regen tx ecocredit send-simple 45.2351 C01-20200101-20220101-001 regen1zz22dfpvw3zqpeyhvhmx944a588fgcal3u7ss3  --retire-to="SOME-LOCATION"
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ryanchristo:
I think this would be a nice addition but `send-simple` is not very descriptive and `send-batch` might be confusing since this is in relation to credits that are issued in batches.

ryanchristo:
I think this would be a nice addition but `send-simple` is not very descriptive and `send-batch` might be confusing since this is in relation to credits that are issued in batches. I think `send` and `send-batch` would be ok though.

PR #1213, opened 2022-06-27
feat(x/ecocredit): add simpler version of `send` command

sfishel18:
## Description

Closes: #1023 

Renames the existing `send` command to `send-bulk`, and adds implementation of a simpler `send` command that can only handle a single transaction of a single credit type but does not require YAML arguments.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


sfishel18:
hi @ryanchristo and @technicallyty, this is still WIP so no need for an in-depth review, but i'm somewhat new to Go and to this project so wanted to check in and make sure this is on the right track. 

i still need to implement the `--retire-to` option and add tests.

also, i found my self getting confused with `batch` in the context of a credit batch and `send-batch` as the new name for what the `send` command currently does, so i went with `send-bulk` instead. but of course it's easy to change.

technicallyty:
hey thanks for the contribution! the intent of the issue was to *only* add a new CLI command (so no additional protobuf messages/rpc methods), and to use the existing `MsgSend` and `Send` method inside the command, as they are setup up already to handle any number of credits. Let me know if that makes sense or if theres any question i can answer 😄

sfishel18:
> hey thanks for the contribution! the intent of the issue was to _only_ add a new CLI command (so no additional protobuf messages/rpc methods), and to use the existing `MsgSend` and `Send` method inside the command, as they are setup up already to handle any number of credits. Let me know if that makes sense or if theres any question i can answer 😄

ahhh, i see. yeah that totally makes sense. i'll take another crack at it, thanks!

ryanchristo (commented):
Thanks again for another contribution!

This is looking good. And it looks like you got some experience with proto even if that's not what was needed for this pull request. We just made some changes to the transaction commands so it looks like there will be a few conflicts to resolve. It would be great to add a test as well, which should be pretty straightforward using an existing transaction command test as an example.

x/ecocredit/client/tx.go
```diff
@@ -318,7 +319,39 @@ Parameters:
...
+  recipient: recipient address`,
+		Args: cobra.ExactArgs(3),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := sdkclient.GetClientTxContext(cmd)
+			if err != nil {
+				return err
+			}
+			credit := core.MsgSend_SendCredits{
+				TradableAmount: args[0],
+				BatchDenom:     args[1],
```
ryanchristo, line 338: I think it would be nice to add an argument for retirement amount as well. You can use an optional flag for retirement jurisdiction. The retirement jurisdiction flag doesn't need to be required in case retired amount is `0`.

`GenerateOrBroadcastTxCLI` calls a message validation function so the command will error if retirement jurisdiction is not set and retired amount is positive, so not need to worry about that logic within the command logic.

x/ecocredit/client/tx.go
```diff
@@ -318,7 +319,39 @@ Parameters:
...
+			}
```
ryanchristo, line 339: I think it would be nice to add an argument for retirement amount as well. You can use an optional flag for retirement jurisdiction. The retirement jurisdiction is only required if retired amount is positive.

`GenerateOrBroadcastTxCLI` calls a message validation function so the command will error if retirement jurisdiction is not set and retired amount is positive, so no need to worry about that within the command logic.

ryanchristo: You will need to set the value provided to the optional retirement jurisdiction flag within this struct though. It will be an empty string if the flag is not set and the message validation will handle the rest.

sfishel18:
> Thanks again for another contribution!
> 
> This is looking good. And it looks like you got some experience with proto even if that's not what was needed for this pull request. We just made some changes to the transaction commands so it looks like there will be a few conflicts to resolve. It would be great to add a test as well, which should be pretty straightforward using an existing transaction command test as an example.

my pleasure! thanks for the detailed feedback, i'll get another rev of this up for review in the next few days

x/ecocredit/client/testsuite/tx.go
```diff
@@ -331,6 +331,88 @@ func (s *IntegrationTestSuite) TestTxCreateBatchCmd() {
 func (s *IntegrationTestSuite) TestTxSendCmd() {
```
sfishel18, line 331: question: how do i run these tests? `go test` doesn't seem to find them

ryanchristo: `make test` will run all tests including integration tests

ryanchristo: `make test` will run all unit and integration tests

technicallyty: you can run _only_ the receiver suite by running the `TestEcocreditIntegration` in `app/testsuite/modules_test.go`

x/ecocredit/client/tx.go
```diff
@@ -297,8 +298,51 @@ Example JSON:
...
-		Use:   "send [recipient] [credits]",
+		Use:   "send [amount] [batch_denom] [recipient]",
+		Short: "Sends credits from the transaction author (--from) to the recipient",
+		Long: `Sends credits from the transaction author (--from) to the recipient.
+
+Parameters:
+  amount: amount to send
+  batch_denom: batch denomination
+  recipient: recipient address`,
+		Args: cobra.ExactArgs(3),
```
technicallyty, line 311: we can populate the `Example` field to the `Command` struct here with something like `Example: "regen tx ecocredit send 20 regen18xvpj53vaupyfejpws5sktv5lnas5xj2phm3cf C01-001-20200403-20210405-001 --from myKey`

x/ecocredit/client/tx.go
```diff
@@ -297,8 +298,51 @@ Example JSON:
...
+			}
+			msg := core.MsgSend{
+				Sender:    clientCtx.GetFromAddress().String(),
+				Recipient: args[2],
+				Credits:   []*core.MsgSend_SendCredits{&credit},
+			}
+			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), &msg)
+		},
+	}
+	cmd.Flags().StringVar(&retireTo, "retire-to", "", "Jurisdiction to retire the credits to. If empty, credits are not retired. (default empty)")
```
technicallyty, line 337: nit: we generally don't use the `Var` type flags - see how we use additional flags here https://github.com/regen-network/regen-ledger/blob/master/x/ecocredit/client/tx.go#L605-L623

x/ecocredit/client/tx.go
```diff
@@ -297,8 +298,51 @@ Example JSON:
...
+	return txFlags(cmd)
+}
+
+// TxSendBulkCmd returns a transaction command that sends credits from one account
+// to another.
```
technicallyty, line 342:
```suggestion
// TxSendBulkCmd returns a transaction command that can send credits from multiple batches from one account
// to another.
```

x/ecocredit/client/tx.go
```diff
@@ -297,8 +298,51 @@ Example JSON:
 // TxSendCmd returns a transaction command that sends credits from one account
 // to another.
```
technicallyty, line 299:
```suggestion
// TxSendCmd returns a transaction command that sends credits from a single batch from one account
// to another.
```

x/ecocredit/client/tx.go
technicallyty: not sure if this is the best doc, maybe @ryanchristo can suggest something cleaner given these udpates

technicallyty (commented):
great work! just a few nits and then this should be good to go! 🎉

sfishel18:
> great work! just a few nits and then this should be good to go! 🎉

thanks! all comments addressed

ryanchristo (commented):
This is looking good. Just a couple small changes and then I think it's good to go.

x/ecocredit/client/tx.go
```diff
@@ -29,6 +29,7 @@ const (
 	FlagRemoveIssuers string = "remove-issuers"
 	FlagReferenceId   string = "reference-id"
 	FlagIssuer        string = "issuer"
+	FlagRetireTo      string = "retire-to"
```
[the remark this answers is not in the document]
technicallyty: this was my bad, i should've updated the issue to use the jurisdiction terminology 🤦🏻 

also, we should add some docs to `Long` specifying that adding `--retirement-jurisdiction` flag will cause the credits to be retired to the `to` address, otherwise they will be sent as tradable

x/ecocredit/client/testsuite/tx.go
```diff
@@ -331,6 +331,88 @@ func (s *IntegrationTestSuite) TestTxCreateBatchCmd() {
...
+			args:      []string{"foo", "bar"},
+			expErr:    true,
+			expErrMsg: "Error: accepts 3 arg(s), received 2",
+		},
+		{
+			name:      "too many args",
+			args:      []string{"foo", "bar", "baz", "foobarbaz"},
+			expErr:    true,
+			expErrMsg: "Error: accepts 2 arg(s), received 3",
+		},
```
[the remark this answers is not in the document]
sfishel18: 🤦 thanks, fixed

x/ecocredit/client/tx.go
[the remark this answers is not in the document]
sfishel18: 👍 done and done!

x/ecocredit/client/testsuite/tx.go
```diff
@@ -331,6 +331,88 @@ func (s *IntegrationTestSuite) TestTxCreateBatchCmd() {
...
+		args      []string
+		expErr    bool
+		expErrMsg string
+	}{
+		{
+			name:      "missing args",
+			args:      []string{"foo"},
+			expErr:    true,
+			expErrMsg: "Error: accepts 3 arg(s), received 1",
+		},
```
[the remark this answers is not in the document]
sfishel18: removed

ryanchristo (approved):
Looks good to me. Thanks again for the contribution @sfishel18 ! :tada:

x/ecocredit/client/testsuite/tx.go
```diff
@@ -441,7 +517,7 @@ func (s *IntegrationTestSuite) TestTxSendCmd() {
 
 	for _, tc := range testCases {
 		s.Run(tc.name, func() {
-			cmd := coreclient.TxSendCmd()
```
ryanchristo, line 444: Ah. One more thing. We need to keep `TxSendCmd` and add `TxSendBulkCmd` rather than only adding `TxSendBulkCmd`. This will make both commands available to use.

technicallyty (approved):
lgtm - thanks!

PR #1086, opened 2022-05-04
refactor(x/ecocredit): reduce event fields and duplication

ryanchristo:
## Description

Closes: #1035

Removes unnecessary and redundant event fields. Also applies consistent use of present tense event names.

For basket `EventCreate`

- Deprecating curator because we can look up the curator with the basket denom and the creator of the basket is always assigned as the curator

For basket `EventPut`

- Deprecating credits because core `EventTransfer` (previously `EventReceive`) is emitted providing information about the sender, receiver, and credits transferred.

- Deprecating amount because sdk `coinbase` is triggered providing information about the denom and amount minted and sdk `transfer` event provides information about sender, receiver, and coins transferred.

For basket `EventTake`

- Deprecating credits because core `EventTransfer` (previously `EventReceive`) is emitted providing information about the sender, receiver, and credits transferred.

- Deprecating amount because sdk `burn` event is emitted providing information about the denom and amount burned and sdk `transfer` event provides information about sender, receiver, and coins transferred.

For core `EventReceive`

- Renamed to `EventTransfer`
- Updated sender and receiver to module address
- Removed `basket_denom` because the `sender` will be the module address in this case

For update events in general

- Removed all fields except for the unique identifier used to look up the item that was updated

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] followed the guidelines for [building modules](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules)
- [ ] included the necessary unit and integration [tests](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#testing)
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
[...]

PR #1224, opened 2022-06-28
feat(x/ecocredit): add bridge receive event and update bridge event

ryanchristo:
## Description

Ref: #1193 / Ref: #1206

This pull request adds `EventBridgeReceive` and updates `EventBridge` to include `amount`. Additional changes included after running `make proto-format` and `make tidy`.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/server/core/bridge.go
```diff
@@ -20,12 +20,15 @@ func (k Keeper) Bridge(ctx context.Context, req *core.MsgBridge) (*core.MsgBridg
...
-		Recipient: req.Recipient,
-		Contract:  req.Contract,
-	}); err != nil {
-		return nil, err
+	for _, credit := range req.Credits {
+		if err = sdkCtx.EventManager().EmitTypedEvent(&core.EventBridge{
+			Target:    req.Target,
+			Recipient: req.Recipient,
+			Contract:  req.Contract,
+			Amount:    credit.Amount,
```
technicallyty, line 28: im wondering if maybe we should put batch denom in here? thinking in terms of a service that's subscribed to this event, would likely want access to the denom to extract the underlying credit info

ryanchristo: We only bridge back to a contract that credits have been bridged from and we don't populate or update any information on the receiving end so as far as I understand we don't need to access any credit info.

ryanchristo: The main goal of storing the contract on the ledger was to prevent the need for additional queries. If there is credit info that we need, we should be including the info in the event, but I don't believe there is.

technicallyty: ahhh right, forgot it was only back to src 😸

technicallyty (approved):
lgtm!

proto/regen/ecocredit/v1/events.proto
```diff
@@ -193,4 +194,19 @@ message EventBridge {
...
   string contract = 3;
-}
\ No newline at end of file
+
+  // amount is the amount of credits.
+  string amount = 4;
+}
+
+// EventBridgeReceive is emitted when credits are bridged from another chain.
+message EventBridgeReceive {
```
clevinson, line 203: What's the use case for this event? I'm wondering if we should also add `amount` here too?

clevinson (approved):
lgtm. Only question is if we should also add amount to BridgeReceive. I'm not sure that `EventBridgeReceive` conveys anything meaningful with just project_id and batch_denom as the only fields.

proto/regen/ecocredit/v1/events.proto
ryanchristo: `amount` would be redundant with `EventMint`, which will be emitted from `Msg/CreateBatch` or `Msg/MintBatchCredits`. 

In general, we are taking a lean approach to event fields based on how events are written in the sdk and `EventBridge` is a special case because we want to make all necessary information easily accessible to the bridge service.

We would ideally include an event specific to each message so that we can track the execution of each message and the two fields here (project id and batch denom) would be the most relevant information to `Msg/BridgeReceive`.

See #1035 and #1086 for reference.

ryanchristo: If anything, we could probably remove project id but it still seemed relevant.

ryanchristo: If anything, we could probably remove project id but it still seems relevant.

PR #1225, opened 2022-06-29
refactor(x/ecocredit): add note and contract to origin tx

ryanchristo:
## Description

Ref: #1193 / Ref: #1206

This pull request adds note and contract to `OriginTx`. At the moment, nothing is done with contract on the server side, which is something we will store in state and handle in a followup. This pull request also refactors `OriginTx` validation to be a receiver method and moves to a separate file given `OriginTx` is no longer specific to `MsgMintBatchCredits`.

This pull request also removes note from `BatchOriginTx` (state object) because note was only intended as a field that would be passed to the events emitted by `MsgCreateBatch` and `MsgMintBatchCredits` (now included in `OriginTx`). Also `batch_denom` has been removed from `BatchOriginTx`, which serves no purpose.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/core/origin_tx.go
```diff
@@ -0,0 +1,30 @@
...
+func (o *OriginTx) Validate() error {
+	if !reOriginTxId.MatchString(o.Id) {
+		return sdkerrors.ErrInvalidRequest.Wrap("origin_tx.id must be at most 128 characters long, valid characters: alpha-numberic, space, '-' or '_'")
+	}
+
+	if !reOriginTxSource.MatchString(o.Source) {
+		return sdkerrors.ErrInvalidRequest.Wrap("origin_tx.source must be at most 32 characters long, valid characters: alpha-numberic, space, '-' or '_'")
+	}
+
+	if len(o.Contract) > 0 && !isValidEthereumAddress(o.Contract) {
```
technicallyty, line 21: the `isValidEthereum` regexp catches the `>0` condition, so could simplify here

technicallyty (approved):
lgtm, 1 nit

x/ecocredit/core/origin_tx.go
ryanchristo: So this is checking valid only if not empty. In some cases, we don't want contract to be required. For example, using mint batch credits when dynamically minting to a credit batch, which could be used as a manual bridge operation when bridging credits from a non-chain registry such as verra. https://github.com/regen-network/regen-ledger/pull/1206#discussion_r904122339

ryanchristo: We could make this required when a param is set to true (e.g. `contractRequired` or something) or if we make this always required we would only be supporting EVM-based chains wherever origin tx is being used.

technicallyty: ahhh right - i keep forgetting about the manual case 😖

clevinson (approved):
lgtm! Just one note that we should add credit class to `BatchOriginTx` table for security purposes, but happy to have that taken on in a follow-up.

proto/regen/ecocredit/v1/state.proto
```diff
@@ -284,10 +284,4 @@ message BatchOriginTx {
   // source is the source chain or registry of the transaction originating
   // the mint process (e.g. polygon, ethereum, verra).
   string source = 2;
-
-  // note is a reference note for accounting that is passed to an event.
-  string note = 3;
-
-  // batch_denom is the denom of the credit batch.
-  string batch_denom = 4;
```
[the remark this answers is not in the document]
ryanchristo: Make sense. Would prefer to use `class_key` for efficient storage but same idea and I agree.

PR #1230, opened 2022-07-02
chore: update codecov configuration

ryanchristo:
## Description

Closes: #1151

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] followed the guidelines for [building modules](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules)
- [ ] included the necessary unit and integration [tests](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#testing)
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed `!` in the type prefix if API or client breaking change
- [ ] confirmed all author checklist items have been addressed 
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

PR #1226, opened 2022-06-29
refactor(x/ecocredit): add batch contract mapping

ryanchristo:
## Description

Closes: #1127 / Ref: #1193 / Ref: #1206

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] followed the guidelines for [building modules](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules)
- [ ] included the necessary unit and integration [tests](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#testing)
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed `!` in the type prefix if API or client breaking change
- [ ] confirmed all author checklist items have been addressed 
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

PR #1231, opened 2022-07-03
fix(x/ecocredit): bridge and bridge receive message validation

ryanchristo:
## Description

Ref: https://github.com/regen-network/regen-ledger/pull/1230#discussion_r912325613 / Ref: https://github.com/regen-network/regen-ledger/pull/1226#discussion_r912388697 / Ref: https://github.com/regen-network/regen-ledger/pull/1226#discussion_r912388847

This pull request updates basic message validation implementation and tests for `MsgBridge` and `MsgBridgeReceive`. More strict requirements were created to ensure credits are properly bridged with all necessary information and tests have been updated to gherkin to make the requirements more clear (and because previous tests were broken and needed to be updated).

This pull request also breaks out ethereum address and transaction hash validation into package `types/eth`. Transaction hash validation was added to ensure id is a valid ethereum/polygon transaction hash when validating `MsgBridgeReceive`. We are less strict with basic origin tx validation in order to support manual operations from other sources.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/core/msg_bridge_receive_test.go
```diff
@@ -3,207 +3,58 @@ package core
...
+
+func (s *msgBridgeReceive) ExpectTheError(a string) {
+	require.EqualError(s.t, s.err, a)
+}
+
+func (s *msgBridgeReceive) ExpectNoError() {
+	require.NoError(s.t, s.err)
+}
+
+func (s *msgBridgeReceive) checkAndSetMockValues() {
```
technicallyty, line 46: im wondering if it might be better to make this type of behavior a step in the feature files? i.e.

```
GIVEN the message """some message..."
AND a project metadata string of length "X"
THEN....
```

fine either way though, just wanted to see how you feel about that approach 😄

technicallyty (approved):
good work! just one question/suggestion

x/ecocredit/core/msg_bridge_receive_test.go
ryanchristo: I like the idea. I also like having the field included in the json. I don't have strong preference. I was also thinking we might just want to include a string that exceeds the length but `foooooooooooooooooo...` looked a bit silly. I used lorum ipsum in another test but that was more fitting because it was a description for a basket. Happy to try out the suggestion here so I made the change and maybe we can reassess at another point. https://github.com/regen-network/regen-ledger/pull/1231/commits/550d76ba55597b25df8c48fe0612319231942ec0

x/ecocredit/core/features/msg_bridge.feature
```diff
@@ -0,0 +1,182 @@
+Feature: MsgBridge
+
+  Scenario: a valid message
+    Given the message
+    """
+    {
+      "owner": "cosmos1depk54cuajgkzea6zpgkq36tnjwdzv4afc3d27",
```
clevinson, line 7: Are we able to use `regen` prefix'd bech32 addresses here instead of cosmos ones?

clevinson (approved):
lgtm! One question abt bech32 addresses, but nonblocking.

x/ecocredit/core/features/msg_bridge.feature
ryanchristo: It's initialized at runtime so I don't think we can within the unit tests but that would be nice. I took a look around and could not find a solution so I opened an issue to further investigate: #1243.

technicallyty: i had this issue in the bridge repo, fixed it with this: https://github.com/rnd-inc/toucan-bridge-service/blob/ty/71-cosmos_client/regen/client.go#L21-L24

ryanchristo: Aha! Updated in the two unit tests here but we should update throughout in a separate pull request and move the config calls to the base suite of each package so that it applies to all unit tests. https://github.com/regen-network/regen-ledger/pull/1231/commits/974d53ef51f70d0f3d808660438e182f0b1c044e

PR #1248, opened 2022-07-08
fix(app): fix upgrade handler to initialize data module

aleem1314:
## Description

 -  Update upgrade handler to add `StoreUpgrades` for data module.
 -  Add issuance dates for new credit batches (for redwood).
 -  Add curator address for new baskets (for redwood).

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ecocredit/keys.go
```diff
@@ -19,8 +19,8 @@ const (
 
 var ModuleSchema = ormapi.ModuleSchemaDescriptor{
 	SchemaFile: []*ormapi.ModuleSchemaDescriptor_FileEntry{
-		{Id: 1, ProtoFileName: api.File_regen_ecocredit_v1_state_proto.Path()},
-		{Id: 2, ProtoFileName: basketapi.File_regen_ecocredit_basket_v1_state_proto.Path()},
+		{Id: 1, ProtoFileName: basketapi.File_regen_ecocredit_basket_v1_state_proto.Path()},
```
aleem1314, line 22: changing the ORM table prefix is a breaking change. Reverting the `basket` prefix to `1`.

app/stable_appconfig.go
```diff
@@ -112,6 +115,22 @@ func (app *RegenApp) registerUpgradeHandlers() {
 
 		return toVersion, nil
 	})
+
+	upgradeInfo, err := app.UpgradeKeeper.ReadUpgradeInfoFromDisk()
```
aleem1314, line 119: [#9381](https://github.com/cosmos/cosmos-sdk/issues/9381).

x/ecocredit/keys.go
ryanchristo: A breaking change is ok for v4.0. Was this causing issues with the migration?

aleem1314: Yes. We'll lose all of the basket state.

ryanchristo: I see. Tested without the change and basket state is empty after the migrations and core state is corrupted including both core state and basket state. Would be nice if we could rename using `Renamed` in `StoreUpgrades` and keep the change here but it looks like that might be more trouble than it's worth.

x/ecocredit/migrations/v3/patch_test.go
```diff
@@ -535,6 +547,384 @@ func TestRedwoodMigrations(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	startDate, _ = time.Parse(time.RFC3339, "2018-05-07T18:07:46Z")
+	endDate, _ = time.Parse(time.RFC3339, "2024-06-07T18:07:53Z")
```
ryanchristo, line 551: A bit concerning that all these start and end dates do not have a 00:00:00 UTC time. We've discussed the possibility of enforcing this but not something we need to change in this pull request.

x/ecocredit/migrations/v3/patch_test.go
```diff
@@ -772,42 +1203,44 @@ func TestRedwoodMigrations(t *testing.T) {
...
+	// C01-20210909-20220101-003  ->  "2022-03-08T17:18:19Z"
+	assertBatchIssuanceDate(t, ctx, ss, "C01-001-18540707-19870212-001", "2022-06-28T18:10:53Z")
+	assertBatchIssuanceDate(t, ctx, ss, "C01-002-19900103-19990103-001", "2022-06-28T08:36:25Z")
+	assertBatchIssuanceDate(t, ctx, ss, "C01-001-20130604-20130605-002", "2022-06-30T07:16:34Z")
+	assertBatchIssuanceDate(t, ctx, ss, "C01-002-20140604-20200603-002", "2022-06-29T08:45:12Z")
+	assertBatchIssuanceDate(t, ctx, ss, "C01-001-20170101-20180101-003", "2022-03-30T07:46:01Z")
+	assertBatchIssuanceDate(t, ctx, ss, "C01-006-20190101-20191231-001", "2022-05-11T11:35:30Z")
+	assertBatchIssuanceDate(t, ctx, ss, "C01-001-20190101-20210101-029", "2022-02-14T09:07:25Z")
+	assertBatchIssuanceDate(t, ctx, ss, "C01-007-20190101-20210101-001", "2022-04-22T16:32:09Z")
+	assertBatchIssuanceDate(t, ctx, ss, "C01-001-20210909-20220101-030", "2022-03-08T17:18:19Z")
```
ryanchristo, line 1243: I'm not too concerned about asserting each issuance date is correct for redwood but it looks like we are only checking a handful here rather than all of them. Ok to leave as is.

ryanchristo (approved):
tACK without mainnet and redwood state. I was able to successfully create a credit class and a basket, perform the upgrade, query the credit class and basket, define a data resolver and then query the resolver. Here are my notes: https://hackmd.io/LO-Lm0SFRNGSyS20AwzHTw.

And thanks for adding the updates for redwood. It's not critical to have these but will be nice to have. Might be more before the official release but we can add these just before if need be.

x/ecocredit/migrations/v3/patch.go
```diff
@@ -119,10 +178,12 @@ func patchRedwood(ctx context.Context, ss api.StateStore,
 	// basket name to curator
 	// rNCT   -> regen1df675r9vnf7pdedn4sf26svdsem3ugavgxmy46
 	// NCT    -> regen1df675r9vnf7pdedn4sf26svdsem3ugavgxmy46
+	// TYLER  -> regen1yqr0pf38v9j7ah79wmkacau5mdspsc7l0sjeva
 
 	basketNameToCuratorMap := make(map[string]string)
 	basketNameToCuratorMap["rNCT"] = "regen1df675r9vnf7pdedn4sf26svdsem3ugavgxmy46"
 	basketNameToCuratorMap["NCT"] = "regen1df675r9vnf7pdedn4sf26svdsem3ugavgxmy46"
+	basketNameToCuratorMap["TYLER"] = "regen1yqr0pf38v9j7ah79wmkacau5mdspsc7l0sjeva"
```
technicallyty, line 186: oops 🤣

ryanchristo:
Removed `!` from the type prefix because this is not an API or client breaking change.

Issue #1245, opened 2022-07-07
Update utility functions to check empty and return consistent errors


ryanchristo:
## Summary

Ref: https://github.com/regen-network/regen-ledger/pull/1233#discussion_r916139233

Currently we check for an empty string inside basic message validation for credit class, project id, batch denom, etc. We should check for empty string within the validation functions defined in `utils.go`. We should also make sure we are returning consistent errors for these validation functions, i.e. all should include the expected format and consistent sdk errors.

## Problem Definition

We are duplicating a lot of empty string checks in message validation and returning inconsistent errors and errors that could provide more information about the expected format.

## Proposal

Add empty string checks for validation functions in `utils.go` (for core, basket, and marketplace) and remove the empty string checks in basic message validation. Also make sure all errors returned include the expected format within an sdk error.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

wgwz:
@ryanchristo i'm interested in working on this one if you agree it could be a good one to work on.

ryanchristo:
Yea, I think this would be a good one. I'll assign to you and feel free to reach out with any questions.

wgwz:
Hey @ryanchristo I have opened up a draft PR that is a start on this. I think I'm headed in the right direction. So far I started with the core module and refactored the validation functions within `core/utils.go`. Let me know what you think, thanks!

ryanchristo:
@wgwz thanks again for tackling the largest chunk of work in #1317! I think there are only a couple left for basket and it would be nice to get those updated now that we've added state validation checks that use the same utils. Unless you've started, do you mind if I finish off those changes?

wgwz:
@ryanchristo np, please feel free, i'll be glad to help with review!

Issue #1147, opened 2022-06-01
Add guidelines on using sdk errors and module errors


ryanchristo:
## Summary

When to use what sdk errors is not clear and often we default to `ErrInvalidRequest`. We should have clear guidelines on using sdk errors and module specific errors. This could be included in the contributing guidelines (#1003) or in a separate document.

## Problem Definition

Error messages are inconsistent and guidelines on when to use what errors would help with writing better error messages and as a result improve the user experience.

## Proposal

Write guidelines on when to use what errors and then open a followup issue to update codebase according to those guidelines.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

aaronc:
This should likely get upstreamed to the SDK

aaronc:
@ryanchristo would you suggest modules don't use ErrInvalidRequest and define custom errors?

Or would it make sense if the SDK has a core set or general purpose errors like invalid request, unauthorized that modules are encouraged to use?

ryanchristo:
There are already guidelines in the sdk and it looks like we could do a better job of following them.

https://docs.cosmos.network/main/building-modules/errors.html

Specifically for the case I mentioned above, the sdk actually suggests something like:

```
	ErrEmptyName      = sdkerrors.Register(ModuleName, 2, "name is empty")
```

ryanchristo:
Note to include use of `sdkerrors.Wrap` in guidelines and to avoid stacking the same error. See https://github.com/regen-network/regen-ledger/pull/1229#discussion_r913141103.

ryanchristo:
Functions that can be used outside the context of a request should not use `ErrInvalidRequest` but it may make sense to use in basic message validation that makes use of the function. For example, we have utility functions for validating the format of a batch denom. The utility function that validates the string would return a parse error and the basic message validation that call the utility function would return an invalid request error that wraps the parse error:

```
func ValidateBatchDenom(denom string) error {
	if denom == "" {
		return ecocredit.ErrParseFailure.Wrap("batch denom cannot be empty")
	}
	matches := regexBatchDenom.FindStringSubmatch(denom)
	if matches == nil {
		return ecocredit.ErrParseFailure.Wrap("invalid batch denom: expected format A00-000-00000000-00000000-000")
	}
	return nil
}
```

And when being used within basic message validation:

```
[...]

PR #1317, opened 2022-07-29
refactor(x/ecocredit): move empty string validation into utils.go

wgwz:
## Description

Ref: https://github.com/regen-network/regen-ledger/issues/1245

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/ecocredit/basket/features/msg_put.feature
+++ b/x/ecocredit/basket/features/msg_put.feature
@@ -79,7 +79,7 @@ Feature: MsgPut
     }
     """
     When the message is validated
-    Then expect the error "credit batch denom cannot be empty: invalid request"
+    Then expect the error "batch denom cannot be empty: parse error: invalid request"
 
   Scenario: an error is returned if a credit batch denom is not formatted
     Given the message
--- a/x/ecocredit/basket/msg_put.go
+++ b/x/ecocredit/basket/msg_put.go
@@ -39,10 +39,6 @@ func (m MsgPut) ValidateBasic() error {
 
 	if len(m.Credits) > 0 {
 		for _, credit := range m.Credits {
-			if len(credit.BatchDenom) == 0 {
-				return sdkerrors.ErrInvalidRequest.Wrap("credit batch denom cannot be empty")
-			}
-
 			if err := core.ValidateBatchDenom(credit.BatchDenom); err != nil {
 				return sdkerrors.ErrInvalidRequest.Wrap(err.Error())
 			}
--- a/x/ecocredit/core/features/msg_bridge_receive.feature
+++ b/x/ecocredit/core/features/msg_bridge_receive.feature
@@ -54,7 +54,7 @@ Feature: MsgBridgeReceive
     }
     """
     When the message is validated
-    Then expect the error "class id cannot be empty: invalid request"
+    Then expect the error "class id cannot be empty: parse error: invalid request"
 
   Scenario: an error is returned if class id is not formatted
     Given the message
@@ -114,7 +114,7 @@ Feature: MsgBridgeReceive
     }
     """
     When the message is validated
-    Then expect the error "project jurisdiction cannot be empty: invalid request"
+    Then expect the error "jurisdiction cannot be empty, expected format <country-code>[-<region-code>[ <postal-code>]]: parse error: invalid request"
 
   Scenario: an error is returned if project jurisdiction is not formatted
     Given the message
--- a/x/ecocredit/core/features/msg_retire.feature
+++ b/x/ecocredit/core/features/msg_retire.feature
@@ -77,7 +77,7 @@ Feature: MsgRetire
     }
     """
     When the message is validated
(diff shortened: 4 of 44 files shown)


x/ecocredit/basket/features/msg_put.feature
```diff
@@ -95,7 +95,7 @@ Feature: MsgPut
     }
     """
     When the message is validated
-    Then expect the error "invalid batch denom: expected format A00-000-00000000-00000000-000: parse error: invalid request"
+    Then expect the error "invalid batch denom: expected format A00-000-00000000-00000000-000: invalid request"
```
ryanchristo, line 98: Here's an example of what might be the preferred format but the deleted `parse error` might actually make sense to keep as the error in the utility function and the error returned within basic message validation would wrap the returned error with an `invalid request` error.

x/ecocredit/basket/msg_put.go
```diff
@@ -39,12 +39,8 @@ func (m MsgPut) ValidateBasic() error {
 
 	if len(m.Credits) > 0 {
 		for _, credit := range m.Credits {
-			if len(credit.BatchDenom) == 0 {
-				return sdkerrors.ErrInvalidRequest.Wrap("credit batch denom cannot be empty")
-			}
-
 			if err := core.ValidateBatchDenom(credit.BatchDenom); err != nil {
-				return sdkerrors.ErrInvalidRequest.Wrap(err.Error())
+				return err
```
ryanchristo, line 43: Here's an example of where we should wrap the error with `sdkerrors.ErrInvalidRequest`.

The error returned from the utility function should not already be wrapped with the `sdkerrors.ErrInvalidRequest` as the utility function may be used outside the context of a request and the parse error might b the preferred error to return in the utility function, which is then wrapped here with `sdkerrors.ErrInvalidRequest`.

x/ecocredit/core/features/msg_bridge_receive.feature
```diff
@@ -129,7 +129,7 @@ Feature: MsgBridgeReceive
     }
     """
     When the message is validated
-    Then expect the error "invalid jurisdiction: foo, expected format <country-code>[-<region-code>[ <postal-code>]]: parse error: invalid request"
+    Then expect the error "invalid jurisdiction: foo, expected format <country-code>[-<region-code>[ <postal-code>]]: invalid request: invalid request"
```
ryanchristo, line 132: In line with my comments above, I think the previous error is the desired error message here, which would include the parse error from the utility function and the invalid request error from basic message validation.

ryanchristo (commented):
Thanks for the contribution! Looks like a good start.

The issue was not very specific so my apologies. Although the parse error that was previously being used is defined in the ecocredit module, it is registered as an sdk error.

```
ErrParseFailure        = sdkerrors.Register(ModuleName, 2, "parse error")
```

I think this error still makes sense when returning an error for an invalid format, and this error would be preferred over using `sdkerrors.ErrInvalidRequest` within the utility functions given the utility functions may be used outside the context of a request.

I've added another note to an issue proposing guidelines for better use of sdk errors, which is also being used to track examples (https://github.com/regen-network/regen-ledger/issues/1147#issuecomment-1199793577).

wgwz:
hey @ryanchristo 👋 i made another pass at this and pushed a new commit. i have still only done the refactoring work for `x/ecocredit/core/utils.go`. it should be closer now, LMK what you think, thanks!

x/ecocredit/core/features/msg_retire.feature
```diff
@@ -127,4 +127,21 @@ Feature: MsgRetire
...
+        {
+          "batch_denom": "C01-001-20200101-20210101-001",
+          "amount": "100"
+        }
+      ],
+      "jurisdiction": ""
+    }
+    """
+    When the message is validated
+    Then expect the error "jurisdiction cannot be empty, expected format <country-code>[-<region-code>[ <postal-code>]]: parse error: invalid request"
```
ryanchristo, line 147: This is a duplicate of `Scenario: an error is returned if jurisdiction is empty`. If jurisdiction is not provided, the zero-value is an empty string, which is why the error is the same here as it is above.

```suggestion
```

ryanchristo (commented):
This looks good to me. The added scenario is a duplicate but otherwise I think this is ready if you want to open a separate pull request to address basket and marketplace utils.

x/ecocredit/core/features/msg_retire.feature
wgwz: Ah ok, I see. I have removed this.

wgwz:
@ryanchristo I do like the idea of merging this and then addressing the other locations in separate PRs. because i may not have the time this week to handle the others. should we open new separate issues? or favor keeping #1245 open and adding a todo list?

ryanchristo (approved):
Looks good to me. Thanks for the contribution!

technicallyty (approved):
👍🏻

PR #1334, opened 2022-08-03
ci: cleanup github actions

aleem1314:
## Description

* Removed experimental build and test actions.
* Update tests to run only when go code changes.
* Update noncritical Simulation actions to corn job.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


.github/workflows/sims-cron.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          go-version: 1.18
+      - name: Display go version
+        run: go version
+      - uses: actions/cache@v2.1.3
+        with:
+          path: ~/go/bin
+          key: ${{ runner.os }}-go-runsim-binary
+      - name: sim-fullapp
+        run: |
```
aleem1314, line 69: Maybe we should add a discord alert for these actions.

.github/workflows/sims-cron.yml
```diff
@@ -0,0 +1,70 @@
+name: Sims Cron Job
```
ryanchristo, line 1: Maybe we could name this `Sims Nightly` and the file `sims-nightly.yml` to be more specific?

.github/workflows/test.yml
```diff
@@ -7,48 +7,31 @@ on:
...
-      - name: Install Go
-        uses: actions/setup-go@v2.1.3
+        - uses: technote-space/get-diff-action@v6.0.1
+        id: git_diff
         with:
-          go-version: 1.18
+          PATTERNS: |
+            **/**.go
+            go.mod
+            go.sum
```
ryanchristo, line 23: I like the idea here but we should make sure tests are still required and this prevents us from doing so. Maybe this could be used for some of the other jobs such as sims (the one's that are not nightly) ?

ryanchristo (commented):
Looking good. I removed the required checks for experimental in settings but we should still make tests required so maybe the diff check would be useful elsewhere? Maybe for the other sims that are not run nightly? It looks like they are still running for this pull request.

We may also want to consider updating `test.mk` to better separate unit tests and integration tests (both of which should still be required but we might be able to speed things up by running them separately). That can wait for another pull request though.

.github/workflows/sims-cron.yml
ryanchristo: I'll look into this and see if we can get that set up. :+1:

.github/workflows/test.yml
aleem1314: > I like the idea here but we should make sure tests are still required and this prevents us from doing so. Maybe this could be used for some of the other jobs such as sims (the one's that are not nightly) ?

This change does not block required checks. If there are no changes in go code, this action skips `sub-modules test and codecov`, and moves to the next step.

ryanchristo (approved):

Issue #1237, opened 2022-07-05
Add event and event value checks to unit tests


ryanchristo:
## Summary

We currently have some values that are specifically passed to events and not stored in state (e.g. retirement jurisdiction, cancel reason, bridge note, etc.). It would be nice to test expected events are properly emitted and include expected values.
 
## Problem Definition

We currently have values passed to events and we are not testing if they are properly set.

## Proposal

Add checks to make sure all expected events are emitted and each event includes expected values.

- [ ] Core Submodule - https://github.com/regen-network/regen-ledger/pull/1394
- [ ] Basket Submodule - https://github.com/regen-network/regen-ledger/pull/1408
- [ ] Market submodule - https://github.com/regen-network/regen-ledger/pull/1410
- [ ] Data module - https://github.com/regen-network/regen-ledger/pull/1414

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1414, opened 2022-08-25
test(x/data): add event tests

technicallyty:
## Description

- adds event checks to x/data tests

Closes: #1237 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/data/server/features/msg_define_resolver.feature
```diff
@@ -14,3 +14,16 @@ Feature: Msg/DefineResolver
...
+    Scenario: EventDefineResolver is emitted
+      Given alice has defined a resolver with url "https://foo.bar"
+      When bob attempts to define a resolver with url "https://foo.bar"
+      Then expect the resolver with id "2" and url "https://foo.bar" and manager bob
+      And expect event with properties
+      """
+      {
+        "id": 1
+      }
+      """
```
technicallyty, line 29: this seems.. wrong? but this is how it passes 🤔

ryanchristo: It looks like it's reading the event from here:

```
func (s *defineResolverSuite) AliceHasDefinedAResolverWithUrl(a string) {
	_, err := s.server.DefineResolver(s.ctx, &data.MsgDefineResolver{
		Manager:     s.alice.String(),
		ResolverUrl: a,
	})
	require.NoError(s.t, err)
}
```

ryanchristo: How we write tests have evolved since these initial tests. This step should probably be a resolver exists with manager, etc. and we would insert a resolver rather than call the define resolver server method.

ryanchristo:
```feature
      Given a resolver with properties
     """
     ...
     """
      When bob attempts to define a resolver with url "https://foo.bar"
```

technicallyty: ahhh i see now, thanks!

types/testutil/events.go
```diff
@@ -25,7 +25,8 @@ func MatchEvent(expected proto.Message, emitted sdk.Event) error {
 // the proto.MessageName of the passed msg.
 func GetEvent(msg proto.Message, events []sdk.Event) (e sdk.Event, found bool) {
 	eventName := proto.MessageName(msg)
-	for _, e := range events {
+	for i := len(events) - 1; i >= 0; i-- {
+		e = events[i]
```
ryanchristo, line 29: I actually think we should get the first event here and update the test to insert into the table on any `Given` steps. The use of the server method for a `Given` step is outdated and the reason why it was failing. The `When` step is what we are testing and `Given` should set up the scenario for the `When`.

ryanchristo: Might also be good to add a comment to this method regarding getting the first event found.

ryanchristo: Another solution would be to reset the context, which is what I did in the bridge integration tests, but I still think we should get the first event and insert on `When`.

technicallyty: the test step was updated to insert here - [fix: use insert](https://github.com/regen-network/regen-ledger/pull/1414/commits/5b06367b45af420b5f184ce3a225848f24ccd0f5)

I updated the `GetEvent` afterwards as it seemed more natural to search for events by most recent 🤷🏻 but no strong feeling either way

ryanchristo (approved):
LGTM

Issue #1328, opened 2022-08-01
Add interchain accounts module and migrations


ryanchristo:
## Summary

We will be updating ibc-go from v2 to v5 (#1327) in the next release and v3 adds support for interchain accounts. We need to decide on the parameters we want to set and add the necessary state migrations to support interchain accounts.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ryanchristo:
Following #1441 the controller module will be tracked in a separate issue #1453.

Issue #1440, opened 2022-08-30
Improve upgrades and testing


technicallyty:
## Summary

We currently have no testing for upgrades. We should add tests for upgrades to better ensure the chain's resulting state and modules are changed accordingly (aside from doing so manually). 

## Problem Definition

We have no tests for upgrades

## Proposal

Consider adopting the [upgrade style of Osmosis](https://github.com/osmosis-labs/osmosis/tree/main/app/upgrades) and add tests similarly.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1441, opened 2022-08-30
feat(app)!: enable ICA Host module

technicallyty:
## Description

- adds interchain accounts

~~TODO: do we want to setup the parameters manually in the upgrade, allowing us to enable ICA right off the bat with some predefined allowed messages, or should we it be initialized as an empty module and let governance enable/add messages?~~

governance can enable and set the messages to be available for host chain execution.

Additionally, some tests/refactoring can be added to ensure this works in #1440 

Closes: #1328 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


app/simulation/utils.go
```diff
@@ -50,3 +51,21 @@ func simulateFromSeed(t *testing.T, app *regen.RegenApp, config simtypes.Config)
...
+				return append(mods[:i], mods[i+1:]...)
+			}
+		}
+		return mods
+	}
+
+	icaModName := ica.AppModule{}.Name()
+
+	app.ModuleManager.OrderExportGenesis = remove(icaModName, app.ModuleManager.OrderExportGenesis)
+}
```
technicallyty, line 71: unfortunately ICA module does not implement `AppModuleSimulation` so the module does initialize with any state during simulations. This in turn causes `ExportGenesis` calls on the `regen.App` to panic due to a bug described here https://github.com/cosmos/ibc-go/issues/2151. 

This is a temporary workaround until the above issue is resolved on the IBC-go repo, or a more preferable solution is found.

app/app.go
```diff
@@ -406,6 +432,11 @@ func NewRegenApp(logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest
 		ecocreditMod,
 		dataMod,
 		groupmodule.NewAppModule(appCodec, app.GroupKeeper, app.AccountKeeper, app.BankKeeper, app.interfaceRegistry),
+		// TODO: decide if we want to be a controller chain. For now, we only setup the host.
+		//	this means actions can be executed on Regen Ledger from another chain.
+		// 	however, until we add the controller, Regen Ledger accounts will not be able to execute messages
+		// 	on other chains.
```
ryanchristo, line 438: We should wire up the controlled and make sure it is disabled. This way it could be enabled via governance:

https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/types/params.go#L9-L12

ryanchristo: Currently the command to query the controller is available but it produces the following error:

```
Error: rpc error: code = Unknown desc = value method github.com/cosmos/ibc-go/v5/modules/apps/27-interchain-accounts/controller/keeper.Keeper.Params called using nil *Keeper pointer: panic
```

app/simulation/utils.go
ryanchristo: Work around for now works for me.

app/stable_appconfig.go
```diff
@@ -36,6 +37,7 @@ func (app *RegenApp) registerUpgradeHandlers() {
 		storeUpgrades := storetypes.StoreUpgrades{
 			Added: []string{
 				group.ModuleName,
+				icahosttypes.StoreKey,
```
ryanchristo, line 40: We should probably have the host disabled upon upgrade and require a governance proposal to enable it.

ryanchristo (commented):
Nice! [Tested adding a message via governance and querying params](https://hackmd.io/helrxk0VRyaFA0JaBf2QHQ). We should have host disabled on upgrade and require a governance proposal to enable it. Would be nice to wire up the controller here as well and make sure it is also disabled upon upgrade.

app/stable_appconfig.go
technicallyty: do you mean to edit the upgrade handler to directly set the host params to false? I don't think we can remove the key in the commented line there

in any case, the host submod will be disabled either way, however disabling it manually in the upgrade handler will at least make it's params queryable post-upgrade. wdyt, edit the handler or leave as is?

ryanchristo: I just left the comment here because it was the only line changed. I do mean setting the host param to false.

ryanchristo: Edit the handler. It will be queryable either way as long as the module is wired up. Currently the upgrade will set the host to enabled by default. We should require a governance proposal to enable it before adding messages.

app/app.go
technicallyty: yeah the controller is a lot more tricky than the host. it needs middleware, and in the demo i linked in discord, it appears the thing chain devs should be doing is creating their own `x/intertx` module (the one in the demo says its experimental and not to use it in prod) ? but it also looks like it can be wired up using a new `IBCFee` module. 

the docs are also a bit confusing for chain devs, as the [integration guide](https://github.com/cosmos/ibc-go/blob/main/docs/apps/interchain-accounts/integration.md) mentions an ICAAuth module that can be used for the controller's middleware, but that module doesn't appear to exist ??

we may just want to start w/ host for now and wait for them to update the docs to see what chaindevs should do for integrating controllers 🤷🏻

app/stable_appconfig.go
technicallyty: ah i missed the default genesis setting it to true, i'll set the params in the upgrade handler 👍🏻

app/app.go
ryanchristo: Can we make this an issue rather than a todo? No need for the todo here but an issue to track would be great.

ryanchristo: We can throw it in the backlog. Probably not something we need for v5.0.

technicallyty: issue: https://github.com/regen-network/regen-ledger/issues/1453

commented in code: https://github.com/regen-network/regen-ledger/pull/1441/commits/bf6f818dbd8b87c8d114cca2d88d171acec213d7

technicallyty:
****

> [tACK](https://hackmd.io/helrxk0VRyaFA0JaBf2QHQ). Unfortunately the error still exists since when querying controller params because the keeper is not implemented. Is there a way around this?

I don't think so. all the cmd's are managed by the module itself. might just have to live with it until the ICA module matures 🤷🏻

ryanchristo (approved):
Meant to approve with the tACK above. Would be nice to get the controller set up but start with it disabled if we find time to dig in further or when the documentation is more up-to-date.

Issue #1577, opened 2022-10-26
Use standard error codes and fix unknown error codes


ryanchristo:
## Summary of Bug

Related to #1147

Some error codes are showing up as `Unknown` and others are not the expected standard for the type of error thrown. We should make sure all errors have an error code and error codes reflect the standard error codes.

Issue originally reported in Discord by Mohammed from Shamba.

## Steps to Reproduce

> We observed that the status on the error is always UNKNOWN even for scenarios where a NOT_FOUND or INVALID_ARGUMENT might be more appropriate.
>
> Take this example request:
> grpcurl -plaintext -d '{"iri":"regen:13toVgf5aZqSVSeJQv562xkkeoe3rr3bJWa29PHVKVf77VAkVMcDvVd.rdf"}' redwood.regen.network:9090 regen.data.v1.Query/AnchorByIRI
>
> The request yields the error:
> Code: Unknown
> Message: data record with IRI: not found
>
> While the message hints that this is a NOT_FOUND error, the code does not reflect that. I think it would be more appropriate if the code returned was gRPC status code 5 rather than code 2 
>
> In another example, let's tweak the request to provide a malformed IRI:
> grpcurl -plaintext -d '{"iri":"invalid-iri.rdf"}' redwood.regen.network:9090 regen.data.v1.Query/AnchorByIRI
>
> This request yields the error:
> Code: Unknown
 > Message: failed to parse IRI invalid-iri.rdf: regen: prefix required: invalid IRI
>
> Once again, while the error message hints of an INVALID_ARGUMENT error, the code returned is gRPC status code 2 rather than 3 which would be more appropriate
>
> And also, to give more context on our implementation, the reason why we are very interested in the error codes is because we want to react to errors the right way. This would give us the ability to send simplified messages to the end user who might not be very interested in technical error messages.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1579, opened 2022-10-27
fix(x/data): return grpc error codes for queries

aleem1314:
## Description

Ref: #1577

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


technicallyty (commented):
I think the preferred path forward with the new `cosmossdk.io/errors` package is to have consumer chains define their own common errors. 

It might be worthwhile to open a new package `errors` and define some common errors, registered with the package above. we can avoid writing the `status.Error(codes.InvalidArgument, "blah blah")` over and over again, and instead just do `ourErrorsPackage.InvalidArgument.Wrap("blah blah")`. 

wdyt?

aleem1314:
> I think the preferred path forward with the new `cosmossdk.io/errors` package is to have consumer chains define their own common errors.
 It might be worthwhile to open a new package `errors` and define some common errors, registered with the package above. we can avoid writing the `status.Error(codes.InvalidArgument, "blah blah")` over and over again, and instead just do `ourErrorsPackage.InvalidArgument.Wrap("blah blah")`. 
 wdyt?

Created `errors` package [a532ffb](https://github.com/regen-network/regen-ledger/pull/1579/files#diff-83d18c5448f32f4da3571b1244040e2501899455bd6383704906992a6a532ffbR1).

errors/grpc.go
```diff
@@ -0,0 +1,21 @@
...
+	ErrUnkown           = errors.RegisterWithGRPCCode(moduleName, 2, codes.Unknown, "unknown")
+	ErrInvalidArgument  = errors.RegisterWithGRPCCode(moduleName, 3, codes.InvalidArgument, "invalid argument")
+	ErrNotFound         = errors.RegisterWithGRPCCode(moduleName, 4, codes.NotFound, "not found")
+	ErrAlreadyExists    = errors.RegisterWithGRPCCode(moduleName, 5, codes.AlreadyExists, "already exists")
+	ErrPermissionDenied = errors.RegisterWithGRPCCode(moduleName, 6, codes.PermissionDenied, "permission denied")
+	ErrInternal         = errors.RegisterWithGRPCCode(moduleName, 7, codes.Internal, "internal")
+	ErrUnavailable      = errors.RegisterWithGRPCCode(moduleName, 8, codes.Unavailable, "unavailable")
+	ErrUnauthenticated  = errors.RegisterWithGRPCCode(moduleName, 9, codes.Unauthenticated, "unauthenticated")
+	ErrUnimplemented    = errors.RegisterWithGRPCCode(moduleName, 10, codes.Unimplemented, "unimplemented")
+)
```
ryanchristo, line 21: Where are these error codes coming from? It looks like the expected not found error should be status code 5 rather than 4 according to this: https://grpc.github.io/grpc/core/md_doc_statuscodes.html. This is also the expected error code that was reported by the user (in the issue summary).

ryanchristo: Also see https://pkg.go.dev/google.golang.org/grpc/codes. I think we should be using these error codes rather than defining our own... Looks like there is some confusion around the issue per @technicallyty's above request.

ryanchristo (changes requested):
> I think the preferred path forward with the new cosmossdk.io/errors package is to have consumer chains define their own common errors.

The goal of this issue was to use the standard gRPC code rather than defining custom errors with different codes. Developers wanting to use their own custom error messages will most likely be writing custom error messages based on the standard gRPC error codes (which is the case for the developers that originally reported this message).

Apologies for the delay on reviewing and the confusion. Maybe we can further discuss on our call tomorrow.

aleem1314:
> > I think the preferred path forward with the new cosmossdk.io/errors package is to have consumer chains define their own common errors.
> 
> The goal of this issue was to use the standard gRPC code rather than defining custom errors with different codes. Developers wanting to use their own custom error messages will most likely be writing custom error messages based on the standard gRPC error codes (which is the case for the developers that originally reported this message).

The current PR fixes the issue related to gRPC status codes. We were getting `Unknown` status because queries are not returning valid gRPC errors.

response on `main` or `v4.1x` branch

```
$ grpcurl -plaintext -d '{"iri":"invalid-iri.rdf"}' redwood.regen.network:9090 regen.data.v1.Query/AnchorByIRI

ERROR:
  Code: Unknown
  Message: failed to parse IRI invalid-iri.rdf: regen: prefix required: invalid IRI

``` 

In this PR

```
$ grpcurl -plaintext -d '{"iri":"invalid-iri.rdf"}' localhost:9090 regen.data.v1.Query/AnchorByIRI

ERROR:
  Code: InvalidArgument
  Message: codespace grpc code 3: invalid argument: failed to parse IRI: failed to parse IRI invalid-iri.rdf: regen: prefix required: invalid IRI
```

errors/grpc.go
aleem1314: > Where are these error codes coming from? It looks like the expected not found error should be status code 5 rather than 4 according to this: https://grpc.github.io/grpc/core/md_doc_statuscodes.html. This is also the expected error code that was reported by the user (in the issue summary).

These error codes are just to make sure we don't have duplicate errors in each codespace. Nothing to do with gRPC status codes. For gRPC errors, we are using the third argument (`codes.Canceled`).

ryanchristo:
I'm still a bit confused why we are registering different error codes. I realize the correct error code is returned in the following example but we are displaying the custom registered error code in the message (a different error code):


```
$ grpcurl -plaintext -d '{"iri":"regen:13toVgf5aZqSVSeJQv562xkkeoe3rr3bJWa29PHVKVf77VAkVMcDvVd.rdf"}' localhost:9090 regen.data.v1.Query/AnchorByIRI

ERROR:
  Code: NotFound
  Message: codespace grpc code 4: not found: data record with IRI: regen:13toVgf5aZqSVSeJQv562xkkeoe3rr3bJWa29PHVKVf77VAkVMcDvVd.rdf
```

ryanchristo:
It would make more sense to me if we were using the same error codes but I'll try to do some digging on best practices.

aleem1314:
> It would make more sense to me if we were using the same error codes but I'll try to do some digging on best practices.

Updated error codes to match gRPC error.

errors/grpc.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"cosmossdk.io/errors"
+	"google.golang.org/grpc/codes"
+)
+
+const moduleName = "grpc"
+
+var (
+	ErrCancelled         = errors.RegisterWithGRPCCode(moduleName, 1, codes.Canceled, "cancelled")
+	ErrUnkown            = errors.RegisterWithGRPCCode(moduleName, 2, codes.Unknown, "unknown")
```
ryanchristo, line 12:
```suggestion
	ErrCanceled         = errors.RegisterWithGRPCCode(moduleName, 1, codes.Canceled, "canceled")
	ErrUnknown          = errors.RegisterWithGRPCCode(moduleName, 2, codes.Unknown, "unknown")
```

ryanchristo (approved):
LGTM

PR #1609, opened 2022-11-17
chore(deps) Dependency Bumps [chain-chores]

Reecepbcups:
Generated with [chain-chores](https://github.com/notional-labs/chain-chores)

Changes:
- cosmos/cosmos-sdk v0.46.4 -> cosmos/cosmos-sdk v0.46.5


ryanchristo:
@Reecepbcups can we opt out of this? We already have dependabot set up for all go modules within our repository. It looks like this only checks the root level gomod and it does not follow semantic commits. It also does not close itself when the changes have been merged. Good idea for other chains though!

Screenshot from 2022-11-18 08-06-49

Issue #1698, opened 2022-12-16
Go releaser failing with latest update to go 1.19


ryanchristo:
## Summary of Bug

https://github.com/regen-network/regen-ledger/actions/runs/3715110559/jobs/6299880767

```
release failed after 12m10s              error=failed to build for darwin_amd64_v1: exit status 2: go: downloading github.com/99designs/go-keychain v0.0.0-20191008050251-8e49817e8af4
```

## Version

`v5.0.0-rc2`

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

clevinson:
Verified that build works locally on both my M1 Mac (which is to be expected), and my old intel Mac (which should be darwin_amd64).  So it looks to be an issue with building for darwin from within a linux environment.

clevinson:
So the issue seems to be independent of whether or not we use cosmos fork of 99designs/keyring (addressed in #1709)

I spent a lot of time digging into this and think it's some problem with how we are doing cross compiling from linux to darwin amd64.  The way we do cross compiling in goreleaser is pulling in a custom fork of osxcross (see [here](https://github.com/regen-network/regen-ledger/blob/main/scripts/release-setup.sh#L11)).

I tried locally updating likhita's fork of osxcross to pull in the latest files from upstream, but that didn't seem to work either.

Today I found this project [goreleaser-cross](https://github.com/goreleaser/goreleaser-cross), which has a prebuilt docker container with osxcross. I think our best bet may be to refactor our goreleaser setup to use this container instead (it seems Akash does the same actually).

If we merge in #1714 for v5.0, then the full resolution of this issue is no longer blocking our upgrade.

ryanchristo:
Reassigning to @aleem1314 to look into setting up goreleaser-cross per our discussion this morning.

PR #1709, opened 2022-12-21
build: update 99designs/keyring to use cosmos fork

clevinson:
## Description

Ref: #1698 

I was finally able to get the issue in #1698 reproducing locally (running https://github.com/nektos/act again).

Unfortunately, just running `act` on its own wouldn't quite work properly. Running the `goreleaser` action requires GitHub Tokens and will push binaries to GitHub release's API, etc.

Instead I ran:
- `act -j release --container-architecture linux/amd64 -v`
- when the process failed (due to missing GitHub tokens), I shelled into the docker container with: `docker exec -it ${container_id} /bin/bash`
- ran `ln -s /opt/hostedtoolcache/go/1.19.4/x64/bin/go /usr/bin/go` to add the go binary in my PATH
- ran ` /opt/hostedtoolcache/goreleaser-action/1.13.1/x64/goreleaser build --skip-before  --skip-validate --help -id regen-darwin-amd64` to build the failing darwin binary

This successfully reproduced the error on goreleaser (`ld` having an unrecognized flag `-dynamic`).

After digging some more, I found https://github.com/cosmos/cosmos-sdk/pull/12952 which seemed to contain some related fixes in the fork that looked similar to changes we made in our go 1.19 related patches:

https://github.com/cosmos/keyring/pull/2/files#diff-33f8fbc61d6878200e61e69725cd724475a0c9dbb73bc63fcfc913a5651764feL175-L176

I thought that I got the goreleaser building successfully after updating the 99designs/keyring replace directive (as in this PR), but on subsequent runs it seems to still be failing...

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/scripts/release-setup.sh
+++ b/scripts/release-setup.sh
@@ -2,10 +2,10 @@
 
 # this script should only be called within github actions running goreleaser
 
-sudo apt-get update
-sudo apt-get install clang
-sudo apt-get install gcc-multilib g++-multilib
-sudo apt-get install gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64
+sudo apt-get update -y
+sudo apt-get install clang -y
+sudo apt-get install gcc-multilib g++-multilib -y
+sudo apt-get install gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64 -y
 
 mkdir /home/runner/work/osxcross
 git clone https://github.com/likhita-809/osxcross-target.git /home/runner/work/osxcross/target
--- a/go.mod
+++ b/go.mod
@@ -187,3 +187,5 @@ replace github.com/regen-network/regen-ledger/x/intertx => ./x/intertx
 replace github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.7.0
 
 replace github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
+
+replace github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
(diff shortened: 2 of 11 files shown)


scripts/release-setup.sh
```diff
@@ -2,10 +2,10 @@
...
 # this script should only be called within github actions running goreleaser
 
-sudo apt-get update
-sudo apt-get install clang
-sudo apt-get install gcc-multilib g++-multilib
-sudo apt-get install gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64
+sudo apt-get update -y
+sudo apt-get install clang -y
+sudo apt-get install gcc-multilib g++-multilib -y
+sudo apt-get install gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64 -y
```
clevinson, line 8: Adding these so that I could build goreleaser locally via `act`. Otherwise the build was hanging when running the GitHub actions locally.

ryanchristo: Do we still need these or was this for testing purposes?

ryanchristo (approved):
LGTM. One question before merging.

scripts/release-setup.sh
clevinson: It shouldn't have any affect on CI runs of goreleaser, but local testing via `act` will not work without these `-y` flags.

I like having them checked in to code so the goreleaser setup can be used for local testing as-is. If you're OK with keeping I'd like to just keep & merge.  If you prefer without then I can remove.

PR #2098, opened 2023-11-29
feat(x/data)!: support public resolvers for IPFS support

aaronc:
## Description

This PR:
* adds the `bool public` field to `MsgDefineResolver`
* renames `MsgDefineResolver` `manager` to `definer`
* renames `MsgRegisterResolver` `manager` to `signer`

These last two proto breaking changes are okay because we are currently working with the as of yet unreleased `regen.data.v2` API.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/data/v1/tx.proto
```diff
@@ -112,11 +112,23 @@ message MsgDefineResolver {
   // the manager should make use of cosmos.authz.
   string manager = 1;
```
clevinson, line 113: Who would be the `manager` for the `ipfs:` resolver?

Can you elaborate on how you would expect this to work?  In reading the documentation for the `manager` field, it seems like only the `manager` of a resolver can register data to a given resolver.

So as I understand it, this PR implies that each user wanting to host data on IPFS would define their own resolver named `ipfs:` ?

I wonder if it would make more sense to instead extend the `MsgRegisterResolver()` message to have an enum field called `resolver_type`.  we could have the default / 0 value of that enum be `HTTP_RESOLVER`, and 1 = `IPFS_RESOLVER`.  This way, if we want to support other decentralized resolvers that are similar to IPFS in the future, we could just extend this enum?

clevinson (commented):
@aaronc So I prematurely approved, and then after closer inspection actually have a few questions about how we expect this be used by users.

proto/regen/data/v1/tx.proto
aaronc: I think you're right. So, there's another step of work to do here. I don't think it will be too hard. I will close this, work on that and ping you when it's R4R again.

aaronc: @clevinson I've updated this PR to support marking a resolver as public to support IPFS. Please take a look again when you get a chance.

clevinson:
Reading through the updates, it looks like the approach you're intending for here is still that the defining of an IPFS resolver is something that is done in state (via someone defining an IPFS resolver via a `MsgDefineResolver()` tx).

So is your idea that someone from the RND team would submit a tx define a resolver for IPFS shortly after this software upgrade happens? Or perhaps we would just bake that into a state migration and ship it along w/ the software upgrade?

In either case though, I don't think we have a uniqueness constraint on resolver_url, so there's nothing preventing multiple users from creating new "IPFS" resolvers with the same `resolver_url` and `public = true`.  If and when this happens, for a user wanting to register their data to IPFS, they now would be presented with several different options of which resolvers to register with, all seemingly the same.

I'm just trying to unpack this here and make sure you're tracking this edge case and that you think this is OK @aaronc. Personally it feels a bit clunky and I'd prefer an approach where as a user wanting to register data to IPFS, there is one clear way to do that, and there's no chance of several different "resolvers" in state that all in practice mean identical things.  But if there are reasons why you thinks approach is better than what I wrote in my comment (https://github.com/regen-network/regen-ledger/pull/2098#discussion_r1513672745), which I think addresses the issue of end-users only having at most 1 way to register their data with IPFS, let me know.

aaronc:
There is a uniqueness check on the resolver,manager pair even when manager is nil in the case of being public. So I don't think double definitions are possible. I can add a test case to check

x/data/server/features/msg_define_resolver.feature
```diff
@@ -15,6 +15,11 @@ Feature: Msg/DefineResolver
       When alice attempts to define a resolver with url "https://foo.bar"
       Then expect the error "a resolver with the same URL and manager already exists: unique key violation"
 
+    Scenario: public resolvers can only be defined once per URL
+      Given a public resolver is defined for the url "ipfs:"
+      When alice attempts to define a public resolver with url "ipfs:"
+      Then expect the error "a resolver with the same URL and manager already exists: unique key violation"
```
aaronc, line 21: @clevinson I added a test here which demonstrates that a public resolver URL can only be defined once.

clevinson (approved):

proto/regen/data/v2/tx.proto
```diff
@@ -106,20 +106,39 @@ message MsgAttestResponse {
...
+  //
+  // If it refers to an HTTP URL, that HTTP service should
+  // respond to a GET request with the IRI of a ContentHash as the path parameter
   // and return the content if it exists or a 404. For graph data, resolvers
   // should use the HTTP Accept header to negotiate the RDF serialization
   // format.
+  //
+  // To use IPFS, the resolver_url ipfs: should be defined with public set to true
+  // and used as the resolver for any data hosted on IPFS. Content hashes must be
+  // adapted to IPFS's CID format. The multicodec raw (0x55) should
```
paul121, line 129: > Content hashes must be adapted to IPFS's CID format.

Just want to clarify how this works.. For both raw and graph content hashes, the `hash` will be set to the full `cid`, where the `cid` was generated using either the multicodec raw (0x55) or multicodec rdfc-1 (0xb403)?

paul121 (commented):
This LGTM, just one question to clarify how an IPFS CID would be used.

proto/regen/data/v2/tx.proto
clevinson: The content hash on Regen would still be the same as if it were registered w/ an HTTP resolver. I think this comment just communicates that users must take the resulting Regen IRI, and convert it to a specific IPFS CID in order to fetch the data from the IPFS network.

aaronc: Exactly

paul121: Ah thanks, this makes more sense. With an example I think this would be very clear. Hopefully soon...

paul121 (approved):

PR #2102, opened 2023-12-07
feat(x/data)!: enable off-chain coordination of supported algorithms

aaronc:
## Description

This PR introduces `regen.data.v2` which:
1. does not validate digest, canonicalization or merkle tree algorithms
2. does not use an enum for raw data extensions

This allows off-chain coordination of these values so that extensions can be made without requiring an upgrade.

`regen/data/v2/types.proto` SHOULD still be used as a registry of agreed upon algorithms, but this can be done between releases without requiring chain upgrades.

This PR deprecates the `regen.data.v1` proto files and bumps the x/data go module to v3.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


scripts/protocgen.sh
```diff
@@ -30,6 +30,4 @@ cd ..
 cp -r github.com/regen-network/regen-ledger/* ./
 rm -rf github.com
 
-go mod tidy
```
aaronc, line 33: This was breaking `make proto-gen` and can be done manually by the user

scripts/protocgen2.sh
```diff
@@ -5,7 +5,7 @@ set -eo pipefail
 protoc_gen_install() {
   go install github.com/cosmos/cosmos-proto/cmd/protoc-gen-go-pulsar@latest #2>/dev/null
   go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest #2>/dev/null
-  go install github.com/cosmos/cosmos-sdk/orm/cmd/protoc-gen-go-cosmos-orm@latest #2>/dev/null
+  go install github.com/cosmos/cosmos-sdk/orm/cmd/protoc-gen-go-cosmos-orm@v1.0.0-alpha.12 #2>/dev/null
```
aaronc, line 8: This was breaking `make proto-gen`

x/data/iri_test.go
```diff
@@ -445,3 +638,12 @@ func TestParseIRI(t *testing.T) {
 		})
 	}
 }
+
+func TestName(t *testing.T) {
```
blushi, line 642: to remove?

aaronc:
> Looks good, just a few nits and pending lint errors
> 
> Should I just use version 2 moving forward in regen-server/iri-gen, in particular in my PR: https://github.com/regen-network/regen-server/pull/423

We have no idea how long this will take to go into production so for now we need to stick with v1.

clevinson:
Thinking about the long-term effects of a versioning name change like this, I'm wondering if we'd be better off with one of the following:
1. using the pre-existing versioning system given by the protocol paths (e.g. just bumping this to `proto/regen/data/v2..`)
2. Updating the existing implementation of the current `Graph` and `Raw` types (in `promo/regen/data/v1...`) to match the desired implementation (what this PR current does in `GraphV2` and `RawV2`).

Conceptually I'm in favor of this general approach (not validating enums on-chain to enable for faster developing & off-chain coordination).  But given that we have such little usage of the data module to date, I think we may be able to skirt around what would otherwise be the strictly correct process for API upgrades, in favor of something that's a bit cleaner for future users.

aaronc:
> Thinking about the long-term effects of a versioning name change like this, I'm wondering if we'd be better off with one of the following:
> 
> 1. using the pre-existing versioning system given by the protocol paths (e.g. just bumping this to `proto/regen/data/v2..`)
> 2. Updating the existing implementation of the current `Graph` and `Raw` types (in `promo/regen/data/v1...`) to match the desired implementation (what this PR current does in `GraphV2` and `RawV2`).

I would be open to either approach. I've checked data module events on mainnet and so far there is no usage. So until we start using it (which is soon) breakage isn't really a concern. With approach 1) we would simply stop supporting v1 I think to keep things simple. Approach 2) is maybe okay since there are no other users but it is a bit un-kosher. Don't have any really strong opinions between the two though. @blushi any thoughts?

blushi:
> > Thinking about the long-term effects of a versioning name change like this, I'm wondering if we'd be better off with one of the following:
> > 
> > 1. using the pre-existing versioning system given by the protocol paths (e.g. just bumping this to `proto/regen/data/v2..`)
> > 2. Updating the existing implementation of the current `Graph` and `Raw` types (in `promo/regen/data/v1...`) to match the desired implementation (what this PR current does in `GraphV2` and `RawV2`).
> 
> I would be open to either approach - and I think this would greatly simply the implementation. I've checked data module events on mainnet and so far there is no usage. So until we start using it (which is soon) breakage isn't really a concern. With approach 1) we would simply stop supporting v1 I think to keep things simple. Approach 2) is maybe okay since there are no other users but it is a bit un-kosher. Don't have any really strong opinions between the two though. @blushi any thoughts?

I think I'd rather go with approach 1) because v1 hasn't been explicitly used on mainnet but it has been in our implementation within the regen server for generating IRIs for credit classes, projects and batches metadata so that would feel cleaner to have a proper version upgrade.

aaronc:
Okay. I'll update this PR to change the proto version to v2 and drop support for v1, unless there are any objections

aaronc:
@clevinson @blushi I've updated this to v2 (and the go.mod to v3) and it's R4R again. The diff is much larger because of breaking changes, but the code changes are actually somewhat simpler.

x/data/iri.go
```diff
@@ -43,18 +45,15 @@ func (chr ContentHash_Raw) ToIRI() (string, error) {
 	bz[1] = byte(chr.DigestAlgorithm)
 	copy(bz[2:], chr.Hash)
 	hashStr := base58.CheckEncode(bz, iriVersion0)
```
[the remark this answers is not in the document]
aaronc: I think it's actually okay to keep the IRI version at 0. Nothing substantively changed in how the IRIs are constructed.

aaronc:
@blushi I've addressed all your comments, except that I'd prefer to keep the IRI on version 0 and updated the comments to reflect that.

x/data/iri.go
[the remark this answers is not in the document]
blushi: right, seems fine now that we've bumped to v2

x/data/types.go
```diff
@@ -26,27 +26,43 @@ func (ch ContentHash) Validate() error {
...
+	}
+
+	if extLen > 6 {
+		return sdkerrors.ErrInvalidRequest.Wrapf("file extension cannot be longer than 6 characters")
+	}
+
+	// check that ext is all lowercase or numeric
+	for _, c := range ext {
+		if c < '0' || c > '9' && c < 'a' || c > 'z' {
+			return sdkerrors.ErrInvalidRequest.Wrapf("file extension must be all lowercase or numeric")
```
blushi, line 48: can we add a test case for this in iri_test.go?

blushi (approved):
LGTM, pre-approving, pending one test case to add

x/data/types.go
aaronc: Added a case

aaronc:
@clevinson could you take another quick look at this?

clevinson (approved):
Sorry this took so long. lgtm!

PR #2147, opened 2024-01-27
chore(setup): upgrade go to 1.21

robert-zaremba:
## Description

* upgrades go to v1.2
* upgrades GitHub actions/setup-go to v5


aaronc:
Let's see if we can get ci to pass with the updated linter

aaronc (approved):

Issue #1323, opened 2022-07-29
Add specification for independent projects


ryanchristo:
## Summary

Add an RFC or well-specified protobuf API for functionality that would support independent projects (projects that can be created independently of a credit class and then later apply to a credit class to be approved by a credit class admin or issuer).

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2150, opened 2024-01-30
feat(x/ecocredit): independent project API

aaronc:
## Description

Closes: #1323 and replaces #1422.

Based on the current way in which project pages are evaluating in app.regen.network, there is a need for an on-chain representation of projects before they have been admitted into credit classes. This PR introduces an application system for a project to enter into a credit class.

The key assumptions this design makes are:
* a project can apply to one credit class at a time via a single issuer of the credit class
* a project must explicitly specify which issuer (registry agent?) they are submitting the application to
* an application can only have three evaluation statues: approved, changes requested and rejected
* a credit class in an existing credit class can apply to another one
* project ID's are created independently from credit classes

This is intended to be done in a non API breaking way.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/regen/ecocredit/v1/tx.proto
```diff
@@ -273,6 +311,128 @@ message MsgCreateProjectResponse {
   string project_id = 1;
 }
 
+// MsgCreateUnregisteredProject is the Msg/CreateUnregisteredProject request type.
+message MsgCreateUnregisteredProject {
+  option (cosmos.msg.v1.signer) = "creator";
+
+  // creator is the address of the account creating the project that will become
+  // the admin of the project upon creation.
+  string creator = 1;
```
S4mmyb, line 320: Do we want to name this `admin` instead of `creator` to remain consistent with the language we're using in our existing framework?

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -273,6 +311,128 @@ message MsgCreateProjectResponse {
...
+  // application_id is the identifier of the application to evaluate.
+  uint64 application_id = 2;
+
+  // evaluation is the evaluation of the application.
+  ApplicationStatus evaluation = 3;
+
+  // reason is any arbitrary string with a maximum length of 256 characters
+  // that includes or references the reason for the approving, requesting changes
+  // to, or rejecting the application.
+  string reason = 4;
```
S4mmyb, line 430: The assumption here is that if it's longer than 256 characters, they could always provide a link to a longer response?

S4mmyb:
Great work @aaronc ! Overall I think this looks good and it's exciting to see an independent projects PR. The key assumptions you made in the design are: 
- a project can apply to one credit class at a time via a single issuer of the credit class
- a project must explicitly specify which `issuer` (registry agent?) they are submitting the application to
- an application can only have three evaluation statues: approved, changes requested and rejected
- a credit class in an existing credit class can apply to another one
- project ID’s are created independently from credit classes

For the most part I agree with this framework, but I am curious about your reasoning for why they must specify the `issuer`. The counter argument I see for this is that if certain issuers for a credit class are known to be more lenient in their reviews than others. Obviously we wouldn't want this to be the case as there should be very clear processes in place where decisions to approve have minimal subjectivity. But there given the complexity of these projects and the nature of project applications (which have a lot of written justification) there is always the possibility some are more lenient than others or have differences of opinions. With that framing in mind, it would probably make more sense to design it such that the application goes to all issuers, and one issuer claims it. This design could also solve business if certain issuers have higher workloads than other or are unavailable to take on applications (if that's a problem). If we do take that framing, one approach would be to add in an additional status in the `ApplicationStatus` field which is `APPLICATION_STATUS_RECIEVED`, which indicates an issuer has received the application and is reviewing it. 

My other question around this is, what happens if an application is rejected. I think we want to allow for multiple applications on the same project, as a rejection doesn't always indicate that the project doens't meet the criteria. Rather it might be incomplete from a reporting standpoint or too premature. If the `issuer` needs additional info and the `ApplicationStatus` changes to `APPLICATION_STATUS_CHANGES_REQUESTED`, does that application close or is it still open (i.e. do they need to reapply)? If it's rejected, do we want to track previous applications against that `project_id` which issuers can reference against to see why it was previously rejected? 

Again, great job.

aaronc:
Thanks for the review @S4mmyb. Would it make sense to allow any issuer to request changes? Should there be some application period whereby any issuer can block an approval if they do so before say a 2 week time period?

Also should we consider allowing multiple credit classes per project?

aaronc:
The more I think about it, the more I think we should simply allow a project to be enrolled in multiple credit classes simultaneously. This would subsume the need for a separate API for pre-financing as in #2152 since these could all be modeled as credit classes themselves. It would also allow reuse of the credit class infrastructure including the anti-spam fee and basketing. Also, we've heard of potential cases of credit stacking. The main argument for not allowing multiple credit class enrollment seems to be more of an argument against stacking which although maybe correct is likely not something we should encode ideologically at the protocol call especially since pre-financing is a valid use case.

proto/regen/ecocredit/v1/tx.proto
aaronc: Yes

proto/regen/ecocredit/v1/tx.proto
aaronc: Sure

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -273,6 +311,128 @@ message MsgCreateProjectResponse {
...
+message MsgCreateUnregisteredProject {
+  option (cosmos.msg.v1.signer) = "creator";
+
+  // creator is the address of the account creating the project that will become
+  // the admin of the project upon creation.
+  string creator = 1;
+
+  // metadata is any arbitrary string with a maximum length of 256 characters
+  // that includes or references metadata to attach to the project.
+  string metadata = 2;
```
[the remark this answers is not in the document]
aaronc: It would be optional

S4mmyb:
> Thanks for the review @S4mmyb. Would it make sense to allow any issuer to request changes? Should there be some application period whereby any issuer can block an approval if they do so before say a 2 week time period?
> 
> Also should we consider allowing multiple credit classes per project?

I think we should have an `APPLICATION_STATUS_CHANGES_REQUESTED` tag if they are requesting changes that simply changes state when an issuer requests changes. But I don't think we should encode any time limits on-chain, at least for now. I think both in a project applying to an issuer or an issuer requesting changes there should be time limits or expectations, but I think those should be handled contractually via legal agreements.

aaronc:
So what if in the credit class project relationship, an issuer can at any time change the status between approved, changes requested and rejected? Also a project could unilaterally withdrawal from a credit class relationship.

proto/regen/ecocredit/v1/state.proto
```diff
@@ -381,3 +387,55 @@ message AllowedBridgeChain {
...
+  string metadata = 5;
+
+  // status is the status of the application. Note that accepted and rejected
+  // applications are removed from the table.
+  ApplicationStatus status = 6;
+}
+
+// Application represents the evaluation status that a credit class issuer
+// assigns to a credit class application.
+enum ApplicationStatus {
```
clevinson, line 424: So possibly we should add an `APPLICATION_STATUS_ASSIGNED` if we are treating applications as a first come first serve by the issuers?

Though in seeing this, I'm wondering if it overcomplicates the process and if we would be better off just choosing either:
- projects apply to a specific issuer
- project applies to the credit class, and any issuer can change the status at any point in time (e.g. issuer A can set it as 'change requested' and issuer B can later accept the project)

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -273,6 +311,128 @@ message MsgCreateProjectResponse {
...
+  string project_id = 2;
+
+  // class_id is the identifier of the credit class which the project is
+  // applying to.
+  string class_id = 3;
+
+  // issuer is the address of the account that is an issuer of the credit class
+  // to whom the project is applying to for approval.
+  string issuer = 4;
+
```
clevinson, line 365:
```suggestion
```

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -273,6 +311,128 @@ message MsgCreateProjectResponse {
...
+}
+
+// MsgCreateUnregisteredProjectResponse is the Msg/CreateUnregisteredProject response type.
+message MsgCreateUnregisteredProjectResponse {
+  // project_id is the unique identifier of the project.
+  string project_id = 1;
+}
+
+// MsgSubmitClassApplication is the Msg/SubmitClassApplication request type.
+message MsgSubmitClassApplication {
```
clevinson, line 347: Can we add a field representing the project term? In the context of a credit class, I believe a project will always have a defined project term (start date & end date), and it is then expected that the registry agent will issue credit batches for each vintage year during that project term.

Currently I don't think projects encode this data on-chain, but it might be useful to do if we want to restrict projects from enrolling from two credit classes of the same credit type during the same time period.

clevinson (commented):
Generally looks good. Excited to see how simple this is (at least at the proto level).

I (similarly to sam) think projects should apply just to the credit class (not a specific issuer), though how we handle the application review process & permissioning may make that a bit difficult (e.g. multiple issuers coordinating on a review... see my in-line comment below).

I'm also in favor of a project enrolling in multiple credit classes simultaneously. If we wanted any on-chain enforcement to restrict projects being enrolled in multiple credit classes, we probably only do that for credit classes of the same type.  In other words, a project should not be able to be enrolled in two credit classes of the same type during the same time period.

aaronc:
Makes sense and I agree with uniqueness around credit type. I'll update this PR to reflect that.

proto/regen/ecocredit/v1/state.proto
aaronc: Currently I'm leaning towards the latter for simplicity

aaronc:
I've updated this design based on the comments above and tried to go for as simple as possible. Would appreciate another review @clevinson @S4mmyb or maybe we chat live to resolve any remaining design issues.

clevinson (commented):
Generally looks great. Only concrete suggestions are:
- changing `ProjectClass` name to `ProjectEnrollment` or `ProjectClassEnrollment`
- similarly, where it says "project class relationship" I think it should instead say "project class enrollment" or "project's enrolled in credit class"

clevinson:
Few more general reflections after my call with @aaronc earlier today.

There are 2 API design choices that IMO are OK for us to move forward with in their current form, but probably warrant a bit more conversation w/ @S4mmyb to make sure we are making the best choice that satisfies most of our design concerns while keeping implementation simple.

## Project Enrollment Statuses (& Right to Withdraw) 

Key questions:
1. Who has the right to early terminate a project on-chain? Both `project_admin` & `issuer` ? or just `issuer`? Currently, this is only able to be done by the `project_admin` and can be done at any time. But maybe it's better to have this action be performed by a credit class issuer (assuming a project has already been accepted).  In this scenario, a project could only withdraw an enrollment if it hasn't been "accepted", but once accepted, ending a project would be called "terminate project" and is an action that only an issuer can perform
2. Should we encode any on-chain of the progression between project statuses? I guess this question is separate from this PR (which is focused on API design), but it is an open question whether an issuer should be able to move a project from an "Accepted" state back to "Change Requested" or "Unspecified". IMO once in "Accepted" state, a project should only be able to be "Terminated" from that point on, and that action must be taken by an Issuer.

## Project Terms

@S4mmyb Do you think we have a clear enough idea of the requirements of "Project Terms" that we should incorporate that into this PR?  From my understanding, projects **always** have a defined term (project start date & end date), so it might make sense to add those to this PR as part of a "project enrollment".  This way, when each project applies to a credit class it has a specified project term in advance.

We could use this project term in a few ways:
- projects could automatically be "unenrolled" in a credit class when the project term ends
- allows for pontential on-chain enforcement of double counting (ensuring no project enrolls in 2 credit classes of the same credit type w/ overlapping project terms).  If we implement project terms now, then this would allow for projects to enroll in multiple credit classes (of the same credit type) as long as the project terms are non overlapping (e.g. carbonplus grasslands for 2020-2025, and in 2023 they could begin enrollment process in another carbon credit class for 2026-2030).


## Fees

Probably for a future design, but we shouldn't forget that at some point we'd like to incorporate fixed fees into the project registration flow.  I know Gregory has expressed interest in future implementations that support bonding / locking REGEN when enrolling a project in a credit class, and that REGEN being burned / slashed if a project prematurely terminates.

@aaronc mentioned on our call that its probably best to wait until we have an MVP implemented and some actual usage before we put in more UX hurdles such on-chain / bonded REGEN fees.

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -28,6 +28,48 @@ service Msg {
...
+  // but can have relationships with multiple credit classes across different credit
+  // types. This can be useful, for example for issuing pre-financing forward contracts
+  // while making progress towards issuing credits in an outcome based program.
+  // Projects that are already accepted into a credit class can only update
+  // their metadata when an issuer has changed the status of the relations
+  // to "changes requested".
+  //
+  // Since Revision 3
+  rpc UpdateProjectClass(MsgUpdateProjectClass)
+      returns (MsgUpdateProjectClassResponse);
```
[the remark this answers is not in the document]
aaronc: Updated everything to `ProjectEnrollment` as that rolls of the tongue most nicely for me

aaronc:
@clevinson I've addressed all your comments regarding naming and general cleanup. Will discuss more details around these big outstanding design issues with @S4mmyb today.

proto/regen/ecocredit/v1/events.proto
```diff
@@ -259,3 +260,55 @@ message EventBurnRegen {
   // reason is the reason for the burn.
   string reason = 3;
 }
+
+// EventUpdateProjectClass is emitted when a project admin submits or updates
+// a project class relationship's metadata, usually relating to credit class
+// application.
+message EventUpdateProjectClass {
```
clevinson, line 267: @aaronc should we update these events to also use the new `ProjectEnrollment` terminology instead of `ProjectClass`?

aaronc: Oh thought I did. Updated now

aaronc:
@clevinson based on my discussion with @S4mmyb we decided that a project developer can only create, update or withdraw an application but cannot unilaterally withdraw from a credit class on chain. They need to coordinate this off-chain with the credit class issuer and then the credit class issuer can terminate. So I've added a TERMINATED status and made withdrawal only relevant to an application. I've also simplified and renamed the API so that there are now only two methods for enrollments:
- `CreateOrUpdateApplication` - for a project admin to create, update or withdraw an application
- `UpdateProjectEnrollment` - for a credit class issuer to approve, request changes to or reject an application, or to terminate a project

These two methods have two corresponding events.

We also agreed to not worry about checking enrollment dates or even double enrollment in two classes of the same credit type at all. We're creating transparency here, rather than trying to orchestrate some rule system that isn't fully defined yet.

Generally, this current design feels pretty simple and flexible. From my conversation with @S4mmyb yesterday, it sounds like this general direction is good enough so I might begin some implementation work on the side.

When you get a chance @clevinson and @S4mmyb, would appreciate another round of reviews so we can iron out any details and approve if this is ready.

aaronc:
Although I said that I think more complex fees could be added later (if needed), I did add a governance-controlled project creation fee which seems important for spam prevention.

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -273,12 +323,106 @@ message MsgCreateProjectResponse {
...
+}
+
+// MsgCreateUnregisteredProjectResponse is the Msg/CreateUnregisteredProject response type.
+message MsgCreateUnregisteredProjectResponse {
+  // project_id is the unique identifier of the project.
+  string project_id = 1;
+}
+
+// MsgCreateOrUpdateApplication is the Msg/CreateOrUpdateApplication request type.
+message MsgCreateOrUpdateApplication {
```
paul121, line 366: Why only one message for create and update? What fields can be updated after creation? This is different than Projects that have separate messages for creation and updating individual fields. These message make it very clear what can and cannot be updated. Their corresponding events also make it simple to subscribe to changes.

For applications it seems like only two fields should be updated: `metadata` and `withdraw`. It seems like you would make a new application instead of updating the project ID or credit class on an existing application. And for withdrawing, it isn't clear if that value can be updated multiple times. If you withdraw an application, would it make sense to "lock" the application state from any future updates and instead require creating a new application?

In this case, messages could be simplified to:
- MsgCreateApplication
- MsgUpdateApplicationMetadata
- MsgWithdrawApplication

proto/regen/ecocredit/v1/tx.proto
```diff
@@ -273,12 +323,106 @@ message MsgCreateProjectResponse {
...
+  //
+  // Since Revision 3
+  cosmos.base.v1beta1.Coin fee = 5;
+}
+
+// MsgCreateUnregisteredProjectResponse is the Msg/CreateUnregisteredProject response type.
+message MsgCreateUnregisteredProjectResponse {
+  // project_id is the unique identifier of the project.
+  string project_id = 1;
+}
```
paul121, line 363: How will project IDs be generated? Will they be auto-incrementing IDs unique across all credit classes?

paul121 (commented):
Overall this looks good. I don't have many opinions about the project applications and enrollments as I'm not as familiar with how these processes work today. But the concepts here make sense, and agree with making applications to a credit class generally and not specific issuers.

One thing that I don't see mentioned above - would it be desirable for credit classes to have a status of whether they are accepting applications for new projects? Perhaps credit classes should not be able to deny creation of applications to their credit class (promote open participation). But there may be scenarios when a credit class is no longer "active" or has been replaced by a new credit class that should be used instead.

Otherwise I'm more concerned with the technical changes and how this might affect existing credit classes and projects.
- What is happening to project IDs? (see above) And consistency/inconsistency of project IDs in batch IDs?
- This introduces the concept of "project termination" from a credit class via the project enrollment. But projects that are created directly to a credit class will not have an enrollment created, thus credit class issuers would not be able to "terminate" these projects, until an application/enrollment is created by the project admin. This would also affect existing projects on-chain today, but could potentially be updated in a migration.

proto/regen/ecocredit/v1/tx.proto
aaronc: The main rationale is that there's a fair amount of ceremony for every new method - separate messages, validation, tests, etc. So I can split it up if the single method is overly complex. It's just a little more annoying for implementation.

Withdrawing would delete the application and you'd need to re-apply.

proto/regen/ecocredit/v1/tx.proto
aaronc: Auto-increment across all projects independent of credit class.

aaronc:
> Perhaps credit classes should not be able to deny creation of applications to their credit class (promote open participation). But there may be scenarios when a credit class is no longer "active" or has been replaced by a new credit class that should be used instead.

@S4mmyb @clevinson any thoughts here?

> * What is happening to project IDs? And consistency/inconsistency of project IDs in batch IDs?

If it's a new project ID, ex. `P007`, then its batches would start `C003-P007-` for the class `C003`.

For existing project IDs, say `C003-001`, we could either reassign them a name based on their integer index. Or we could keep the same ID, but make choices when batches are created:
* if creating a batch within `C003` then we still name it `C003-001-` etc.
* if creating a batch in say `C004`, then we could use its unique integer index (say `4`) and the batch would be `C004-P004` in which case `P004` is basically its alternate project ID based on the new ID scheme

> * This introduces the concept of "project termination" from a credit class via the project enrollment. But projects that are created directly to a credit class will not have an enrollment created, thus credit class issuers would not be able to "terminate" these projects, until an application/enrollment is created by the project admin. This would also affect existing projects on-chain today, but could potentially be updated in a migration.

Actually, we'd have a migration for existing credit classes to create the enrollment record.

proto/regen/ecocredit/v1/tx.proto
aaronc: Okay I can make separate messages. Not that much harder

aaronc: Is one message okay for update enrollment?

paul121: Hmm yeah one message for update enrollment makes more sense. It would be common to update both the metadata and status at the same time. Although this would be the only ecocredit entity without its own dedicated update metadata msg :shrug:

aaronc:
I've clarified that whenever an application is withdrawn or status is set to rejected or terminated, that the data is deleted from state allowing the project to apply again to the same credit class in the future. Let me know if this is undesirable for any reason. I know we had discussed keeping the on-chain state for terminated projects @S4mmyb, but in this case, we would need to allow issuers to set the state backed to accepted in order for the project to re-enroll. It seems simpler to just delete and then allow reapplication in the case where projects are terminated which would make the re-enrollment behavior uniform across withdrawn, rejected and terminated states.

proto/regen/ecocredit/v1/tx.proto
S4mmyb: I think of the two options, separating out withdraw into it's own message makes the most sense. Like @paul121 said, the only two things we would really update are the `metadata` and `withdraw`. Having a separate withdraw message makes sense to me. Having a separate update metadata message I'm not sure about, mainly because I I don't know how often we would actually update application metadata or what that process would look like. If an issuer requested changes made to the application, would the project admin do we think it's important to have the issuer update the metadata to reflect the revised / new evidence?

S4mmyb (approved):
Overall I think this looks good to me. I agree with @paul121 's comments about how updating project-ids will happen as projects enter/exit credit classes, or are enrolled in multiple credit classes. For instance, if I have an independent project with id `P002` then enroll in Credit Class `C08` as the 1st project of that credit class, would the id change to `C08-001`? What happens if I unenroll at a later point in time? Does it go back to being `P002` or is it reassigned a new id based on autoincrementing? Also do we reserve the `C08-001` id and just say that that project has been terminated? I have similar questions around id assignment for multiple enrollment. 

Another question I had was who sets the governance-gated project-creation fee. To me this makes sense that it happens at the credit class level as the application process / overhead might differ depending on the nature of the credit class. Is that the intent? Would that be the same thing, or different from a burn fee? 

> One thing that I don't see mentioned above - would it be desirable for credit classes to have a status of whether they are accepting applications for new projects? Perhaps credit classes should not be able to deny creation of applications to their credit class (promote open participation). But there may be scenarios when a credit class is no longer "active" or has been replaced by a new credit class that should be used instead.

Great question @paul121. In general I think it makes sense to allow credit classes to have open vs closed enrollment, but I feel like at this point in time the real-world exploration of what it means to govern a credit class is too premature to really know how we should implment that. Would the admin set it? Could the registry set it? What happens if someone wants to take a deactivated credit class out of retirement? That said, it would be pretty simple to just add an open/closed field set by the credit class admin that would allow applications, and if it is maybe we just go for it.

proto/regen/ecocredit/v1/tx.proto
aaronc: No, all project IDs would be stable and uncorrelated with credit classes

aaronc:
> Overall I think this looks good to me. I agree with @paul121 's comments about how updating project-ids will happen as projects enter/exit credit classes, or are enrolled in multiple credit classes. For instance, if I have an independent project with id `P002` then enroll in Credit Class `C08` as the 1st project of that credit class, would the id change to `C08-001`? What happens if I unenroll at a later point in time? Does it go back to being `P002` or is it reassigned a new id based on autoincrementing? Also do we reserve the `C08-001` id and just say that that project has been terminated? I have similar questions around id assignment for multiple enrollment.

No, it would always be `P002` and your batches would be `C008-P002`. We wouldn't change IDs.

> Another question I had was who sets the governance-gated project-creation fee. To me this makes sense that it happens at the credit class level as the application process / overhead might differ depending on the nature of the credit class. Is that the intent? Would that be the same thing, or different from a burn fee?

I think there's some misunderstanding here. Project creation generally happens independently from applying to credit classes. If we want an application fee for credit classes, that's a different consideration and I'm not sure we need to put that on chain right now. I just want some small anti-spam fee (maybe ~$10USD) but am not trying to deal with everything that might be needed for credit classes

> > One thing that I don't see mentioned above - would it be desirable for credit classes to have a status of whether they are accepting applications for new projects? Perhaps credit classes should not be able to deny creation of applications to their credit class (promote open participation). But there may be scenarios when a credit class is no longer "active" or has been replaced by a new credit class that should be used instead.
> 
> Great question @paul121. In general I think it makes sense to allow credit classes to have open vs closed enrollment, but I feel like at this point in time the real-world exploration of what it means to govern a credit class is too premature to really know how we should implment that. Would the admin set it? Could the registry set it? What happens if someone wants to take a deactivated credit class out of retirement? That said, it would be pretty simple to just add an open/closed field set by the credit class admin that would allow applications, and if it is maybe we just go for it.

We could have a boolean `active` field on credit classes that can be turned on/off to block applications, but honestly not sure that's needed at this point.

proto/regen/ecocredit/v1/tx.proto
paul121: >  Having a separate update metadata message I'm not sure about, mainly because I I don't know how often we would actually update application metadata or what that process would look like. If an issuer requested changes made to the application, would the project admin do we think it's important to have the issuer update the metadata to reflect the revised / new evidence?

My understanding: When an issuer requests changes to an application I would expect the issuer to update the Enrollement status & metadata. I would then expect the project admin to update the Application metadata to address the requested changes. How revisions are represented is outside of the scope here, but could be included within the metadata content.

clevinson (approved):
Lgtm! Approving with a few minor suggestions, but hoping @aaronc can just make the final call with this and others' input in mind.

The one thing I'll add about the separating out of `MsgCreateOrUpdateApplication` is that it's not totally clear from documentation how the user should set the `metadata` field when they are using this msg to withdraw an application. I guess they should leave `metadata` blank?

With this confusion, I do have a preference for us to have a separate explicit msg for withdraw. As for the other actions (create & update), I think both strategies are fine, and don't have a strong preference as to whether they are consolidated or have separate Msg's.

aaronc:
I was assuming that withdrawal would allow metadata and that this would get passed onto the event. In that case do separate withdrawal and create/update messages still make sense? Like I said the arguments and validation would otherwise be 100% the same including metadata although we could remove the awkward withdraw bool.

aaronc:
I'd like to merge this as is without adding a separate withdraw method. Each new method results in > 200 LOC when we include tests and all and with limited resources, maintainability and correctness are big concerns. Reducing implementation complexity can help with that even though I acknowledge it makes the resulting API a bit more complex.

PR #2167, opened 2024-02-13
feat(x/ecocredit):c independent projects

aaronc:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


aaronc:
I am finally marking this as ready for review. Most of the remaining lint errors appear pedantic and can be addressed later (or disabled).

x/ecocredit/base/keeper/features/msg_create_batch.feature
```diff
@@ -18,40 +18,40 @@ Feature: Msg/CreateBatch
     Background:
       Given a credit type with abbreviation "C"
       And a credit class with class id "C01" and issuer alice
-      And a project with project id "C01-001"
+      And a project with project id "P001" enrolled in "C01"
```
paul121, line 21: There is no scenario to cover issuing a credit batch for a project that does not have an `ACCEPTED` enrollment status. Maybe this Rule could be expanded: "The project must exist and have an accepted enrollment to a credit class"

CHANGELOG.md
```diff
@@ -16,6 +16,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - projects to be created without being already enrolled in a credit class
+  - projects to apply for one or more credit class and for project-credit class relationships to be dynamically managed on-chain
+
+#### Changed
+- [#2167](https://github.com/regen-network/regen-ledger/pull/2167):
+  - `FormatProjectID` just takes a sequence number and doesn't need a class ID.
+  - `FormatBatchDenom` requires a class ID because the class is no longer implied by the project ID.
+  - the credit type `P` is now reserved because this is the new project prefix
+  - project and batch string validation is now more lenient and just checks for a non-empty string
+  - `GetCreditTypeFromBatchDenom` takes a `*Denom` parameter instead of a string.
```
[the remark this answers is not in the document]
paul121: See below comment, this function could maybe be renamed

proto/regen/ecocredit/v1/query.proto
```diff
@@ -883,8 +887,28 @@ message QueryProjectEnrollmentsRequest {
...
+
+  // status is the status of the enrollment.
+  ProjectEnrollmentStatus status = 4;
+
+  // application_metadata is any arbitrary metadata set by the project
+  // admin related to its application to the credit class.
+  string application_metadata = 5;
+
+  // enrollment_metadata is any arbitrary metadata set by the credit class
+  // admin evaluating the project's application to the credit class.
```
paul121, line 912:
```suggestion
  // issuer evaluating the project's application to the credit class.
```
It seems that only credit class issuers can update project enrollment?

x/ecocredit/base/keeper/msg_create_or_update_application.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	} else if err != nil {
+		return nil, err
+	}
+
+	if msg.Withdraw {
+		action = types.EventUpdateApplication_ACTION_WITHDRAW
+		if err := k.stateStore.ProjectEnrollmentTable().Delete(ctx, enrollment); err != nil {
+			return nil, err
+		}
```
paul121, line 56: Should a project admin be able to withdraw their application if it already has project enrollment status `ACCEPTED` ?

x/ecocredit/base/keeper/msg_create_project.go
```diff
@@ -41,14 +45,26 @@ func (k Keeper) CreateProject(ctx context.Context, req *types.MsgCreateProject)
...
-		ReferenceId:  req.ReferenceId,
-	}); err != nil {
+	project.Admin = adminAddress
+	project.Jurisdiction = req.Jurisdiction
+	project.Metadata = req.Metadata
+	if req.ReferenceId != "" {
+		project.ReferenceId = req.ReferenceId
+		// only set deprecated class key if reference id is not empty to support the use case of reference IDs
+		project.ClassKey = classInfo.Key //nolint:staticcheck
+	}
```
paul121, line 55: It seems that the reference ID can only be set by credit class issuers when directly creating + enrolling a project into a credit class via `MsgCreateProject` and there is no way to update this value. If a project is created by an admin and applies to a credit class it doesn't seem that the reference ID could be set. Is this intentional? Would it make sense to move the reference ID to the ProjectEnrollmentTable?

x/ecocredit/base/keeper/features/msg_update_project_enrollment.feature
```diff
@@ -0,0 +1,95 @@
...
+        | Bob unspecified to accepted       | UNSPECIFIED       | abc          | Bob    | ACCEPTED   | foo123       | unauthorized | UNSPECIFIED       | abc               |
+        | I01 changes requested to accepted | CHANGES_REQUESTED | bar456       | I01    | ACCEPTED   | foo123       |              | ACCEPTED          | foo123            |
+        | Bob changes requested to accepted | CHANGES_REQUESTED | bar456       | Bob    | ACCEPTED   | foo123       | unauthorized | CHANGES_REQUESTED | bar456            |
+        | I01 update accepted metadata      | ACCEPTED          | foo123       | I01    | ACCEPTED   | bar357       |              | ACCEPTED          | bar357            |
+        | Bob updated accepted metadata     | ACCEPTED          | foo123       | Bob    | ACCEPTED   | bar357       | unauthorized | ACCEPTED          | foo123            |
+
+  Rule: valid state transitions performed by issuers which remove enrollment entries are:
+  UNSPECIFIED -> REJECTED
+  CHANGES_REQUESTED -> REJECTED
+  ACCEPTED -> TERMINATED
```
paul121, line 39: Just want to double check if there are any issues having a project that is `TERMINATED` be removed from the project enrollment state. Considering that a project may have have credit batches issued while it was `ACCEPTED`. I checked the implementation to make sure we can still query those credit batches and it seems like that query would still work (not changed in this PR). Are there any other cases to consider?

paul121 (commented):
Nice work @aaronc this is a lot but looking pretty good to me. I think I have review most of these changes but might take a second look with fresh eyes later.

I was going to try and run this locally to test a few things out but just saw the separate PR #2202 that adds CLI commands. Would this be worth testing at this point?

aaronc:
> I was going to try and run this locally to test a few things out but just saw the separate PR #2202 that adds CLI commands. Would this be worth testing at this point?

If you're going to test manually, the CLI commands would be useful. How should we handle this? Would you test in that branch? Or should I just merge the CLI into here?

x/ecocredit/base/keeper/msg_create_or_update_application.go
paul121: Makes sense. This logic might require a change then because there doesn't seem to be a check for the enrollment status here. Could use a test scenario as well

paul121:
> > I was going to try and run this locally to test a few things out but just saw the separate PR #2202 that adds CLI commands. Would this be worth testing at this point?
> 
> If you're going to test manually, the CLI commands would be useful. How should we handle this? Would you test in that branch? Or should I just merge the CLI into here?

I can do either way. Comparing the two branches, looks like there aren't many changes. Maybe merging here is easiest. https://github.com/regen-network/regen-ledger/compare/aaronc/independent-projects-impl...aaronc/independent-projects-cli

x/ecocredit/base/keeper/features/msg_update_project_enrollment.feature
aaronc: I think it's okay because removing it will allow it to re-enroll.

x/ecocredit/base/keeper/msg_create_project.go
aaronc: The reference ID is only relevant in the case of bridging and I can consider it a legacy use case. It needs to be in the project table to impose an unique constraint on it, again only needed for bridging.

aaronc:
> > > I was going to try and run this locally to test a few things out but just saw the separate PR #2202 that adds CLI commands. Would this be worth testing at this point?
> > 
> > 
> > If you're going to test manually, the CLI commands would be useful. How should we handle this? Would you test in that branch? Or should I just merge the CLI into here?
> 
> I can do either way. Comparing the two branches, looks like there aren't many changes. Maybe merging here is easiest. [aaronc/independent-projects-impl...aaronc/independent-projects-cli](https://github.com/regen-network/regen-ledger/compare/aaronc/independent-projects-impl...aaronc/independent-projects-cli)

I went ahead and merged the CLI code into this PR.

x/ecocredit/base/keeper/msg_create_project.go
paul121: > The reference ID is only relevant in the case of bridging and I can consider it a legacy use case. 

Oh interesting, I did not realize this was only intended for bridging. In fact we are using the reference ID field in another way on some projects we are working on (SeaTrees, ZFP). It is useful as a key to reference on-chain projects with existing "external IDs" programs may already be using. This makes it much easier to create scripts that update projects in bulk when metadata might be changing in an external CMS of some kind.

It would be good to know if this will be supported going forward. These projects/processes are being deployed on Redwood now and could be deployed on mainnet in the coming months.

> // reference_id is any arbitrary string that can be use to reference project.

https://buf.build/regen/regen-ledger/docs/main:regen.ecocredit.v1#regen.ecocredit.v1.ProjectInfo

aaronc: I don't think we're planning to remove the support, but I'd prefer not to add or update the functionality much unless there's a specific use case. Basically, it can only be added at creation, not modified and there will be an unique constraint on reference ID and credit class.

x/ecocredit/server/utils/utils.go
```diff
@@ -5,21 +5,19 @@ import (
...
 	"github.com/regen-network/regen-ledger/types/v2/math"
-	"github.com/regen-network/regen-ledger/x/ecocredit/v3/base"
 )
 
 // GetCreditTypeFromBatchDenom extracts the classID from a batch denom string, then retrieves it from the params.
-func GetCreditTypeFromBatchDenom(ctx context.Context, store api.StateStore, denom string) (*api.CreditType, error) {
-	classID := base.GetClassIDFromBatchDenom(denom)
-	classInfo, err := store.ClassTable().GetById(ctx, classID)
+func GetCreditTypeFromBatchDenom(ctx context.Context, store api.StateStore, batch *api.Batch) (*api.CreditType, error) {
+	cls, err := store.ClassTable().Get(ctx, batch.ClassKey)
```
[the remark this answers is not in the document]
aaronc: Renamed to `GetCreditTypeFromBatch` and updated the docs

x/ecocredit/base/keeper/features/msg_create_batch.feature
aaronc: Added a test to cover this: https://github.com/regen-network/regen-ledger/pull/2167/commits/4336def78e798ac87a8fe61b63a31103e217cb50

x/ecocredit/base/keeper/msg_create_or_update_application.go
aaronc: Added a test scenario here: https://github.com/regen-network/regen-ledger/pull/2167/commits/7c6fdb2e4be2202be56aed6ed48436e9da221f2f and turns out this logic wasn't properly implemented. So good catch @paul121 👍

x/ecocredit/base/keeper/features/msg_create_or_update_application.feature
```diff
@@ -61,4 +61,11 @@ Feature: Msg/CreateOrUpdateApplication
       Then expect error contains "unauthorized"
       And expect the application for "P001" to "C01" exists with metadata "abc123"
 
-  Rule: events get emitted
+  Rule: project admins cannot withdraw a project with an accepted enrollment (a credit class issuer must do that)
```
paul121, line 64:
```suggestion
  Rule: project admins cannot withdraw a project with an accepted enrollment
```

Just got hung up on this getting my head back into things.. this implies a credit class issuer can withdraw a project. But only project admins can attempt a withdraw via `MsgCreateorUpdateApplication`. To remove the project enrollment from state credit class issuers must change enrollment status from `ACCEPTED` -> `TERMINATED`. Maybe just drop `(a credit class issuer must do that)` for clarity?

x/ecocredit/base/keeper/features/msg_create_batch.feature
paul121: Looks good

x/ecocredit/base/keeper/msg_create_or_update_application.go
paul121: Implementation looks good, left a suggestion re: the test rule text

x/ecocredit/base/keeper/msg_update_project_enrollment.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+	case ecocreditv1.ProjectEnrollmentStatus_PROJECT_ENROLLMENT_STATUS_ACCEPTED:
+		switch newStatus {
+		case ecocreditv1.ProjectEnrollmentStatus_PROJECT_ENROLLMENT_STATUS_ACCEPTED:
+			// Valid case for just updating metadata of an accepted project
+		case ecocreditv1.ProjectEnrollmentStatus_PROJECT_ENROLLMENT_STATUS_TERMINATED:
+			remove = true
+		default:
+			return nil, sdkerrors.ErrInvalidRequest.Wrapf("invalid status transition from %s to %s", existingStatus, newStatus)
+		}
```
paul121, line 62: I think these state transitions from enrollment status `ACCEPTED` could have better tests. Right now I only see one test case for the valid `ACCEPTED` -> `TERMINATED` transition, and one for the invalid `ACCEPTED` -> `REJECTED` transition.

x/ecocredit/base/keeper/features/msg_update_project_enrollment.feature
```diff
@@ -0,0 +1,95 @@
...
+        }
+      """
+      Examples:
+        | scenario                            | cur_status        | cur_metadata | issuer | new_status | new_metadata | err          | exists |
+        | IO1 unspecified to rejected         | UNSPECIFIED       | abc          | I01    | REJECTED   | baz789       |              | false  |
+        | Bob unspecified to rejected         | UNSPECIFIED       | abc          | Bob    | REJECTED   | baz789       | unauthorized | true   |
+        | I01 changes requested to rejected   | CHANGES_REQUESTED | bar456       | I01    | REJECTED   | baz789       |              | false  |
+        | Bob changes requested to rejected   | CHANGES_REQUESTED | bar456       | Bob    | REJECTED   | baz789       | unauthorized | true   |
+        | I01 accepted to rejected            | ACCEPTED          | foo123       | I01    | REJECTED   | baz789       | invalid      | true   |
+        | Bob accepted to rejected            | ACCEPTED          | foo123       | Bob    | REJECTED   | baz789       | unauthorized | true   |
```
paul121, line 61: Related to my other comment, I think this is the only test for an invalid state transition from enrollment status `ACCEPTED`.

aaronc: Added more tests. Let me know if that looks good enough

paul121: Yep, looks good!

paul121 (changes requested):
This is working well! I didn't find any major state or logic issues while testing. Tests seem to cover this pretty well :+1:

Just found various little things, mostly that a few of the CLI commands weren't quite wired up and could maybe be improved for consistency.

~(@aaronc oops I had both PRs open and left in-line comments [over here](https://github.com/regen-network/regen-ledger/pull/2202#pullrequestreview-2150219526) on the wrong one.. I might move them over here if I have a bit more time this evening)~ done

Simple fixes I can't suggest via github UI:
- Need to add project enrollment query cmds to x/ecocredit/client/query.go QueryCmd():

    ```
            baseclient.QueryProjectEnrollmentCmd(),
            baseclient.QueryProjectEnrollmentsCmd(),
    ```

- Error with the example batch.json in `create-batch` command help text. It is missing a `class_id`
- Doesn't need to be fixed in this PR, but the `ecocredit` script in images/regen-sandbox/setup/ecocredit.sh needs some updating (some commands fail due to these changes)

The general cases I tested:
- Creating projects independent vs direct, testing correct permissions to update project metadata, etc
- Creating application, withdrawing application, testing correct permissions + good/bad project + class IDs, metadata
- Updating enrollment status, testing correct permissions, correct status transitions, good/bad project + class IDs, metadata
- Creating credit batch, testing correct permissions, good/bad project + class IDs
- "Full flow": Create project, apply, enroll, issue credits, terminate enrollment, try to issue credits, re-enroll, issue credits

I think this covered using/testing the following CLI commands:

- tx ecocredit:
  - create-class
  - create-project 
  - update-project-metadata
  - create-or-update-application
  - withdraw-application
  - update-project-enrollment
  - create-batch
  - gen-batch-json

- q ecocredit:
  - batches
  - class, classes
  - class-issuers
  - project, projects
  - project-enrollment
  - project-enrollments


Could improve error messages when proj + class enrollments do not exist/have an entry in the enrollment DB table:
- withdraw-application
  - when the credit class or project ID does not exist gets generic error msg: `raw_log: 'failed to execute message; message index: 0: not found'`
  - when valid credit class and project ID, but no application/enrollment between the two, gets a better error msg: `raw_log: 'failed to execute message; message index: 0: cannot withdraw non-existent application: invalid request'`
- create-batch
  - when credit class does not exist or when no enrollment for project + class: `raw_log: 'failed to execute message; message index: 0: not found'`
  - when project does not exist, better error msg: `raw_log: 'failed to execute message; message index: 0: could not get project with id P005: not found: invalid request'`

Project enrollment status - just an idea
- There is no way for a credit class to signal that they have "received" an application and are beginning review, maybe give an expected timeframe, etc. I assume this could happen via enrollment metadata.
  - The credit class can only add metadata when setting the status to ACCEPTED or CHANGES_REQUESTED.
  - Would it make sense to add a project enrollment status for "PENDING" / "IN REVIEW" so there is an intermediary status?

x/ecocredit/base/client/query.go
```diff
@@ -667,3 +667,66 @@ func QueryAllowedBridgeChainsCmd() *cobra.Command {
 	flags.AddPaginationFlagsToCmd(cmd, "allowed-bridge-chains")
 	return qflags(cmd)
 }
+
+func QueryProjectEnrollment() *cobra.Command {
```
paul121, line 671:
```suggestion
func QueryProjectEnrollmentCmd() *cobra.Command {
```

x/ecocredit/base/client/query.go
```diff
@@ -667,3 +667,66 @@ func QueryAllowedBridgeChainsCmd() *cobra.Command {
...
+				ProjectId: args[0],
+				ClassId:   args[1],
+			})
+			return printQueryResponse(ctx, res, err)
+		},
+	}
+	return qflags(cmd)
+}
+
+func QueryProjectEnrollments() *cobra.Command {
```
paul121, line 693:
```suggestion
func QueryProjectEnrollmentsCmd() *cobra.Command {
```

x/ecocredit/base/client/tx.go
```diff
@@ -109,48 +113,77 @@ regen tx ecocredit create-class regen1elq7ys34gpkj3jyvqee0h6yk4h9wsfxmgqelsw,reg
...
+	var classID, referenceID, projectFee string
+
 	cmd := &cobra.Command{
-		Use:   "create-project [class-id] [jurisdiction] [metadata] [flags]",
-		Short: "Create a new project within a credit class",
-		Long: `Create a new project within a credit class.
+		Use:   "create-project [jurisdiction] [metadata] [flags]",
+		Short: "Create a new project",
+		Long: `Create a new project. By default, this creates an unregistered project that is not enrolled in any credit
+class. If the class-id flag is provided, the signer must be an issuer of the class and the project will be enrolled in
```
paul121, line 122:
```suggestion
class. If the class flag is provided, the signer must be an issuer of the class and the project will be enrolled in
```

x/ecocredit/base/client/tx.go
```diff
@@ -109,48 +113,77 @@ regen tx ecocredit create-class regen1elq7ys34gpkj3jyvqee0h6yk4h9wsfxmgqelsw,reg
...
-- reference-id: a reference ID for the project`,
-		Example: `regen tx ecocredit create-project C01 "US-WA 98225" regen:13toVgf5UjYBz6J29x28pLQyjKz5FpcW3f4bT5uRKGxGREWGKjEdXYG.rdf
-regen tx ecocredit create-project C01 "US-WA 98225" regen:13toVgf5UjYBz6J29x28pLQyjKz5FpcW3f4bT5uRKGxGREWGKjEdXYG.rdf  --reference-id VCS-001`,
-		Args: cobra.ExactArgs(3),
+- project-fee: the fee that the project creator will pay to create the project. It must be >= the
+required project creation fee param. If the project creation fee is zero, no fee is required. 
+We explicitly include the project creation fee here so that the project creator acknowledges paying 
+the fee and is not surprised to learn that the they paid a fee without consent.
+- class:      the ID of the credit class. If this is provided, the signer must be an issuer of the class.
+- reference-id: a reference ID for the project. Only valid if class-id is also provided.`,
```
paul121, line 137:
```suggestion
- reference-id: a reference ID for the project. Only valid if class is also provided.`,
```

x/ecocredit/base/client/tx.go
```diff
@@ -109,48 +113,77 @@ regen tx ecocredit create-class regen1elq7ys34gpkj3jyvqee0h6yk4h9wsfxmgqelsw,reg
...
+		Example: `regen tx ecocredit create-project "US-WA 98225" regen:13toVgf5UjYBz6J29x28pLQyjKz5FpcW3f4bT5uRKGxGREWGKjEdXYG.rdf
+regen tx ecocredit create-project "US-WA 98225" regen:13toVgf5UjYBz6J29x28pLQyjKz5FpcW3f4bT5uRKGxGREWGKjEdXYG.rdf  --class-id C01 --reference-id VCS-001`,
```
paul121, line 139:
```suggestion
regen tx ecocredit create-project "US-WA 98225" regen:13toVgf5UjYBz6J29x28pLQyjKz5FpcW3f4bT5uRKGxGREWGKjEdXYG.rdf  --class C01 --reference-id VCS-001`,
```

x/ecocredit/base/client/tx.go
```diff
@@ -159,24 +192,25 @@ regen tx ecocredit create-project C01 "US-WA 98225" regen:13toVgf5UjYBz6J29x28pL
...
 - issuer:          the account address of the credit batch issuer
 - project-id:      the ID of the project
+- class-id:		   the ID of the credit class to which the batch belongs and in which the project is enrolled
 - issuance-count:  the number of issuance items to generate
 - metadata:        any arbitrary metadata to attach to the credit batch
 - start-date:      the beginning of the period during which this credit batch was
                    quantified and verified (format: yyyy-mm-dd)
 - end-date:        the end of the period during which this credit batch was
                    quantified and verified (format: yyyy-mm-dd)`,
 		Example: "regen tx ecocredit gen-batch-json regen1elq7ys34gpkj3jyvqee0h6yk4h9wsfxmgqelsw C01-001 2 regen:13toVgf5UjYBz6J29x28pLQyjKz5FpcW3f4bT5uRKGxGREWGKjEdXYG.rdf 2020-01-01 2021-01-01",
```
paul121, line 210:
```suggestion
		Example: "regen tx ecocredit gen-batch-json regen1elq7ys34gpkj3jyvqee0h6yk4h9wsfxmgqelsw P001 C01 2 regen:13toVgf5UjYBz6J29x28pLQyjKz5FpcW3f4bT5uRKGxGREWGKjEdXYG.rdf 2020-01-01 2021-01-01",
```

x/ecocredit/base/client/tx.go
```diff
@@ -795,3 +830,134 @@ Example JSON:
...
+		ProjectId:    args[0],
+		ClassId:      args[1],
+		Metadata:     metadata,
+		Withdraw:     withdraw,
+	}
+
+	return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), msg)
+}
+
+func TxUpdateProjectEnrollment() *cobra.Command {
```
paul121, line 909: consistency

```suggestion
func TxUpdateProjectEnrollmentCmd() *cobra.Command {
```

x/ecocredit/client/tx.go
```diff
@@ -36,6 +36,8 @@ func TxCmd(name string) *cobra.Command {
 		baseclient.TxUpdateProjectMetadataCmd(),
 		baseclient.TxUpdateBatchMetadataCmd(),
 		baseclient.TxBridgeCmd(),
+		baseclient.TxCreateOrUpdateApplicationCmd(),
+		baseclient.TxWithdrawApplicationCmd(),
```
paul121, line 40:
```suggestion
		baseclient.TxCreateOrUpdateApplicationCmd(),
		baseclient.TxWithdrawApplicationCmd(),
		baseclient.TxUpdateProjectEnrollmentCmd(),
```

aaronc:
@paul121 I believe I have addressed all of the simple changes you have suggested.

Regarding:
> There is no way for a credit class to signal that they have "received" an application and are beginning review, maybe give an expected timeframe, etc. I assume this could happen via enrollment metadata.
The credit class can only add metadata when setting the status to ACCEPTED or CHANGES_REQUESTED.
Would it make sense to add a project enrollment status for "PENDING" / "IN REVIEW" so there is an intermediary status?

I think two straightforward ways are either:
1. allowing metadata to be added by the issuer in UNSPECIFIED state (without transitioning to ACCEPTED or CHANGES_REQUESTED)
2. or adding a PENDING state like you suggest

Either will involve adding some new test cases, but option 1 seems probably simpler to implement at this stage. Would that be okay or do you think there's a strong argument for a PENDING case? If we add a pending case would we allow transitioning from CHANGES_REQUESTED to PENDING?

paul121:
> Either will involve adding some new test cases, but option 1 seems probably simpler to implement at this stage. Would that be okay or do you think there's a strong argument for a PENDING case? If we add a pending case would we allow transitioning from CHANGES_REQUESTED to PENDING?

I don't have a strong argument for a PENDING case. I agree option 1 makes sense for now, it's more general and flexible than adding any specific state. Chatting with @clevinson today he said there might be cases when a credit class doesn't want to reveal they are starting a formal review anyways, so yeah, let’s keep it simple for now, allowing metadata in the UNSPECIFIED state.

aaronc:
@paul121 I updated `UpdateProjectEnrollment` to allow transitions from `UNSPECIFIED` or `CHANGES_REQUESTED` to `UNSPECIFIED` with updates to metadata in https://github.com/regen-network/regen-ledger/pull/2167/commits/fd96f717e8fa1574b9fe8a941b8a6fc82c752434. Please review that commit and let me know what you think.

x/ecocredit/base/client/tx.go
```diff
@@ -109,48 +113,77 @@ regen tx ecocredit create-class regen1elq7ys34gpkj3jyvqee0h6yk4h9wsfxmgqelsw,reg
...
+
 	cmd := &cobra.Command{
-		Use:   "create-project [class-id] [jurisdiction] [metadata] [flags]",
-		Short: "Create a new project within a credit class",
-		Long: `Create a new project within a credit class.
+		Use:   "create-project [jurisdiction] [metadata] [flags]",
+		Short: "Create a new project",
+		Long: `Create a new project. By default, this creates an unregistered project that is not enrolled in any credit
+class. If the class flag is provided, the signer must be an issuer of the class and the project will be enrolled in
+the class automatically.
```
paul121, line 123: Just took a look at the tests, and it looks like an integration test is failing due to changes from this command `regen tx ecocredit create-project`. The class-id should be passed as a flag now.

Test failure: https://github.com/regen-network/regen-ledger/actions/runs/9872434922/job/27262543729#step:5:216

Integration test: https://github.com/regen-network/regen-ledger/blob/3d818cf6e01af92eed25de5c17728a79070f56a3/x/ecocredit/client/testsuite/suite.go#L380-L390

paul121 (commented):
@aaronc the changes to `UpdateProjectEnrollment` look good. Just noticing a test failure related to the CLI changes. Not sure if there are other integration tests to review as well

paul121 (approved):
Added a commit to fix one remaining integration test. I think this is good now.

x/ecocredit/base/keeper/msg_create_project.go
```diff
@@ -63,28 +79,23 @@ func (k Keeper) CreateProject(ctx context.Context, req *types.MsgCreateProject)
 	}, nil
 }
 
-// genProjectID generates a projectID when no projectID was given for CreateProject.
+// createNewProject generates a projectID when no projectID was given for CreateProject.
 // The ID is generated by concatenating the classID and a sequence number.
```
paul121, line 83:
```suggestion
// createNewProject creates a new project when no class was given for CreateProject.
```

clevinson (commented):
I've gone through all the code and generally looks good.

Handful of small test / comment related suggestions.  In general I think we should lean towards using the new project ID's in all of our feature files / comments where possible, unless there's a specific reason to keep the legacy format.

One thing I'm a bit unclear on is how this PR handles creation of new batches for legacy projects (whether or not those batches are under a legacy credit class).  By legacy here I simply mean a project or credit class created prior to this upgrade.

In the state migration, it looks like we don't update any existing Project ID's to the new format (as stored in the ORM table). So if a project (`C01-001`) has a pre-existing batch (`C01-001-20210101-20220101-001`), and then has a new batch issued after this update, I think the new batch would be `C01-C01-001-20220101-20230101-002`, since the new formatting of batch denom explicitly adds credit class ID to the beginning, even if project ID is in the legacy format.

Similarly, if this project enrolls in a second credit class, and creates a batch, then that batch would be `C02-C01-001-20210101-20220101-001`.

I understand why it's preferable to preserve legacy Project IDs and not change them, as that may break compatibility with certain RDF datasets, Sanity, or credit batch linkages. But are there any alternatives where we could maybe keep the legacy format for legacy enrollments, and generate a new format project ID only when a legacy project creates a new enrollment?  Another alternative may be just not allowing legacy projects to have multiple enrollments at all.

aaronc:
Here's my proposed solution for generating new batches IDs for existing projects:
1) if they are emitting a batch in their original credit class, then just use the old batch format, i.e. `C01-001...`
2) add an `alt_batch_prefix` field to `Project` that only can get set during the migration which gives them some alternate prefix to use if they emit a batch with a new credit class, that batch prefix could be something like `L001` with `L` for legacy to distinguish them from the new `P` projects

How does that sound?

clevinson:
So just to confirm,

If credit legacy project `C01-001` registers then with a credit class `C02`, the first batch under this new credit class would look like `C02-L001-20210101-20220101-001` ?

I'm trying to think if there are any queries we should be considering that may have weird or potentially confusing UX with this.

One that comes to mind -  when querying for `ProjectEnrollments`, one would query by the original project ID (eg `C01-001`) to get a list of all enrolled credit classes, right?

This doesn't seem the prettiest but allowing queries by the alt project id also seems confusing.

The only other option I can think of is preventing legacy projects from enrolling in new credit classes. Personally this path seems cleanest to me, but I'm happy to default to whichever approach you think is best. 

Assuming we do want to allow new enrollments from legacy projects, I think the proposal above makes sense and would support it.

PR #2286, opened 2025-09-08
feat: remove intertx module

vishal-kanna:


x/ecocredit/marketplace/types/v1/msg_gov_set_fee_params.go
```diff
@@ -19,12 +18,12 @@ func (m *MsgGovSetFeeParams) ValidateBasic() error {
 		return err
 	}
 
-	_, err = types.AccAddressFromBech32(m.Authority)
+	_, err = sdk.AccAddressFromBech32(m.Authority)
```
aleem1314, line 21: Remove global Bech32 decoding and use address code instead.

